Issue #31, opened 2023-11-28
[RC-40] [Feature] Allows to use dbt jinjas function and macros in the query


popcornylu:
### Describe the feature

Currently, recce only support the macro `ref(<node>)`. It would be useful to use macros like we define a model.

Recce can consider to support
- Support [dbt jinja functions](https://docs.getdbt.com/reference/dbt-jinja-functions)
- Support [macros](https://docs.getdbt.com/docs/build/jinja-macros)
- Support [macros from installed packages](https://docs.getdbt.com/docs/build/packages)

<sub>[RC-40](https://linear.app/piperider/issue/RC-40/[feature]-allows-to-use-dbt-jinjas-function-and-macros-in-the-query)</sub>

PR #49, opened 2023-12-05
[Feature] support dbt macros

qrtt1:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

feature added

leverage dbt packages 

#31 

demos
----

```yaml
{% set a_query %}
select * from {{ ref("cloud_user_last_seen") }} where id>100
{% endset %}

{% set b_query %}
select * from {{ ref("cloud_user_last_seen") }} where id<100
{% endset %}

{{ audit_helper.compare_queries(
    a_query=a_query,
    b_query=b_query,
    primary_key="id"
) }}
```

image


popcornylu:
@qrtt1 thanks

Two recommendataion.
1. Use the dbt compile by default. I think there is no reason to use the old one. (i.e. Remove the feature flg `recce_sql_engine` )
2. Can we just put the code in `dbt.py`? Or move the dbt to the package `recce.dbt`?
3. Use `audit_helper` seems not suitable for current query diff, could you change to use [dbt_profiler.get_profile](https://github.com/data-mie/dbt-profiler/tree/0.8.1/#get_profile-source)

qrtt1:
@popcornylu old implementation removed

image

popcornylu:
@qrtt1  it seems that the query for base always run the result of the current environment.

Could you provide a screenshot that show there is a difference found?

recce/dbt.py
```diff
@@ -111,37 +175,24 @@ def find_source_by_name(self, source_name, table_name, base=False) -> Optional[S
...
-
-            relation = self.adapter.Relation.create_from(self.project, source)
-            return str(relation)
-
-        template = Template(sql_template)
-        sql = template.render(ref=ref, source=source)
-
         adapter = self.adapter
         with adapter.connection_named('test'):
+            sql = self.generate_sql(sql_template)
```
popcornylu: base argument is missing?

popcornylu (approved):

qrtt1:
Diff of profiler :D

```sql
{{ dbt_profiler.get_profile(relation=ref("cloud_user_last_seen")) }}
```
```patch
diff --git a/models/marts/cloud_user_last_seen.sql b/models/marts/cloud_user_last_seen.sql
index a03db61..f15b4e8 100644
--- a/models/marts/cloud_user_last_seen.sql
+++ b/models/marts/cloud_user_last_seen.sql
@@ -9,6 +9,7 @@ WITH ReportDate AS (
         MAX(created_at) AS last_seen,
         DATEDIFF(day, MIN(created_at), MAX(created_at)) AS active_days
     FROM {{ ref('stg_cloud__audit_logs') }}
+    WHERE user_id > 100
     GROUP BY user_id
 )
 


```

image

PR #62, opened 2023-12-13
[Feature] Add checklist backend

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- checklist create and update API

rc-46

NONE

### DIFF
--- a/recce/apis/check_api.py
+++ b/recce/apis/check_api.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from typing import Optional
 from uuid import UUID
 
@@ -10,8 +11,8 @@ check_router = APIRouter(tags=['check'])
 
 
 class CreateCheckIn(BaseModel):
-    name: str
-    description: str = ''
+    name: Optional[str] = f"Check-{datetime.utcnow().isoformat()}"
+    description: Optional[str] = ''
     run_id: str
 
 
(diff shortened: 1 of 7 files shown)


recce/apis/run_func.py
```diff
@@ -0,0 +1,39 @@
...
+
+from recce.apis.types import RunType, RunResultType
+from recce.server import dbt_context
+
+
+class SqlQueryParams(TypedDict):
+    sql_template: str
+
+
+class RunExecutor:
```
popcornylu: From the name, SqlQueryExecutor is a RunExecutor, is it more suitable to use subclass?

recce/apis/check_api.py
```diff
@@ -0,0 +1,114 @@
...
+        if not check_record:
+            raise HTTPException(status_code=404, detail=f"Run ID '{check.run_id}' not found")
+        checks_db.append(check_record)
+
+    return {
+        'id': check_record.id,
+        'name': check_record.name,
+        'description': check_record.description,
+        'params': check_record.params,
+        'last_run': check.run_id
```
popcornylu: `last_run` should be run entity rather than id

recce/apis/run_api.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class CreateRunIn(BaseModel):
+    type: str
+    params: dict
+    check_id: Optional[str]
+
+
+@run_router.post("/runs", status_code=201)
+async def create(run: CreateRunIn):
```
popcornylu, line 28: Move the create run from check to `POST /checks/<check_id>/run`.

The `POST /runs` should only supports the non-check run.

recce/apis/run_api.py
```diff
@@ -0,0 +1,80 @@
...
+    from recce.apis.run_func import RunExecutor
+
+    run_record = None
+    result_type = None
```
popcornylu: After implementing FE, i think result_type is not useful by now, we can just remove it.

recce/apis/run_api.py
```diff
@@ -0,0 +1,80 @@
...
+        result_type = executor.get_result_type()
+
+        run_record = Run(run_type, run.params)
+    else:
+        found = False
+        for check in checks_db:
+            if str(check.id) == run.check_id:
+                executor = RunExecutor().get_executor(check.type)
+
+                result['base'], result['base_error'] = executor.execute(check.params, base=True)
```
popcornylu: A executor should not always run twice on base&current. The exectur signature should change. Use the abstract base to implement.

https://docs.python.org/zh-tw/3/library/abc.html

recce/apis/run_func.py
```diff
@@ -0,0 +1,39 @@
...
+    @staticmethod
+    def get_executor(executor_type: RunType):
+        if executor_type == RunType.QUERY_DIFF:
+            return SqlQueryExecutor()
+        else:
+            raise ValueError("Invalid executor type")
+
+
+class SqlQueryExecutor:
```
popcornylu: Try to modify this way.


SqlQueryExectuor -> QueryDIffExecutor.

Run base and current query in the `execute()` method

PR #74, opened 2023-12-18
[Feature] rc-75 Add row count diff in nodw view page

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Add number of row count information in the NodeView page
- Add backend API to query row count by model name

RC-75


js/src/components/lineage/NodeView.tsx
```diff
@@ -36,21 +94,44 @@ interface NodeViewProps {
 
 export function NodeView({ node, onCloseNode }: NodeViewProps) {
   const router = useRouter();
+  const [rowCount, setRowCount] = useState<ModelRowCount>();
```
popcornylu: Please use the const {isLoading, data: rowCount} useQuery = ...

js/src/components/lineage/NodeView.tsx
```diff
@@ -36,21 +94,44 @@ interface NodeViewProps {
...
+  const [isRowCountLoaded, setRowCountIsLoaded] = useState(false);
   const { setSqlQuery } = useRecceQueryContext();
   const withColumns =
     node.resourceType === "model" ||
     node.resourceType === "seed" ||
     node.resourceType === "source";
   const { isOpen, onOpen, onClose } = useDisclosure();
   const { MismatchSummaryModal } = useMismatchSummaryModal();
+  const { icon: resourceTypeIcon } = getIconForResourceType(node.resourceType);
+  const { mutate: queryRowCount } = useMutation({
```
popcornylu: This is not mutation, please use useQuery({...., skip: node.resourceType === "model"})

popcornylu (changes requested):
Pls use useQuery insteadof useMutation

popcornylu:
In data warehouse, to query row count is sometimes costly. Especially for view where the underlying is a join operation. So

1. Could we query the row count until the user request it?
2. And when i discuss with CL, the original context is that can we visualize the models with row count diff? I mentioned that we can provide a button to query all and see the diff, but it's sometimes costly for view model. So my recommendation is to query only the table/incremental models, because they are table in warehouse, where the row count is metadata in the warehouse. It is super cheap.

kentwelcome:
> In data warehouse, to query row count is sometimes costly. Especially for view where the underlying is a join operation. So
> 
> 1. Could we query the row count until the user request it?
Sure. I agree the cost of query row count is expensive. However, the row count info is still valuable on the NodeView page. 
I will propose:
1. In the NodeView page, let the `Number of row` tag as a clickable button. We only query the row count when the user clicks it.
2. In the summary view, we will show all the schema of modified models. In this case, we can query all row counts of the changed models and show them to the user.

Do those propose meet your expectations?

popcornylu:
> > In data warehouse, to query row count is sometimes costly. Especially for view where the underlying is a join operation. So
> > 
> > 1. Could we query the row count until the user request it?
> 
> Sure. I agree the cost of query row count is expensive. However, the row count info is still valuable on the NodeView page. I will propose:
> 
> 1. In the NodeView page, let the `Number of row` tag as a clickable button. We only query the row count when the user clicks it.
> 2. In the summary view, we will show all the schema of modified models. In this case, we can query all row counts of the changed models and show them to the user.
> 
> Do those propose meet your expectations?

1. OK
2. Schema chage != row count change. It's totally different check. For example, If i add a filter in the upstream intermediate model with filter, but the downstream mark customer dimension table is impacted. In this case, there is no schema change, and the impacted models is not modified model. 

What makes more sense to me is

As a submitter, I would like to fetch row counts for all (or impacted downstream) MATERIALIZED models. So that i can get the row count from the node view and (optionally) know the models with row count changes.

The points are
1. Materialized models
2. Not only modified model
3. One click to fetch all table's row counts
4. No strong opinion on where to show the one click button.

kentwelcome:
> > > In data warehouse, to query row count is sometimes costly. Especially for view where the underlying is a join operation. So
> > > 
> > > 1. Could we query the row count until the user request it?
> > 
> > 
> > Sure. I agree the cost of query row count is expensive. However, the row count info is still valuable on the NodeView page. I will propose:
> > 
> > 1. In the NodeView page, let the `Number of row` tag as a clickable button. We only query the row count when the user clicks it.
> > 2. In the summary view, we will show all the schema of modified models. In this case, we can query all row counts of the changed models and show them to the user.
> > 
> > Do those propose meet your expectations?
> 
> 1. OK
> 2. Schema chage != row count change. It's totally different check. For example, If i add a filter in the upstream intermediate model with filter, but the downstream mark customer dimension table is impacted. In this case, there is no schema change, and the impacted models is not modified model.
> 
> What makes more sense to me is
> 
> As a submitter, I would like to fetch row counts for all (or impacted downstream) MATERIALIZED models. So that i can get the row count from the node view and (optionally) know the models with row count changes.
> 
> The points are
> 
> 1. Materialized models
> 2. Not only modified model
> 3. One click to fetch all table's row counts
> 4. No strong opinion on where to show the one click button.

From my point of view, I don't think a single button to load all the row counts is a good design.
If we can predict the user's behavior, we should prepare it before the user clicks the button. In this case, I would prefer to auto-load the row counts of child nodes when the user clicks to show the row count of a model. 

Seems like we still don't know how to present the row count.
Let me merge this PR, and we can improve the strategy of row count display in the future.

PR #82, opened 2023-12-20
[Feature] Allow make value-diff as the check item

qrtt1:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

feature added

Make the value-diff being able to add to the check list

rc-66


playground.py
```diff
@@ -0,0 +1,20 @@
+import time
```
popcornylu: It seems it should not be committed?

popcornylu (commented):
Some main problem
1. If there are table with rename column, it will fail.
   
image


2. (Good to have improvlement) If the use select a non pk column. It will take a lot of time. Please check if it is primary key before run the mistmatch. possible solution `not_nul`l and `unique` test on both side. [ref](https://docs.getdbt.com/reference/resource-properties/data-tests#out-of-the-box-data-tests) or run sql script like
```
with T as (
   select UNIQUE_EVENT_ID as pk from stg_oss__events
) 
select 
   count(*) as row_count, 
   count(pk) as not_null, 
   count(distinct pk) as uniques 
from T
```

recce/dbt.py
```diff
@@ -486,65 +486,90 @@ def refresh(self, refresh_file_path: str = None):
...
+        column_results = []
+        primary_key_result = []
+
+        def log_callback(data, info=False):
+            if isinstance(data, tuple) and len(data) == 4 and 'perfect match' in data[1]:
+                column_results.append(data)
+
+                # full data for primary key to make added and removed
+                if data[0].lower() == primary_key.lower():
+                    primary_key_result.append(data)
```
popcornylu: Case 1: If there are missing value, the value is wrong. 

image
image


Case 2:
If both side are null, it should also mark as matched

Please checks all the 7 cases
```
            when a_query.EVENT_TYPE = b_query.EVENT_TYPE then '✅: perfect match'
            when a_query.EVENT_TYPE is null and b_query.EVENT_TYPE is null then '✅: both are null'
            when a_query.EVENT_ID is null then '🤷: ‍missing from a'
            when b_query.EVENT_ID is null then '🤷: missing from b'
            when a_query.EVENT_TYPE is null then '🤷: value is null in a only'
            when b_query.EVENT_TYPE is null then '🤷: value is null in b only'
            when a_query.EVENT_TYPE != b_query.EVENT_TYPE then '🙅: ‍values do not match'
            else 'unknown' -- this should never happen
```

popcornylu (changes requested):
The value diff result is wrong. Please see the review comments.

playground.py
qrtt1: removed

recce/dbt.py
```diff
@@ -486,65 +486,141 @@ def refresh(self, refresh_file_path: str = None):
...
+                                  log_callback=log_callback,
+                                  base_relation=self.get_base_relation(model),
+                                  pick_columns_to_compare=pick_columns_to_compare)
+                              )
+
+            data = []
+            for k, v in column_groups.items():
+                matched = v['matched']
+                mismatched = v['mismatched']
+                record = [k, matched, 100 * (matched / (matched + mismatched))]
```
popcornylu: Please handle the corner case (matched + mismatched) = 0

popcornylu (commented):
Please fix the only issue. Others LGTM

PR #89, opened 2023-12-26
[Feature] validate Primary Key for value-diff

qrtt1:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

enhancement

value-diff must choose a primary key to compare data, we show the validation errors to invalid primary key.

a valid primary key must be:
* unique
* not null

rc-92

### DIFF
--- a/recce/dbt.py
+++ b/recce/dbt.py
@@ -498,6 +498,7 @@ class DBTContext:
 
     def columns_value_mismatched_summary(self, primary_key: str, model: str):
         column_groups = {}
+        errors = self.verify_primary_key(primary_key, model)
 
         def log_callback(data, info=False):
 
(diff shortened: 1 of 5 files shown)


recce/dbt.py
```diff
@@ -498,6 +498,7 @@ def callback(x):
 
     def columns_value_mismatched_summary(self, primary_key: str, model: str):
         column_groups = {}
+        errors = self.verify_primary_key(primary_key, model)
```
popcornylu, line 501: Please make the validation error failed fast. That is, if the PK is not unque and with null, don't do the following check. It will query wrong result and potentially have at most n^2 records to query. (e.g  in a 1,000 rows table, if we select a all null column, it will get 1,000,000 outer join result)

recce/dbt.py
```diff
@@ -623,6 +624,52 @@ def pick_columns_to_compare(c1, c2):
...
+                errors=errors
             )
             return result
+
+    def verify_primary_key(self, primary_key: str, model: str):
+        errors = []
+
+        def callback(check_name, executor, sql, is_base: bool):
+            table = executor(sql)
+            invalids = len(table.rows)
```
popcornylu, line 648: Please don't pull all the invalids data to local. It would take long time. See how dbt test work

https://app.snowflake.com/gkfbopt/vg68680/#/compute/history/queries/01b13db4-0001-96dd-0001-40ae01fb7a1a/detail?autoRefreshInSeconds=0

It only get the count, rather than individual records.

popcornylu (changes requested):

recce/dbt.py
qrtt1: got if, skip execution when find any errors

recce/dbt.py
qrtt1: update the check by counting

PR #100, opened 2024-01-04
[Feature] Fix and enhance profile task

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- show error when profile failed
- allow to cancel profile task
- fix cannot visualize profile result in snowflake 
- fix cannot visualize boolean value of profile result in data grid

RC-104

NONE


popcornylu (changes requested):
1. Could we preserve the column name cases? It should be all uppercases
2. Could the column name text aligned to left
3. Add to checklist should be only enabled if run is succefully.

image
image

popcornylu (approved):

PR #103, opened 2024-01-05
[Feature] Copy the query diff result as image and save to clipboard

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Save query diff result as an image
- Save the image into clipboard
- Highlight the copied area when mouse hover the copy button

RC-93

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -25,6 +25,7 @@
     "dagre": "^0.8.5",
     "framer-motion": "^10.16.4",
     "html-to-image": "^1.11.11",
+    "html2canvas": "^1.4.1",
     "lodash": "^4.17.21",
     "monaco-editor": "^0.44.0",
     "next": "14.0.1",
--- a/js/pnpm-lock.yaml
+++ b/js/pnpm-lock.yaml
@@ -38,6 +38,9 @@ dependencies:
   html-to-image:
     specifier: ^1.11.11
     version: 1.11.11
+  html2canvas:
+    specifier: ^1.4.1
+    version: 1.4.1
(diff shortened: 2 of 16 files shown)


popcornylu (commented):
It seems that it requires every check view add the screenoshot logic. Could we add this feature as container?

```
<ScreenshotBox>
   <QueryDataGrid />
</ScreenshotBox>
```

kentwelcome:
> It seems that it requires every check view add the screenoshot logic. Could we add this feature as container?
> 
> ```
> <ScreenshotBox>
>    <QueryDataGrid />
> </ScreenshotBox>
> ```

Sure!

popcornylu:
> > It seems that it requires every check view add the screenoshot logic. Could we add this feature as container?
> > ```
> > <ScreenshotBox>
> >    <QueryDataGrid />
> > </ScreenshotBox>
> > ```
> 
> Sure!

Here is the example
https://github.com/DataRecce/recce/pull/106

kentwelcome:
Ultimately, I decided to wrap the <DataGrid> component by <ScreenshotDataGrid>.
The screenshot will only take on the <DataGrid> component. 

Due to the <DataGrid> having its scrolling logic. The header will keep showing on the top when the user scrolls the table.

If we use a <ScreenShotBox> or <Box> component to wrap up the <DataGrid> component, scrolling logic will be running on <Box>. This means the table's header won't be kept on the top of the screen.

When users take a screenshot, keeping the header on the top of the table will make the screenshot more readable.

PR #105, opened 2024-01-08
[Feature] Query diff view allows to filter out unchanged rows or columns

popcornylu:
In the query diff view, we support to show only the changed one.
- We add the “Changed only” toggle, the default value is false, because if it set to true, we cannot select the primary key.
- Add the color of added/removed columns
- Add the color of added/removed rows (according to the PK settings)
- Change the color of cell value changed.
- Move the “Add to check” to be close to the query result.

image
image

## PR Checklist

- [x] DCO signed


wcchang1115 (approved):
LGTM

PR #107, opened 2024-01-09
[Bug] Copy checklist under HTTP

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

- Clipboard API works only under HTTPS, we need another way to do it.

RC-114

NONE


popcornylu:
There is a library [copy-to-clipboard](https://github.com/sudodoki/copy-to-clipboard) to do this. If we will add a lot of copy logic, it worths to use 3rd party library

popcornylu (approved):
LGTM

PR #138, opened 2024-01-30
[Feature] Replace pandas as recce native dataframe presentation.

popcornylu:
The reasons

- The type of dbt connector returns the `agate.Table` type. It is not necessary to transform to pandas dataframe
- In current implementation, the type of columns is inferred by data rather than the [agate data  type](https://agate.readthedocs.io/en/latest/api/data_types.html)
- The {data: [{<key>: <value>, …}...]} format is less efficient to transform or stored in memory. Would like to change to {data: [[v1, v2, v3>, …], …]}
- Reduce the dependency. Pandas is not a must.

**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Enhancement

I remove the pandas dependency from recce. In the CLI, because `recce query` and `recce diff` is hidden, I keep the pandas impelmentation, but will warm the user to install pandas if import failed.

It could be imcompatible with the existing state file. But because we are still in the very early stage, it should be only few people may encounter this issu.e


kentwelcome (approved):
LGTM

PR #172, opened 2024-02-15
[Feature] Add warning when the columns are invalid to be a primary key

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- hint users that the chosen columns are not valid to be the primary key

RC-125

NONE

### DIFF
--- a/js/src/components/query/QueryDiffResultView.tsx
+++ b/js/src/components/query/QueryDiffResultView.tsx
@@ -13,8 +13,9 @@ import {
   IconButton,
   Spacer,
   Tooltip,
+  VStack,
 } from "@chakra-ui/react";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { toDataDiffGrid } from "./querydiff";
 
 import "./styles.css";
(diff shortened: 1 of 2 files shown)


js/src/components/query/QueryDiffResultView.tsx
```diff
@@ -70,10 +76,15 @@ export const QueryDiffResultView = ({
       }
     };
 
+    const handleInvalidPrimaryKey = (base: string[], current: string[]) => {
```
popcornylu: Please use the return value rather than a callback. 

Reason: the invalid primary key is resulted from the input `options.primaryKeys`, but not user's action.

js/src/components/query/QueryDiffResultView.tsx
```diff
@@ -86,6 +97,22 @@ export const QueryDiffResultView = ({
     onViewOptionsChanged,
   ]);
 
+  const invalidPKeyMessage = useMemo(() => {
+    if (invalidPKey.base.length === 0 && invalidPKey.current.length === 0) {
+      return null;
+    }
+    const base = invalidPKey.base.join(", ");
```
popcornylu: why not use `primaryKeys`

js/src/components/query/QueryDiffResultView.tsx
```diff
@@ -119,22 +146,27 @@ export const QueryDiffResultView = ({
         gap="5px"
         alignItems="center"
         px="10px"
-        bg={warning ? "orange.100" : "inherit"}
+        bg={warning || invalidPKeyMessage ? "orange.100" : "inherit"}
```
popcornylu: Please use the same variable naming

warning -> waringLimit
invalidPKeyMessage -> warningPK

PR #210, opened 2024-03-06
[Bug] Show the first check of the loaded checklist

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

- After loading a state file, it should select the first loaded check.

DRC-263

NONE


popcornylu (changes requested):
In the current implementation, when i load a state file, the left selection panel would not be selectable.

PR #288, opened 2024-04-22
[Bug] State file type validation rule

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

- In review mode, recce server would read state file and do basic type validation.

DRC-387

NONE

### DIFF
--- a/js/src/lib/api/info.ts
+++ b/js/src/lib/api/info.ts
@@ -111,7 +111,7 @@ export interface gitInfo {
 }
 
 export interface pullRequestInfo {
-  id?: string;
+  id?: string | number;
   title?: string;
   url?: string;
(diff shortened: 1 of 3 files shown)


js/src/lib/api/info.ts
```diff
@@ -111,7 +111,7 @@ export interface gitInfo {
 }
 
 export interface pullRequestInfo {
-  id?: string;
+  id?: string | number;
```
popcornylu, line 114: It is only available since 3.10. Please use Union instaed

popcornylu (changes requested):

popcornylu (approved):

PR #299, opened 2024-05-02
[Feature] Support to run query diff on the data warehouse

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- allow to do query diff on data warehouse

DRC-375


js/src/components/query/querydiff.tsx
```diff
@@ -437,3 +437,240 @@ export function toDataDiffGrid(
     invalidPKeyCurrent,
   };
 }
+
+function _getJoinColumnMap(df: DataFrame) {
```
popcornylu: Is the logic copied from value_diff? If yes, could we share the same code?

recce/models/types.py
```diff
@@ -10,6 +10,7 @@ class RunType(Enum):
     SIMPLE = 'simple'
     QUERY = "query"
     QUERY_DIFF = 'query_diff'
+    QUERY_DIFF_JOIN = 'query_diff_join'
```
popcornylu: I remember that in the design, we should share the same run type 'query_diff'

js/src/components/query/QueryPage.tsx
```diff
@@ -138,12 +157,19 @@ export const QueryPage = () => {
             onViewOptionsChanged={setViewOptions}
             onCancel={handleCancel}
           >
-            {(props) => (
-              <QueryDiffResultView
-                {...props}
-                onAddToChecklist={addToChecklist}
-              />
-            )}
+            {(props) =>
```
popcornylu: Please merge the two components to one. So that the check page can use the single QueryDiffResultView

popcornylu (changes requested):

popcornylu (approved):
LGTM

PR #307, opened 2024-05-07
[Feature] Support to parse query diff with joins result when doing summary

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- The query diff now has two different result formats, so the recce summary needs to support both.

DRC-411

NONE

### DIFF
--- a/recce/tasks/query.py
+++ b/recce/tasks/query.py
@@ -236,5 +236,19 @@ class QueryDiffResultDiffer(TaskResultDiffer):
     def _check_result_changed_fn(self, result):
         base = result.get('base')
         current = result.get('current')
+        diff = result.get('diff')
 
-        return TaskResultDiffer.diff(base, current)
+        if diff is None:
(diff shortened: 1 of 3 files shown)


recce/tasks/query.py
```diff
@@ -233,5 +233,19 @@ class QueryDiffResultDiffer(TaskResultDiffer):
     def _check_result_changed_fn(self, result):
         base = result.get('base')
         current = result.get('current')
+        diff = result.get('diff')
 
-        return TaskResultDiffer.diff(base, current)
+        if diff is None:
+            return TaskResultDiffer.diff(base, current)
+        else:
+            base = dict(columns=[], data=[])
```
popcornylu: Would it better to just check the row count of `diff` dataframe. If the size is not zero, there is diff?

popcornylu (approved):
LGTM

PR #310, opened 2024-05-08
[Feature] Support changed only toggling on run toolbar

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- If there is no change, it shows no change and cannot toggle back when toggle the "Changed only" checkbox

DRC-401


popcornylu (changes requested):
If there are warnings, should always show the warnings in the toolbar.

popcornylu (approved):

PR #404, opened 2024-08-20
Resetting invocation_context on generate_sql function when dbt_version > 1.8

jefersonmsantos:
Fix [405](https://github.com/DataRecce/recce/issues/405)

On new dbt 1.8 version there is error  LookupError: <ContextVar name='DBT_INVOCATION_CONTEXT_VAR' at 0x10c4889f0>   when using function "Run query" or "Run diff". This happens because class DbtAdapter sets the invocation context with function set_invocation_context({}) when dbt version is 1.8, but on tasks/query.py file the default_context is regenererated before calling function generate_sql from DbtAdapter.

image
image

To fix this, added a new call of set_invocation_context({}) on start of function generate_sql if dbt version is greater or equal than 1.8.

**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug fix

Added a new call of set_invocation_context({}) on start of function generate_sql if dbt version is greater or equal than 1.8 to fix the issue described below.

With the fix, `Run Diff` works fine with dbt version 1.8:

image

Tests:
image

Fix [405](https://github.com/DataRecce/recce/issues/405)
When using new dbt 1.8 version, `Run Diff` feature will throw error `LookupError: <ContextVar name='DBT_INVOCATION_CONTEXT_VAR' at 0x10c4889f0>`

This happens because class DbtAdapter sets the invocation context with function set_invocation_context({}) when dbt version is 1.8, but on tasks/query.py file the default_context is regenererated before calling function generate_sql from DbtAdapter.

image

"NONE"

### DIFF
--- a/recce/adapter/dbt_adapter/__init__.py
+++ b/recce/adapter/dbt_adapter/__init__.py
@@ -406,7 +406,7 @@ class DbtAdapter(BaseAdapter):
         if dbt_version >= dbt_version.parse('v1.8'):
             from dbt_common.context import set_invocation_context
             set_invocation_context({})
-            
+
         node_id = str("generated_" + uuid.uuid4().hex)
         node = parser.parse_remote(sql_template, node_id)
         process_node(self.runtime_config, manifest, node)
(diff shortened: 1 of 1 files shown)


popcornylu:
Thanks for your PR. 
image
Could you please address the flake8 issue? Everything else looks good to me!

jefersonmsantos:
> Thanks for your PR.
> image
> Could you please address the flake8 issue? Everything else looks good to me!

Addressed.
image

popcornylu (approved):

PR #529, opened 2024-12-09
[Bug] DRC-951 dbt run defer could make the server cannot launch

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bugfix

- Change to use `get_lineage` instead of `get_model` as the source to judge
  whether the schema diff has changed.
- When the data source is Snowflake, `get_model` method will trigger the DB
  connection. In this case, we don't need to connect to DB. The manifest
  file already contains all the information we need.
- Only `recce run` will trigger preset check pre-approved logic. This logic won't be invoked when launching the server.

### DIFF
--- a/recce/run.py
+++ b/recce/run.py
@@ -57,7 +57,10 @@ def schema_diff_should_be_approved(check_params: dict) -> bool:
 
         if 'node_id' in check_params:
             # If the node_id is provided, then use it
-            selected_node_ids = check_params.get('node_id', [])
+            if isinstance(check_params['node_id'], str):
+                selected_node_ids = [check_params['node_id']]
+            else:
+                selected_node_ids = check_params.get('node_id', [])
         else:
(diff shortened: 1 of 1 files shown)


recce/run.py
```diff
@@ -45,34 +45,47 @@ def load_preset_checks(checks: list):
...
-
-    # If the diff is empty, then the check should be approved
-    if bool(diff) is False:
-        return True
+    try:
+        context = default_context()
+
+        if 'node_id' in check_params:
+            # If the node_id is provided, then use it
+            selected_node_ids = check_params.get('node_id', [])
```
popcornylu: `node_id` can be `string` or `string[]`


`string` if it is added from NodeView
`string[]` if it is added from multi-node selection

kentwelcome: fix

popcornylu (approved):

PR #541, opened 2024-12-17
Add command options to command events

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
- Chore

We want to collect some command arguments to understand user behavior.
In this PR, we add
- cloud (True/False)
- review (True/False)
- debug (True/False)
- hashed target_path: we want to know if people don't use the default value
- hashed target_base_path: we want to know if people don't use the default value

- PLA-213

### DIFF
--- a/recce/event/track.py
+++ b/recce/event/track.py
@@ -11,7 +11,7 @@ from click.core import Command
 from rich.console import Console
 from rich.markup import escape
 
-from recce import event, get_runner
+from recce import event, get_runner, get_cli_args()
 from recce.core import load_context
 from recce.exceptions import RecceException
 from recce.git import current_branch, hosting_repo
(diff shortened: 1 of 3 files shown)


recce/event/track.py
```diff
@@ -109,6 +114,12 @@ def invoke(self, ctx: Context) -> t.Any:
             if branch is not None:
                 props['branch'] = sha256(branch.encode()).hexdigest()
 
+            if target_path is not None:
```
wcchang1115, line 117: Hi even,

The `target_path` and `target_base_path` here contain default values `target` and `target-base`.
So you would still need to compare with the `hash("target")` and `hash("target-base")` value to check if the customed path was given.

Is it what you expect? If yes, then it's all good to me.

wcchang1115 (approved):
LGTM

recce/event/track.py
even-wei: Yes, I want to know if it's different and how many different kind of different name people use, but I don't want to collect sensitive data.

So I hash it and hash the default value while analyzing it.

PR #560, opened 2024-12-26
[Fix] DRC-979 Allow Recce Server to launch without target-base

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Allow `recce server` to launch without `target-base` folder
- Show a warning message if the `target-base` folder not provide
CleanShot 2024-12-26 at 17 14 08

- Enable `onboarding-mode` as True in server API `/api/flag` if the server launch without `target-base` folder
CleanShot 2024-12-26 at 17 15 11

- Add unit-test for this case

### DIFF
--- a/recce/adapter/base.py
+++ b/recce/adapter/base.py
@@ -10,6 +10,11 @@ logger = logging.getLogger('uvicorn')
 
 
 class BaseAdapter(ABC):
+
+    @classmethod
+    def load(cls, **kwargs):
+        raise NotImplementedError()
+
     @abstractmethod
     def get_lineage(self, base: Optional[bool] = False):
         raise NotImplementedError()
--- a/recce/adapter/dbt_adapter/__init__.py
+++ b/recce/adapter/dbt_adapter/__init__.py
@@ -148,6 +148,10 @@ def previous_state(state_path: Path, target_path: Path, project_root: Path) -> P
     if dbt_version < 'v1.5.2':
         return PreviousState(state_path, target_path)
     else:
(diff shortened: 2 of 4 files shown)


wcchang1115 (commented):
Hi Kent,
I found there's a warning from dbt.
`Warning: The state and target directories are the same: 'target'. ...`
I think we don't "overwrite state", so we don't need to provide the message.

I found that overwriting the log level can dismiss the message.
(but I only have a quick test in dbt 1.8, not sure if it works in other versions)
```
def previous_state() -> PreviousState:
    WarnStateTargetEqual.level_tag = lambda x: EventLevel.DEBUG
    return PreviousState(state_path, target_path, project_root)
```
Do you think is it worth trying?

kentwelcome:
> Hi Kent, I found there's a warning from dbt. `Warning: The state and target directories are the same: 'target'. ...` I think we don't "overwrite state", so we don't need to provide the message.
> 
> I found that overwriting the log level can dismiss the message. (but I only have a quick test in dbt 1.8, not sure if it works in other versions)
> 
> ```
> def previous_state() -> PreviousState:
>     WarnStateTargetEqual.level_tag = lambda x: EventLevel.DEBUG
>     return PreviousState(state_path, target_path, project_root)
> ```
> 
> Do you think is it worth trying?

This warning message will likely happen when dbt> = 1.6.
Let's try to hide this message. This experiment is focused on the beginner of recce.
Throwing unexpected warning messages could decrease the confidence of our users.

wcchang1115 (approved):
LGTM, thanks!

PR #564, opened 2024-12-31
[Feature] DRC-1001 Hiden diff actions when signel env mode

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Hiden the diff action buttons and Query&Check tabs when server is launched with single environment mode
CleanShot 2024-12-31 at 15 20 44
CleanShot 2024-12-31 at 15 21 48

### DIFF
--- a/js/app/page.tsx
+++ b/js/app/page.tsx
@@ -253,7 +253,7 @@ function NavBar() {
   const { isDemoSite, reviewMode, fileMode, cloudMode, isLoading } =
     useLineageGraphContext();
   const [location, setLocation] = useLocation();
-  const { data: flag, isLoading: flagLoading } = useRecceServerFlag();
+  const { data: flag, isLoading: isFlagLoading } = useRecceServerFlag();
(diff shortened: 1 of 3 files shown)


js/app/page.tsx
```diff
@@ -274,22 +286,24 @@ function NavBar() {
...
               >
                 {name}
                 {badge}
               </Tab>
             );
           })}
         </Box>
-        {!isLoading && !isDemoSite && <Filename />}
-        {!isLoading && (
+        {!isLoading && !isFlagLoading && !isDemoSite && <Filename />}
```
wcchang1115: I'm not sure do you need `isFlagLoading` here?

wcchang1115 (approved):

js/app/page.tsx
kentwelcome: You are right! I think we don't need to check `isFlagLoading` here.

PR #571, opened 2025-01-08
[Chore] Improve is_ci function

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Chore

- Remove `isatty` from the `is_ci` function.
- Add common CI environments to identify

DRC-1016

### DIFF
--- a/recce/__init__.py
+++ b/recce/__init__.py
@@ -21,7 +21,7 @@ def is_ci_env():
 
     for env_var, expected_value in ci_environments.items():
         env_value = os.environ.get(env_var)
-        if env_value:
+        if env_value is not None:
             # If we just need the variable to exist
             if expected_value is None:
                 return True
--- a/recce/__init__.py
+++ b/recce/__init__.py
@@ -1,19 +1,33 @@
 import os
-import sys
 
 
 def is_ci_env():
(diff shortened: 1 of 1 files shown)


recce/__init__.py
```diff
@@ -1,19 +1,33 @@
...
+        'TEAMCITY_VERSION': None,       # TeamCity
+        'BITBUCKET_COMMIT': None,       # Bitbucket Pipelines
+        'BUILDKITE': 'true',            # Buildkite
+        'CODEBUILD_BUILD_ID': None,     # AWS CodeBuild
+        'AZURE_PIPELINES': 'true',      # Azure Pipelines
+    }
+
+    for env_var, expected_value in ci_environments.items():
+        env_value = os.environ.get(env_var)
+        if env_value:
```
wcchang1115: should we change this truthy check to `if env_value is not None` in the case that we only need the variable to exist? e.g., `export VAR=""`

wcchang1115 (changes requested):

recce/__init__.py
even-wei: Good point

wcchang1115 (approved):
LGTM, Thanks!

PR #574, opened 2025-01-09
[Feature]support breaking change analysis

popcornylu:
This PR support he breaking change analysis.

For modified nodes, it will use raw(templated) sql code. A node would mark as non-breaking change if
1. The sql sementic tree is identical
2. Or only new column added
Otherwise, it would mark as breaking change.

For non-breaking change, it will not mark the downstream as impacted. See the screenshot

image

**PR checklist**

- [x] DCO signed

**What type of PR is this?**


kentwelcome (approved):
LGTM

PR #576, opened 2025-01-10
[Feature] Overwrite response 304 when etag not match

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Instead of disable cache with `no-store`, we overwrite the 304 response when the etag is not matched.

DRC-995

NONE


wcchang1115:
After discussion, we prefer not to overwrite the cache behavior.
Just use `no-store` to disable cache.

PR #578, opened 2025-01-13
[Fix] DRC-1031 Under single env mode, detect manifest changed issue

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bugfix

- We should reload both the base and current artifacts.
- Currently, we only update the current artifact. The base artifact will not be changed.


wcchang1115 (approved):
LGTM, thanks!

PR #595, opened 2025-02-03
Fix: Primary key dropdown selector takes too long to populate

### LINKED ISSUE #594: [DRC-1058] [Bug] Primary key dropdown selector takes too long to populate when there are many columns
### Current Behavior

The primary key dropdown selector freezes the browser and repeatedly triggers the "Wait or Exit" dialog when there are many columns. It also searches the full project instead of the model(s) relevant to the query/comparison.

Image

Image

### Expected Behavior

The dropdown should be fast. It would be better to limit the list to columns from the relevant model(s) or use a lazy or type-ahead search that retrieves columns only after the user enters a few characters.

### Steps To Reproduce

1. Run the recce server with a decent-sized project (it can't be too small)
2. Click on the primary key dropdown -> this will hang the browser and eventually prompt to "wait" or "exit page", usually multiple times.

### Relevant log output

```shell

```

### Environment

- recce: 0.51.3
- OS: WSL 2 on Windows 100
- Python: 3.12.5
- Data Warehouse: Snowflake
- dbt: 1.9.2 (core)

### Additional Context

_No response_

<sub>[DRC-1058](https://linear.app/recce/issue/DRC-1058/[bug]-primary-key-dropdown-selector-takes-too-long-to-populate-when)</sub>

popcornylu:
Fix #594

**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug fix

The hand craft dropdown menu would show items. It would impact the performance of frontend.

#594

### DIFF
--- a/js/src/utils/DropdownValuesInput.tsx
+++ b/js/src/utils/DropdownValuesInput.tsx
@@ -86,7 +86,7 @@ export const DropdownValuesInput = (props: DropdownValuesInputProps) => {
             backgroundColor={"white"}
           />
           {values.length === 0 && (
-          <InputRightElement>
+            <InputRightElement>
               <Icon
                 as={FaChevronDown}
                 color="blue.500"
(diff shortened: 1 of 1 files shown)


kentwelcome (approved):
LGTM

PR #600, opened 2025-02-10
[Fix] DRC-1072 Single env mode should not effect when review enabled

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug fix

- Single environment mode should not work when the review mode is enabled


wcchang1115 (approved):
LGTM, Thanks!

PR #620, opened 2025-02-21
[Fix] DRC-1092 Skip auto-focus logic when some specific run types

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bugfix

- Skip auto focus logic when the run type is `query`, `query_diff` or `row_count`
- Execute `Run Query` won't break the `Sandbox` or `Single Env Query` Modal

### DIFF
--- a/js/src/components/lineage/LineageView.tsx
+++ b/js/src/components/lineage/LineageView.tsx
@@ -601,7 +601,7 @@ export function PrivateLineageView(
     }
     if (
       !runResultType ||
-      ["query_diff", "query", "row_count"].find((x) => x === runResultType)
+      ["query_diff", "query", "row_count"].includes(runResultType)
     ) {
       // Skip the following logic if the run result type is not related to a node
       return;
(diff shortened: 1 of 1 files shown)


js/src/components/lineage/LineageView.tsx
```diff
@@ -598,6 +599,14 @@ export function PrivateLineageView(
       // Skip the following logic if the run result is not open
       return;
     }
+    if (
+      !runResultType ||
+      ["query_diff", "query", "row_count"].find((x) => x === runResultType)
```
wcchang1115: Hi Kent,

Would it be better to use `[xxx].includes`?
We could get a boolean directly.

wcchang1115 (approved):
Just found a tiny improvement.
Thank you!

js/src/components/lineage/LineageView.tsx
kentwelcome: Thx suggestion!

PR #657, opened 2025-03-14
[Feature] DRC-1105 Notify user Recce have a new version to update

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
UX enhance

- The notification will be shown on both CLI console and web page.
  - CLI notification
CleanShot 2025-03-14 at 14 55 37
  - Frontend notification 
Google Chrome 2025-03-14 14 52 23

### DIFF
--- a/js/app/page.tsx
+++ b/js/app/page.tsx
@@ -6,6 +6,7 @@ import {
   TabList,
   Tab,
   ChakraProvider,
+  Code,
   Box,
   Flex,
   Link,
@@ -17,6 +18,7 @@ import {
   Progress,
   HStack,
   extendTheme,
+  useToast,
 } from "@chakra-ui/react";
 import React, { ReactNode, useLayoutEffect } from "react";
 import { QueryClientProvider, useQuery } from "@tanstack/react-query";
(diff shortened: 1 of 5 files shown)


js/app/page.tsx
```diff
@@ -220,7 +279,7 @@ function NavBar() {
                 onClick={() => {
                   setLocation(href);
                 }}
-                isDisabled={isLoading || isFlagLoading || disable}
+                isDisabled={(isLoading ?? isFlagLoading) || disable}
```
wcchang1115: Hi Kent,

I think we should not refer `isFlagLoading` when `isLoading` is undefined.
Should we chang to `{isDisabled={!!isLoading || isFlagLoading || disable}}` to fix the lint warning?

recce/__init__.py
```diff
@@ -55,4 +58,17 @@ def get_version():
         return version
 
 
+def fetch_latest_version():
+    try:
+        url = 'https://pypi.org/pypi/recce/json'
+        response = requests.get(url)
```
wcchang1115: I think we can add a timeout here. What do you think?
https://requests.readthedocs.io/en/stable/user/advanced/#timeouts

wcchang1115 (changes requested):

js/app/page.tsx
kentwelcome: Fixed!

recce/__init__.py
kentwelcome: Configure 3 3-second timeout. And if the version contains `dev`, it won't trigger fetch version mechanism.

wcchang1115 (approved):
LGTM, thanks!

PR #674, opened 2025-03-28
[Chore] refactor the LineageView

popcornylu:
This is a refactoring ticket. It also fix some problems
- The select/unselect CLL will break the impact radius for non-breaking change
- The focus node should not re-center the node.

Code base  enhacement
1. Remove the view state isSelected`, `isHighlighted` in the `LinageGraphNode`. Use the LinageViewContext instead.
2. Minimum the state in the LinaegeView

**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Chore + Bug fix


js/src/components/lineage/LineageView.tsx
```diff
@@ -735,15 +735,17 @@ export function PrivateLineageView(
...
+      const newSelectedNodeIds = new Set(selectedNodeIds);
+      if (selectedNodeIds.has(nodeId)) {
+        newSelectedNodeIds.delete(nodeId);
+      } else {
+        newSelectedNodeIds.add(nodeId);
+      }
+
+      setSelectedNodeIds(newSelectedNodeIds);
+      if (newSelectedNodeIds.size === 0) {
+        setSelectMode("single");
```
wcchang1115, line 765: Does `single` mode work like `init` mode now?
(not a change request, but would like to know the usage of `single` mode after the PR)

wcchang1115 (approved):
LGTM, thanks!

js/src/components/lineage/LineageView.tsx
popcornylu: I think it should also change. I will create another PR to address this problem.

PR #694, opened 2025-04-17
[Feature] Refine UI language

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Enhancement

- Refine UI language

### DIFF
--- a/js/src/components/run/RunResultPane.tsx
+++ b/js/src/components/run/RunResultPane.tsx
@@ -20,7 +20,12 @@ import {
   MenuList,
   MenuItem,
   MenuDivider,
-  Tooltip,
+  PopoverBody,
+  PopoverTrigger,
+  PopoverContent,
+  Popover,
+  Link,
+  Portal,
 } from "@chakra-ui/react";
 import { useCallback, useState } from "react";
 import { createCheckByRun } from "@/lib/api/checks";
@@ -41,6 +46,7 @@ import { LearnHowLink, RecceNotification } from "../onboarding-guide/Notificatio
 import { useRecceInstanceContext } from "@/lib/hooks/RecceInstanceContext";
 import { TbCloudUpload } from "react-icons/tb";
 import { useRecceShareStateContext } from "@/lib/hooks/RecceShareStateContext";
+import { PUBLIC_CLOUD_WEB_URL } from "@/lib/const";
 
 interface RunPageProps {
   onClose?: () => void;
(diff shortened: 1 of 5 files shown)


kentwelcome:
UI of Share button
CleanShot 2025-04-18 at 10 56 20

kentwelcome:
UI of the Share Button on Run Result view
CleanShot 2025-04-18 at 10 57 27

js/src/components/run/RunResultPane.tsx
```diff
@@ -144,16 +150,32 @@ const RunResultShareMenu = ({
...
-              <Button
-                leftIcon={<ExternalLinkIcon />}
-                size="sm"
-                onClick={() => {
-                  window.open("https://cloud.datarecce.io/settings#tokens", "_blank");
-                }}>
-                Enable sharing
-              </Button>
-            </Tooltip>
+            <Popover trigger="hover" placement="bottom-start">
```
kentwelcome: The component `<MenuItem>` is already clickable; we don't need to add a button inside the MenuItem.

kentwelcome:
The components `<RunResultShareMenu>` and `<TopLevelShare>` will handle both conditions when the `api_token` is set or not. 
However, both of those components will be rendered only when `authed` is true. Which means the `api_token` should be provided.
https://github.com/DataRecce/recce/blob/b516928c5d1fe5b3ed6fd2eb51fa9002cef35e74/js/src/components/run/RunResultPane.tsx#L296
https://github.com/DataRecce/recce/blob/b516928c5d1fe5b3ed6fd2eb51fa9002cef35e74/js/app/page.tsx#L297
In this case, the condition to show the `Enable sharing` button will never happen.

kentwelcome (changes requested):
The components `<RunResultShareMenu>` and `<TopLevelShare>` will handle both conditions when the `api_token` is set or not. 
However, both of those components will be rendered only when `authed` is true. Which means the `api_token` should be provided.
https://github.com/DataRecce/recce/blob/b516928c5d1fe5b3ed6fd2eb51fa9002cef35e74/js/src/components/run/RunResultPane.tsx#L296
https://github.com/DataRecce/recce/blob/b516928c5d1fe5b3ed6fd2eb51fa9002cef35e74/js/app/page.tsx#L297
In this case, the condition to show the `Enable sharing` button will never happen.

js/src/components/run/RunResultPane.tsx
kentwelcome: You can reference what we do in the `Explore` menu button.
https://github.com/DataRecce/recce/blob/b516928c5d1fe5b3ed6fd2eb51fa9002cef35e74/js/src/components/lineage/NodeView.tsx#L101

wcchang1115: Hi Kent,
As discussion, I will have another story for showing the Enable sharing button.
And I removed `Button` and used plain `MenuItem`.

kentwelcome (approved):
LGTM

PR #723, opened 2025-05-05
[Feature] Replace dbt-audit-helper with Jinja template

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

This PR replaces the dbt-audit_helper macro with Jinja templates.
We use only a few macros, so we replace them to remove the dependency.

DRC-1268

- Remove checking dbt-audit-helper
- Remove checking dbt-audit-helper in test
- Replace original macro with Jinja templates in value-diff and query-diff
- Remove checking in Recce UI

### DIFF
--- a/recce/tasks/query.py
+++ b/recce/tasks/query.py
@@ -19,11 +19,8 @@ if typing.TYPE_CHECKING:
 class QueryMixin:
     @classmethod
     def execute_sql_with_limit(
-        cls,
-        sql_template,
-        base: bool = False,
-        limit: Optional[int] = None
-    ) -> Tuple['agate.Table', bool]:
+        cls, sql_template, base: bool = False, limit: Optional[int] = None
+    ) -> Tuple["agate.Table", bool]:
         """
         Execute a SQL template and return the result as an agate table.
         :param sql_template: SQL template to execute
@@ -32,8 +29,10 @@ class QueryMixin:
         :return: Tuple of agate table and whether there are more rows to fetch
         """
         from jinja2.exceptions import TemplateSyntaxError
+
         dbt_adapter = default_context().adapter
         from dbt.exceptions import TargetNotFoundError
+
         try:
             sql = dbt_adapter.generate_sql(sql_template, base)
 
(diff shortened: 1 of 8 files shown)


popcornylu:
After discussion, some changes need to address
1. Remove the `dbt_utils` depedency
2. Use sqlglot to [transpile](https://github.com/tobymao/sqlglot?tab=readme-ov-file#formatting-and-transpiling) to target warehouse.

recce/tasks/query.py
```diff
@@ -139,83 +148,148 @@ def __init__(self, params):
...
+                primary_key=primary_key,
+                limit=QUERY_LIMIT,
+            ),
+        )
+
+        adapter_name = dbt_adapter.runtime_config.credentials.type
+        sql = transpile(
+            sql, read="duckdb", write=adapter_name, identify=True, pretty=True
+        )[0]
+        print(sql)
```
popcornylu: Could we remove the print statement?

recce/tasks/query.py
```diff
@@ -139,83 +148,148 @@ def __init__(self, params):
...
             self.check_cancel()
 
-            if self.legacy_surrogate_key:
-                new_primary_key = 'dbt_utils.surrogate_key(primary_key)'
-            else:
-                new_primary_key = 'dbt_utils.generate_surrogate_key(primary_key)'
+            primary_key_str = " || '-' || ".join(
+                [f"coalesce(cast({p} as TEXT), '')" for p in primary_keys]
+            )
+            primary_key = f"md5(cast({primary_key_str} as TEXT))"
```
popcornylu: The `primary_key` is only used in the order by. Do we still need to use surrogate key? Or just use

```
...
order by pk1, pk2, in_a desc, in_b desc
```

popcornylu (changes requested):
Could we change the way to use the primary key?

recce/tasks/query.py
even-wei: Removed, thx

recce/tasks/query.py
even-wei: Totally makes sense. Modified.

recce/tasks/query.py
```diff
@@ -139,83 +148,140 @@ def __init__(self, params):
...
+            context=dict(
+                base_query=base_query,
+                current_query=current_query,
+                primary_key=primary_key,
+                limit=QUERY_LIMIT,
+            ),
+        )
+
+        adapter_name = dbt_adapter.runtime_config.credentials.type
+        sql = transpile(
```
popcornylu: Remove sqlglot because we cannot guarantee the custom sql from user is duckdb compatible

popcornylu (changes requested):
Per discussion. 

1. Remove sqlglot
2. Make sure that the sql code is compatible in most of the warehouse.

PR #733, opened 2025-05-09
[Feature] Replace dbt profiler with jinja template

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

Decouple dbt-profiler dependency by replacing the macro with a rebuilt Jinja template.

DRC-1274

### DIFF
--- a/js/src/components/lineage/LineageViewTopBar.tsx
+++ b/js/src/components/lineage/LineageViewTopBar.tsx
@@ -347,7 +347,6 @@ const ControlItem = (props: {
 export const LineageViewTopBar = () => {
   const { deselect, focusedNode, selectedNodes, ...lineageViewContext } =
     useLineageViewContextSafe();
-  const { isActionAvailable } = useLineageGraphContext();
   const { readOnly } = useRecceInstanceContext();
   const { data: flags } = useRecceServerFlag();
   const isSingleEnvOnboarding = flags?.single_env_onboarding;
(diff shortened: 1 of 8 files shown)


popcornylu (commented):
In the current implementation, it leverate the dbt adapter to solve the compatible issue. However, it would be hard to do unittest.

My recommendataion
1. For type mapping, (map type string to number,text,bool,array, struct), use the python code
2. For quoting, casting, use the sqlglot transpile. In the profile diff, there is not user custom query code, so it would be easier to build the sql totally by our own sql code
3. Separate the sql generate code to indivdual function and unittest it with different warehouses.  The testing could be just sqlglot sql parsable.

popcornylu:
```
class DummyAdapter:
    def quote(self, col):
        return f'"{col}"'

class DummyDbt:
    def type_bigint(self):
        return "bigint"

    def type_numeric(self):
        return "numeric"

    def type_string(self):
        return "varchar"
        
db_types = ["bigquery", "postgres", "redshift", "athena", "sqlserver"]

for db_type in db_types:
    context = {
        "column_type": "int64",
        "column_name": "some_column",
        "db_type": db_type,
        "relation": "my_table",
        "adapter": DummyAdapter(),
        "dbt": DummyDbt(),
    }

    sql = Template(sql_template).render(context)
    dialect = db_type
    parse_one(sql, read=dialect)
```

popcornylu (approved):
LGTM

PR #735, opened 2025-05-13
[Feature] DRC-1272 Show restart button when the end of share instance

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Show the `restart` button when the Share Instance has expired
Google Chrome 2025-05-13 16 08 01
- The `share_url` will be provided by EnvVar `RECCE_SHARE_URL` when execute `read-only` command

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -20,7 +20,6 @@ from recce.util.recce_cloud import (
     RecceCloudException,
     get_recce_cloud_onboarding_state,
 )
-
 from .core import RecceContext, set_default_context
 from .event.track import TrackCommand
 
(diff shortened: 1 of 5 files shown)


recce/cli.py
```diff
@@ -1021,6 +1020,8 @@ def share(state_file, **kwargs):
 @click.option("--host", default="localhost", show_default=True, help="The host to bind to.")
 @click.option("--port", default=8000, show_default=True, help="The port to bind to.", type=int)
 @click.option("--lifetime", default=0, show_default=True, help="The lifetime of the server in seconds.", type=int)
+@click.option("--share-url", help="The share URL triggers this instance.", type=click.STRING, hidden=True,
```
wcchang1115: Hi Kent,
Because we have made `read-only` command hidden, I think we don't need the option hidden there.
What do you think?

wcchang1115 (approved):
Only a minor question.
Others are good to me, thanks!

recce/cli.py
kentwelcome: I will remove the hidden flag of `--share-url`

PR #738, opened 2025-05-16
[Feature] parse the model to column dependency

popcornylu:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
New feature

This is the CLL enhancement features. 

Now when we parse a sql, we can understand
1. For all columns, the transformation type and upstream columns
2. (⭐  New) the upstream columns

This PR has not change the API and frontend.

### DIFF
--- a/recce/util/cll.py
+++ b/recce/util/cll.py
@@ -134,7 +134,25 @@ def cll_old(sql, schema=None, dialect=None) -> Dict[str, ColumnLevelDependencyCo
     return result.columns
 
 
-def _cll_union_scope(scope: Scope, scope_cll_map: dict[Scope, CllResult]) -> CllResult:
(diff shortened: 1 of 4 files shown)


recce/util/cll.py
```diff
@@ -91,94 +80,237 @@ class ColumnLevelDependencyColumn:
...
-def cll(sql, schema=None, dialect=None) -> Dict[str, ColumnLevelDependencyColumn]:
-    # given a sql, return the columns depends on
+            return None
+
+    for proj in scope.expression.selects:
+        type = "source"
+        column_depends_on: List[ColumnLevelDependsOn] = []
+
+        # instance of Column
+        if isinstance(proj, exp.Alias):
```
wcchang1115: Hi Pop, 
Do we still need to check `proj` type here?
It seems that we have assumed `proj` is `exp.Alias` here, otherwise later `root.walk` could have a problem.

PR #739, opened 2025-05-19
[Feature] Mark onboarding state as `installed` when api-token is set

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

- Update the option description of `--cloud-token` specific for GitHub
- Verify the api-token with Recce Cloud when it's provided by option
- Store the user provied api-token to recce user profile if it's valided
- Update user's onboarding_state to `installed` when user provided api-token

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -9,17 +9,15 @@ import uvicorn
 from recce import event
 from recce.artifact import download_dbt_artifacts, upload_dbt_artifacts
 from recce.config import RECCE_CONFIG_FILE, RECCE_ERROR_LOG_FILE, RecceConfig
-from recce.event import get_recce_api_token, update_user_profile
 from recce.git import current_branch, current_default_branch
 from recce.run import check_github_ci_env, cli_run
 from recce.state import RecceCloudStateManager, RecceShareStateManager, RecceStateLoader
 from recce.summary import generate_markdown_summary
+from recce.util.api_token import prepare_api_token
 from recce.util.logger import CustomFormatter
 from recce.util.recce_cloud import (
-    RECCE_CLOUD_API_HOST,
     RecceCloudException,
-    get_recce_cloud_onboarding_state,
-)
+    get_recce_cloud_onboarding_state, )
 from .core import RecceContext, set_default_context
 from .event.track import TrackCommand
 
@@ -102,7 +100,8 @@ recce_options = [
 
 recce_cloud_options = [
     click.option("--cloud", is_flag=True, help="Fetch the state file from cloud."),
-    click.option("--cloud-token", help="The token used by Recce Cloud.", type=click.STRING, envvar="GITHUB_TOKEN"),
+    click.option("--cloud-token", help="The GitHub token used by Recce Cloud.", type=click.STRING,
+                 envvar="GITHUB_TOKEN"),
     click.option(
         "--state-file-host",
         help="The host to fetch the state file from.",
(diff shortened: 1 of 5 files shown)


recce/util/api_token.py
```diff
@@ -0,0 +1,69 @@
...
+            if not valid:
+                raise RecceConfigException("Invalid Recce Cloud API token")
+            update_recce_api_token(api_token)
+            console.print("[[green]Success[/green]] Update the user profile for Recce Cloud API Token.")
+
+    if api_token:
+        cloud_onboarding_state = get_recce_cloud_onboarding_state(api_token)
+        if cloud_onboarding_state == "new":
+            # Mark the onboarding state as "installed" if the user is new
+            set_recce_cloud_onboarding_state(api_token, "installed")
```
wcchang1115, line 67: Hi Kent,
Does "installed" here mean the user has signed up for Recce cloud?
Do you think we should reserve this word for installing Recce APP?

recce/util/api_token.py
```diff
@@ -0,0 +1,69 @@
...
+):
+    """
+    Prepare the API token for the request.
+    """
+    # Verify the API token for Recce Cloud Share Link
+    api_token = get_recce_api_token()
+    new_api_token = kwargs.get("api_token")
+    if api_token != new_api_token and new_api_token is not None:
+        # Handle the API token provided by option `--api-token`
+        valid = RecceCloud(new_api_token).verify_token()
```
wcchang1115, line 36: Hi Kent,
Because I saw you have added an attribute `token_type` when `RecceCloud` initialization.
Do you think it's a good idea to tell users when we detect they used their GitHub token before verification?

recce/cli.py
```diff
@@ -249,7 +253,9 @@ def diff(sql, primary_keys: List[str] = None, keep_shape: bool = False, keep_equ
 @click.option("--port", default=8000, show_default=True, help="The port to bind to.", type=int)
 @click.option("--lifetime", default=0, show_default=True, help="The lifetime of the server in seconds.", type=int)
 @click.option("--review", is_flag=True, help="Open the state file in the review mode.")
-@click.option("--api-token", help="The token used by Recce Cloud API.", type=click.STRING, envvar="RECCE_API_TOKEN")
+@click.option(
+    "--api-token", help="The API token used by Recce Cloud Share Link.", type=click.STRING, envvar="RECCE_API_TOKEN"
```
wcchang1115, line 257: Hi Kent,
Would "used by Share Link" be limited to the usage?
What do you think "The personal token generated by Recce Cloud"?

wcchang1115 (approved):
Hi Kent,
I listed three minor problems, and they don't affect the whole flow.
If they are fine, we can just merge it.

recce/util/api_token.py
kentwelcome: Yes. `installed` could be reserved for installing Recce App. 
We can change use `launched` here.

recce/cli.py
kentwelcome: Good idea!

recce/util/api_token.py
kentwelcome: I think we should tell users when they are using a GitHub token with Recce. However, the `RecceCloud` constructor may not be a good place. I will create a story to follow up.

wcchang1115: No problem, thanks!

PR #749, opened 2025-05-26
[Bug] Format query when there's added columns

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

- If there are added columns, we can't directly run query diff with the primary key in the warehouse, because there's a set operation running under the hood.

- The model has added column
截圖 2025-05-26 下午5 47 42

- the corresponding default SQL statement
截圖 2025-05-26 下午5 48 27


popcornylu (changes requested):
Per discussion, there are these issues

1. If both side have no columns, it show blank query. Please we show `select *` instead?
1. If a model only exist only in one side, show full query
1. If no schema for both side, show `select *`

wcchang1115:
Okay, it's intuitive to show what columns I saw from the schema view.

popcornylu (approved):
LGTM

PR #751, opened 2025-05-28
[Bug] Fix the layout issue in the CLL with model-to-column lineage

popcornylu:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

Before
image

After
image

### DIFF
--- a/js/src/components/lineage/lineage.ts
+++ b/js/src/components/lineage/lineage.ts
@@ -409,7 +409,7 @@ export function toReactflow(
         if (!nodeAndColumnSet.has(source)) {
           continue;
         }
-        if (nodeColumnSet.has(parentColumn.column)) {
+        if (!nodeAndColumnSet.has(target)) {
           continue;
         }
 


js/src/components/lineage/lineage.ts
```diff
@@ -434,10 +397,22 @@ export function toReactflow(
...
+          : new Set<string>();
 
       for (const parentColumn of nodeDependsOn) {
         const source = `${parentColumn.node}_${parentColumn.column}`;
         const target = node.id;
+
+        if (!nodeAndColumnSet.has(source)) {
+          continue;
+        }
+        if (nodeColumnSet.has(parentColumn.column)) {
```
wcchang1115: Hi Pop,
Should we use the model name and column name as the key here?
For the case where the same column name appears in two models.

wcchang1115 (commented):
Hi Pop,
I got a question, could you check it?
Thanks!

js/src/components/lineage/lineage.ts
popcornylu: Good catch. Fixed.

wcchang1115 (approved):
LGTM, thanks!

PR #768, opened 2025-06-12
[chore] Update wording for single environment mode

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Chore

1. Update wording for `recce debug` and launching single env mode
2. Fix logic while launching Recce with no folder for the current env

PLA-458

- `recce debug` without base
image

- `recce debug` all set
image

- `recce server` without current
image

- `recce server` in single mode
image

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -415,16 +415,31 @@ def server(host, port, lifetime, state_file=None, **kwargs):
         exit(1)
     auth_options["api_token"] = api_token
 
-    target_path = Path(kwargs.get("target_path", "target"))
-    target_base_path = Path(kwargs.get("target_base_path", "target-base"))
     # Check Single Environment Onboarding Mode if the review mode is False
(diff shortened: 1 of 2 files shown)


ijac13:
In without base, does overwrite the default environment mean user should overwrite target as target-base? Isn't that still one environment since no adding schema?

ijac13:
Screenshot of recce debug doesn't update, still shows "single environment.".

even-wei:
@ijac13 
> In without base, does overwrite the default environment mean user should overwrite target as target-base? Isn't that still one environment since no adding schema?

It's to guide the user to run dbt or configure it.

> Screenshot of recce debug doesn't update, still shows "single environment.".

Updated. Please check the screenshot (it's edited).

ijac13:
Thanks @even-wei I'm fine with the updated copy.

PR #795, opened 2025-06-30
[Bug] Fix missing change status for non-nodes lineage node

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

We currently have a limitation in identifying non-node objects that are `added/removed/modified` from the `lineage_diff` result.
So the workaround is to fall back to old logic to identify it in the lineage view.


popcornylu (approved):
LGTM

PR #847, opened 2025-08-02
Feature/drc 1552 oss allow to do data diff when launching with a shared link

popcornylu:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

When --share-url is provides
1. always use the state file in the cloud
2. always use the review mode, that is, the artifacts are from the cloud.
3. Support server mode = normal, which is used to enable data diff.

Examples
```
recce server --cloud --share-url <share_url>
recce server --cloud --share-url <share_url> --mode read-only
recce server --cloud --share-url <share_url> --mode preview
```

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -436,11 +436,21 @@ def server(host, port, lifetime, state_file=None, **kwargs):
     if server_mode == RecceServerMode.server:
         flag = {"single_env_onboarding": False, "show_relaunch_hint": False}
         if is_cloud:
-            cloud_options = {
-                "host": kwargs.get("state_file_host"),
-                "github_token": kwargs.get("cloud_token"),
-                "password": kwargs.get("password"),
-            }
+            share_url = kwargs.get("share_url")
+            if share_url:
+                is_review = kwargs["review"] = True
+                share_id = share_url.split("/")[-1]
+                cloud_options = {
+                    "host": kwargs.get("state_file_host"),
+                    "api_token": api_token,
+                    "share_id": share_id,
+                }
+            else:
+                cloud_options = {
+                    "host": kwargs.get("state_file_host"),
+                    "github_token": kwargs.get("cloud_token"),
+                    "password": kwargs.get("password"),
+                }
 
         # Check Single Environment Onboarding Mode if the review mode is False
         project_dir_path = Path(kwargs.get("project_dir") or "./")
@@ -453,6 +463,7 @@ def server(host, port, lifetime, state_file=None, **kwargs):
             kwargs["target_base_path"] = kwargs.get("target_path")
     elif server_mode == RecceServerMode.preview:
         if is_cloud:
+            is_review = kwargs["review"] = True
             share_url = kwargs.get("share_url")
             share_id = share_url.split("/")[-1] if share_url else None
             cloud_options = {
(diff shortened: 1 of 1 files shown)


recce/cli.py
```diff
@@ -453,6 +475,15 @@ def server(host, port, lifetime, state_file=None, **kwargs):
             kwargs["target_base_path"] = kwargs.get("target_path")
     elif server_mode == RecceServerMode.preview:
         if is_cloud:
+            # recce server --cloud --share-url <share-url> --mode preview
+            # Use state file stored for the share url
+            #
+            # Read-only mode disable these features
```
wcchang1115: typo: "Preview"?

recce/cli.py
```diff
@@ -464,16 +495,28 @@ def server(host, port, lifetime, state_file=None, **kwargs):
...
+            # - run query
+            # - use checklist
+            # - share
+            #
+            # Usage:
+            #    Used in cloud managed instance. Launch when user click a share link.
+            is_review = kwargs["review"] = True
```
wcchang1115: Should we put the `is_review = True` outside the `is_cloud` statement?
Because we can launch `read_only` mode without cloud in shared instance.

wcchang1115 (changes requested):
Hi Pop, 
I only found two minor issues for further check.

recce/cli.py
popcornylu: Here, i only want to change the behavior in the "Cloud managed recce instance" and keep the behavior for the non-cloud mode as it's used to be 

For the local dev, users still can determine if they want to use the manifest from local state file or not.

recce/cli.py
popcornylu: Fixed

recce/cli.py
```diff
@@ -433,14 +439,35 @@ def server(host, port, lifetime, state_file=None, **kwargs):
         "api_token": api_token,
     }
 
+    share_url = kwargs.get("share_url")
+    if share_url:
+        share_id = share_url.split("/")[-1]
```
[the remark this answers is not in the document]
popcornylu: False alert

recce/cli.py
```diff
@@ -433,14 +439,35 @@ def server(host, port, lifetime, state_file=None, **kwargs):
...
-            cloud_options = {
-                "host": kwargs.get("state_file_host"),
-                "github_token": kwargs.get("cloud_token"),
-                "password": kwargs.get("password"),
-            }
+            if share_url:
+                # recce server --cloud --share-url <share-url>
+                # Use state file stored for the share url
+                # Forces use of the review mode.
+                is_review = kwargs["review"] = True
```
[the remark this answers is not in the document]
popcornylu: Keep for downstream usage

recce/cli.py
```diff
@@ -464,16 +498,26 @@ def server(host, port, lifetime, state_file=None, **kwargs):
...
+            # Use state file stored for the share url
+            #
+            # Read-only mode disable these features
+            # - run query
+            # - use checklist
+            # - share
+            #
+            # Usage:
+            #    Used in cloud managed instance. Launch when user click a share link.
+            is_review = kwargs["review"] = True
```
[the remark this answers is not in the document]
popcornylu: Keep for downstream usage

recce/cli.py
```diff
@@ -453,8 +480,15 @@ def server(host, port, lifetime, state_file=None, **kwargs):
...
-            share_id = share_url.split("/")[-1] if share_url else None
+            # recce server --cloud --share-url <share-url> --mode preview
+            # Use state file stored for the share url
+            #
+            # Preview mode disable these features
+            # - run query
+            #
+            # Usage:
+            #    Used in cloud managed instance. For cloud onboarding to preview the uploaded artifacts.
+            is_review = kwargs["review"] = True
```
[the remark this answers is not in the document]
popcornylu: Keep for downstream usage

recce/cli.py
popcornylu: Because the legacy code in the recce-instance-launcher, we have to force to enter review mode to make it compatible.

wcchang1115 (approved):
LGTM, thanks!

PR #897, opened 2025-10-29
[Feature] Support the mcp-server to allow ai agent to interact with recce

popcornylu:
1. Install by  `pip install -e '.[mcp]'`
2. Add mcp to project scope `claude mcp add --transport stdio recce-mcp --scope project recce mcp-server`
3. Run `claude` and check mcp available by `/mcp`
4. Prompt `Get the change from recce'

**PR checklist**

- [x] DCO signed

**What type of PR is this?**

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist = dbt1.6, dbt1.7, dbt1.8, dbt1.9, dbtlatest
 
-[testenv:dbt{1.6,1.7,1.8,latest}]
+[testenv:dbt{1.6,1.7,1.8,1.9,latest}]
 basepython = python3.10
 deps =
     pytest


tox.ini
```diff
@@ -1,32 +1,36 @@
 [tox]
-envlist = dbt1.5, dbt1.6, dbt1.7, dbt1.8, dbtlatest
+envlist = dbt1.6, dbt1.7, dbt1.8, dbt1.9, dbtlatest
 
 [testenv:dbt{1.5,1.6,1.7,1.8,latest}]
```
wcchang1115: should remove `testenv:dbt{1.5}` and add `testenv:dbt{1.9}` to match `envlist`?

tox.ini
```diff
@@ -1,32 +1,36 @@
...
 basepython = python3.10
 deps =
     pytest
+    pytest-asyncio
     pytest-cov
     pandas
     duckdb
     httpx
-    dbt1.5: dbt-duckdb==1.5.*
+    {dbt1.8,dbtlatest}: mcp>=1.0.0
```
wcchang1115: is it intentional to pick only `dbt1.8` and `dbtlatest` to save time?
or it should be between `dbt1.8` and `dbtlatest`

wcchang1115 (changes requested):

tox.ini
popcornylu: It is because i want to test mcp and non-mcp case...  so i only test the mcp for version with EVEN number.

tox.ini
popcornylu: Fixed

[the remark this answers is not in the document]
popcornylu: It is because i want to test mcp and non-mcp case... so i only test the mcp for version with EVEN number.

wcchang1115 (approved):
LGTM, thanks!

PR #923, opened 2025-11-10
Remove dbt_version from RecceStateMetadata

even-wei:
## Summary
Remove the redundant `dbt_version` field from `RecceStateMetadata` in state files. The dbt version information is already available in `manifest.metadata.dbt_version` from the artifact files.

## Problem
When running `recce summary`, users see false warning messages:
```
[WARN] dbt version mismatch. Local dbt adapter: 1.10.8 vs. Generated Manifest dbt adapter: 0
[WARN] Version mismatch can lead to issues when generating queries
```

This happens because:
1. `RecceStateMetadata.dbt_version` defaults to "0" and is never populated
2. After introducing Sessions, artifacts are stored separately in S3, not in state files
3. The dbt version is already available from manifest metadata

## Changes
- Remove `dbt_version` field from `RecceStateMetadata` class
- Remove dbt version mismatch warning logic from `RecceState.from_json()`
- Remove unused `console` import from `recce/state/state.py`

## Benefits
✅ Eliminates false warning messages  
✅ Cleaner state file metadata with only used fields  
✅ Simpler code that's easier to maintain  
✅ Backward compatible - old state files with `dbt_version` can still be loaded  

## Testing
- All 68 state-related tests pass
- Verified backward compatibility with old state file format
- Verified new state files don't include `dbt_version`
- Code passes flake8 linting and Black formatting

### DIFF
--- a/recce/state/state.py
+++ b/recce/state/state.py
@@ -8,7 +8,6 @@ from typing import Dict, List, Optional
 from pydantic import BaseModel, Field
 
 from recce import get_version
-from recce.config import console
 from recce.exceptions import RecceException
 from recce.git import current_branch
 from recce.models.types import Check, Run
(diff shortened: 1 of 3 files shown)


gcko (approved):
LGTM

Issue #931, opened 2025-11-12
[Bug] Visual issues in the checklist


alexeyegorov:
### Current Behavior

During a review of changes, we run several queries on specific changed models to review data changes. Whenever anything is detected, we add such checks to the checklist which is synced to the cloud and can be later reviewed again. The preview is prepared in the CI step of the pipeline and whenever we start recce with `recce server --cloud --review` and switch to those checks we immediately see the results. However, we experience the following issues:
* in the version 1.23.0: when scrolling through the results, the header is transparent (minor issue)
  <img width="2764" height="430" alt="Image" src="https://github.com/user-attachments/assets/48f968d9-a36b-4b6d-b604-74ec7e6b8b72" />
 
* in the version 1.24.0:
  * the header is still transparent (as above)
  * the results for some checks are not shown unless you rerun the check
  <img width="2940" height="1110" alt="Image" src="https://github.com/user-attachments/assets/aa5f521e-1bef-4c2f-9541-3358e28dd68c" />
  
  * for the new entries (green) the values of the id column are not displayed
  <img width="2308" height="762" alt="Image" src="https://github.com/user-attachments/assets/f7401bb1-2465-4b9d-a4e4-3e9cc56fc7f0" />

### Expected Behavior

* the header should not be transparent (or at least not that much)
* the results should be displayed directly (as before 1.24.0)
* the values of the newly added entries in the ID column should be shown (as before 1.24.0)

### Steps To Reproduce

We follow the same best practices by Recce as before and execute `recce server --cloud --server` as before.
The issue with checks not showing results and empty new entries is happening when using "custom queries" and selecting a primary key for a diff.

### Relevant log output

```shell

```

### Environment

- recce: 1.23.1, 1.24.0
- OS: Mac OSX (latest)
- Python: 3.11.13
- Data Warehouse: Databricks
- dbt: 1.10.11

kentwelcome:
Hi @alexeyegorov, thanks for reporting the issue. 
For the transparent header issue, we plan to fix it in the next release.
Let's focus on the second issue here.

wcchang1115:
Hello @alexeyegorov,

The following visual problems have been fixed in `v1.24.2`:
- the header should not be transparent
- the values of the newly added entries in the ID column should be shown

We are still working on the remaining one:
- the results should be displayed directly

Thanks for reporting these issues!

alexeyegorov:
@wcchang1115 @kentwelcome you are fast, guys! 😬💯

[...]

PR #934, opened 2025-11-13
bugfix: DRC-2118 Ensure header has background color so it is not transparent

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
BugFix

This PR ensures the column headers have an applied background color. Now they do not overlap with row values when scrolling

https://github.com/user-attachments/assets/146cf2f2-2f54-4b5b-a175-875b405af932

DRC-2118
Fixes issue #931

1. Create a query with many rows
2. scroll the list, note that the column header has a background color

Column header no longer overlaps with row values visually

### DIFF
--- a/js/src/components/query/styles.css
+++ b/js/src/components/query/styles.css
@@ -4,7 +4,8 @@
 }
 
 .diff-header-added {
-  background-color: #3aa53a;
+  /* green-700 theme.ts */
+  background-color: #15803D;
   color: white;
 }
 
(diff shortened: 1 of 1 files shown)


iamcxa (approved):
LGTM, Maybe we can have a plan to take all constant colors and make them a variable like ` border-color: var(--rdg-border-color);`?

gcko:
@iamcxa We have theme colors already defined, we can add them to the `global.css` file and import them into the `styles.css` file

PR #935, opened 2025-11-13
feat: DRC-2058 Only add unique or not_null columns to the list

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Feature

Constrain the suggested Primary Keys based on dbt tests - whether the user has either `unique` or `not_null` defined

DRC-2058

- Check the logic in QueryForm.tsx is correct

The Primary Key suggested list will be more accurate to potential primary keys, rather than a list of all columns throughout the entire schema


gcko:
BEFORE showed this amount of columns:
Screenshot 2025-11-13 at 15 15 15

Now it looks like this:
Screenshot 2025-11-14 at 10 04 44

PR #952, opened 2025-11-18
fix: address CodeQL security warning in cloud integration tests

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Security fix

Split the cloud integration test workflow into two jobs to address CodeQL security warning about "Checkout of untrusted code in a privileged context".

The `authorize` job checks PR author permissions without checking out any untrusted code. The `smoke-test-cloud` job only runs if authorization passes. This ensures the authorization check happens before any PR code is checked out.

Fixes CodeQL security warning in `.github/workflows/integration-tests-cloud.yaml`

The workflow now has two jobs instead of one, but the behavior for authorized users remains the same.

NONE

---


.github/workflows/integration-tests-cloud.yaml
```diff
@@ -29,29 +62,23 @@ jobs:
...
-
-      - uses: actions/checkout@v3
+      # Safe checkout: only proceeds after authorization check passes
+      # codeql[js/actions/pull-request-target] - Authorization verified in previous job
+      - uses: actions/checkout@v4
         with:
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.pull_request.head.sha }}
-          token: ${{ secrets.RECCE_CLOUD_TOKEN }}
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
```
wcchang1115: Hi even,
Because we changed the event to `pull_request`, I think the checkout can be simplified as the default usage.
```
    - uses: actions/checkout@v4
      with:
        persist-credentials: false
```

.github/workflows/integration-tests-cloud.yaml
```diff
@@ -7,15 +7,48 @@ on:
...
+            echo "authorized=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          PERMISSION=$(gh api repos/${{ github.repository }}/collaborators/${{ github.triggering_actor }}/permission --jq '.permission' 2>/dev/null || echo "none")
+          if [[ "$PERMISSION" == "write" || "$PERMISSION" == "admin" ]]; then
+            echo "authorized=true" >> $GITHUB_OUTPUT
+          else
+            echo "authorized=false" >> $GITHUB_OUTPUT
+            echo "::warning::PR author ${{ github.triggering_actor }} does not have write permission. A maintainer must re-run this workflow after reviewing the code."
```
wcchang1115: Hi Even,
Like we discussed, re-run can't work after the change to `pull_request`.
And in your new version, the PR from forked repo will stops at the authorize job without failing.

So I'm thinking we can refine the warning statement.
e.g., The cloud smoke test is about to be migrated. Skip for the repository for the forked repo.

What do you think?

PR #975, opened 2025-12-02
Feature/refactor routing

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Refactor

Refactor Routing - Remove Wouter in place of using NextJS standard routing

This removes hashbang `#!` style routing in favor of standard url routing. This also allows users to refresh a page 

1. hashbang `#!` is removed - if a hashbang is provided it will rewrite to the new style

URL's will now be `/lineage` instead of `#!/lineage`

### DIFF
--- a/js/app/MainLayout.tsx
+++ b/js/app/MainLayout.tsx
@@ -85,7 +85,6 @@ function Main({ children, lineage, isLineageRoute }: MainProps) {
   const pathname = usePathname();
 
   const _isRunResultOpen = isRunResultOpen && !pathname.startsWith("/checks");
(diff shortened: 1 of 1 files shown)


kentwelcome:
When the RunResultPane is shown on the button, switching the tab between `Lineage` and `Query` should not affect the `RunResultPane`. Before, the `RunResultPane` would stay on the screen when switching between the `Lineage` and `Query` tabs.
CleanShot 2025-12-04 at 11 10 57

kentwelcome:
CleanShot 2025-12-04 at 11 12 47
The loading indicator when opening the `Linage` tab can use the `Linear Progress` instead of the circular progress.

PR #981, opened 2025-12-05
fix: Don't show 100% if the value is not 1 in the value diff

popcornylu:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug fix

If the value is 99.999% it show to 100% by rounding.

Now, it show ~99.99%
image

DRC-2241

### DIFF
--- a/js/src/components/valuediff/ValueDiffResultView.tsx
+++ b/js/src/components/valuediff/ValueDiffResultView.tsx
@@ -182,7 +182,7 @@ function _ValueDiffResultView(
           if (value > 0.9999 && value < 1) {
             displayValue = "~99.99 %";
           } else if (value < 0.0001 && value > 0) {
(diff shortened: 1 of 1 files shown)


js/src/components/valuediff/ValueDiffResultView.tsx
```diff
@@ -176,11 +176,19 @@ function _ValueDiffResultView(
...
-            {value != null ? `${(value * 100).toFixed(2)} %` : "N/A"}
-          </Box>
-        );
+        let displayValue = "N/A";
+
+        if (value != null) {
+          if (value > 0.9999 && value < 1) {
+            displayValue = "~99.99 %";
+          } else if (value < 0.0001 && value > 0) {
+            displayValue = "~00.01 %";
```
gcko: `~0.01`

PR #992, opened 2025-12-09
feat: Change the default filter for package

popcornylu:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug

In the recce-mcp or ai summary in the cloud, we should exclude the diff for the non current projects' nodes.

For example, if the users user elementary, it would have the diff if product&pr have have different strategy to include the packages.

By default, we should only care about the nodes under current project.


even-wei (approved):
LGTM

PR #1011, opened 2025-12-12
fix: Strip SQL identifier quotes from primary keys before returning to frontend

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Bug fix

Quoted primary keys (e.g., `"customerID"`) are needed for SQL execution to preserve case sensitivity, but should be stripped before returning to the frontend. Previously, the frontend received the quoted key in `params.primary_keys` (e.g., `"customerID"`) but the column key in the result was unquoted (`customerID`), causing a `DataGridValidationError: Primary key column '"customerID"' not found in columns`.

**Changes:**
- Update `normalize_keys_to_columns()` to strip SQL identifier quotes before matching
- Support all quote types: double quotes (`"col"`), backticks (`` `col` ``), and square brackets (`[col]`)
- Remove redundant quote-stripping code in `RunDAO.find_run_by_id()` (was incomplete - only handled double quotes and mutated objects in place)
- Add comprehensive tests for quote-stripping functionality (12 new tests)

Fixes DRC-2292

The `strip_identifier_quotes()` function was already added in a previous commit. This PR integrates it into `normalize_keys_to_columns()` so quotes are stripped consistently for all primary key handling.

Yes - fixes a bug where value diff checks with quoted primary keys would fail with `DataGridValidationError`.

Test screenshots:
image
image
image
image


gcko (approved):
LGTM

PR #1035, opened 2025-12-25
feat(experiments): add metadata storage and dbt-core decoupling experiments

even-wei:
Add experiment framework to validate two goals:
1. Store essential dbt metadata in warehouses alongside dbt runs
2. Decouple Recce from dbt-core using sqlglot + SQLAlchemy

Experiments included:
- metadata_extractor.py: Version-agnostic extraction from manifest/catalog (v9-v12+)
- sql_adapter.py: Unified SQL adapter using sqlglot + SQLAlchemy
- warehouse_storage.py: Store/retrieve metadata from warehouse tables
- run_experiments.py: End-to-end test runner
- EXPERIMENT_PLAN.md: Detailed plan with phases and success criteria

Key findings:
- Essential fields (nodes, sources, parent_map, checksum) are identical across v9-v12
- Can extract metadata without any dbt-core dependency
- Catalog schema is stable at v1 across all dbt versions
- Only ~13% of manifest is actually used by Recce (macros are 86%)

Signed-off-by: Claude <claude@anthropic.com>
Signed-off-by: Claude <noreply@anthropic.com>

### DIFF
--- a/experiments/RESEARCH_CONCLUSION.md
+++ b/experiments/RESEARCH_CONCLUSION.md
@@ -1,0 +1,362 @@
+# Research Conclusion: Recce dbt Artifact Consumption & dbt Package Feasibility
+
+## Executive Summary
+
+**Feasibility: ✅ CONFIRMED**
+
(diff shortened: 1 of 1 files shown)


even-wei:
It's a no go due to macro issue.

PR #1041, opened 2025-12-28
feature/drc-2060-ui-update-change-to-0-when-no-difference-in-row-count

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**
Bug fix

Change calculateDelta to return "0" instead of "=" when values match

drc-2060

- You should now see 0 instead of = in the delta column for row count diff

Screenshot 2025-12-28 at 15 31 17

You will now see 0 instead of = in the delta column for row count diff


kentwelcome (approved):
LGTM

PR #1073, opened 2026-01-23
refactor: DRY refactoring - UI components and utilities

danyelf:
Extract shared components and utilities:

- viewOptionsHandlers - shared view options utilities for UI components
- FormLoadingErrorState - shared loading/error state component (reverted - inline checks preferred)
- verify_state_loader_or_exit() - CLI helper for state loader verification
- RecceCloud inherits CloudBase - removes duplicate cloud client code
- DbtVersion semantic version properties
- CheckValidator factory for DRY validation
- BasePerformanceTracker - shared timing logic base class
- createRunFromResult adapter for UI


danyelf:
this seems to be a bit of a mess.

PR #1086, opened 2026-01-28
feat: Handle cloud_user_context WebSocket message for user attribution (DRC-2643)

gcko:
**PR checklist**
- [x] DCO signed

**What type of PR is this?**
Feature

When one user launches a shared instance in Recce Cloud and another user connects to that instance, actions (creating checklist items, comments) were being attributed to the launching user instead of the actual user performing the action.

This PR implements the OSS-side handling for cloud user context, enabling proper user attribution in shared instances.

**Changes:**
- Add `CloudUserContext` and `CloudUserContextMessage` models for parsing messages
- Add `WebSocketConnectionManager` with per-connection user context storage
- Update WebSocket endpoint to handle `cloud_user_context` messages and send acknowledgment
- Add HTTP middleware to extract user context from `X-Recce-User-*` headers
- Pass cloud user context for check creation (`created_by` field)
- Add `X-Recce-Acting-User-Id` header support for Cloud API calls (comments)
- Add 36 unit and integration tests

**Architecture:**
Two mechanisms for user context propagation:
1. **WebSocket messages** - Cloud sends `cloud_user_context` when a user connects via the proxy
2. **HTTP headers** (NEW) - Middleware extracts `X-Recce-User-Id`, `X-Recce-User-Login`, `X-Recce-User-Email` from proxied HTTP requests

**Current Limitations:**
The HTTP header mechanism is required for full functionality because:
- WebSocket context is stored per-connection in the connection manager
- HTTP requests run in separate async contexts and cannot access WebSocket-stored context
- The middleware is in place but currently a no-op until Cloud proxy sends headers on HTTP requests

Fixes DRC-2643

- **Cloud API requirement**: The Cloud API needs to respect the `X-Recce-Acting-User-Id` header for comment attribution
- **Cloud proxy requirement**: The proxy needs to send `X-Recce-User-*` headers on HTTP requests (not just WebSocket)
- Cloud-side WebSocket proxy changes are in a separate PR in recce-cloud-infra
- Backward compatible: old instances ignore unknown message types/headers, old cloud doesn't send them
- Follow-up task created: DRC-2645 for Cloud proxy HTTP header changes

Yes - When multiple users access the same shared instance, their actions (creating checks, adding comments) will now be correctly attributed to them instead of the instance launcher. Full functionality requires coordinated Cloud proxy updates.

### DIFF
--- a/recce/util/cloud/base.py
+++ b/recce/util/cloud/base.py
@@ -78,14 +78,6 @@ class CloudBase:
         # Include acting user header for shared instance user attribution
         if acting_user_id:
             headers["X-Recce-Acting-User-Id"] = acting_user_id
-            # DEBUG: Log for DRC-2643 investigation
-            import logging
-            logger = logging.getLogger("uvicorn")
-            logger.info(f"[DRC-2643] CloudBase._request: Adding X-Recce-Acting-User-Id={acting_user_id} to {method} {url}")
-        else:
-            import logging
-            logger = logging.getLogger("uvicorn")
-            logger.info(f"[DRC-2643] CloudBase._request: NO X-Recce-Acting-User-Id (acting_user_id is None) for {method} {url}")
 
         url = self._replace_localhost_with_docker_internal(url)
         return requests.request(method, url, headers=headers, **kwargs)
--- a/recce/util/cloud/checks.py
+++ b/recce/util/cloud/checks.py
@@ -87,12 +87,6 @@ class ChecksCloud(CloudBase):
             >>> check = client.create_check("org123", "proj456", "sess789", check_data)
             >>> print(f"Created check with ID: {check['id']}")
         """
-        import logging
-        logger = logging.getLogger("uvicorn")
-        logger.info(
-            f"[DRC-2643] ChecksCloud.create_check: acting_user_id={acting_user_id}, session={session_id}"
-        )
-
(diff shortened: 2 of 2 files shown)


gcko:
## Response to Review Comments

Thanks @kiloconnect and @Copilot for the thorough review! You correctly identified a flaw in the original implementation.

### The Issue
The original code stored user context in a ContextVar during WebSocket message handling, but HTTP requests (check creation, comments) run in separate async contexts and cannot access this value. `get_current_cloud_user()` would always return `None`.

### The Fix (commit 1ad6371a)
We've added **HTTP header-based middleware** to address this:
- Added `extract_cloud_user_context` middleware in `server.py`
- Middleware extracts user context from `X-Recce-User-Id`, `X-Recce-User-Login`, `X-Recce-User-Email` headers
- Middleware calls `set_current_cloud_user()` before the request handler runs
- Added 9 new tests for this functionality

### Current Status
- ✅ OSS middleware is in place and tested
- ⏳ Requires Cloud proxy to send these headers on proxied HTTP requests (not just WebSocket messages)
- 📋 Created follow-up task [DRC-2645](https://linear.app/recce/issue/DRC-2645) for the Cloud proxy changes

### Regarding datetime.utcnow() deprecation
Good point about `datetime.utcnow()` being deprecated in Python 3.12+. We'll address this in a separate cleanup.

even-wei (approved):
LGTM

PR #1282, opened 2026-04-08
feat(breaking): add column rename detection to schema diff analysis

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Feature — adds column rename detection to the breaking change analysis pipeline.

When a dbt model column is renamed (e.g., `number_of_orders` → `order_count`), the schema diff currently shows it as a separate "removed" + "added" pair. This is noisy and obscures the developer's intent.

This PR adds rename detection to the existing breaking change analysis in `_diff_select_scope()`. After the column comparison loop identifies removed and added columns, it compares their underlying SQL expressions (stripped of aliases). Columns with identical expressions are paired as renames instead of remove+add.

**Changes:**
- `recce/models/types.py`: Add `"renamed"` to `ChangeStatus` literal type; add `rename_map: Optional[dict[str, str]]` to `NodeChange` model
- `recce/util/breaking.py`: Add `_unalias()` helper and rename detection logic (~30 lines) in `_diff_select_scope()`
- `tests/util/test_breaking.py`: Add 12 new rename detection tests + update 2 existing tests

Resolves DRC-3185

- This is Phase 1 (backend only). Frontend rendering of "renamed" status will come in Phase 3.
- The `rename_map` field on `NodeChange` has no consumers yet — it carries metadata for future frontend use.
- Known gaps deferred to Phase 3: `rename_map` case normalization in adapter, union scope `rename_map` merging, star propagation category escalation for "renamed". All documented in the review report.
- The rename detection uses greedy first-match pairing, which is correct for the common case (1:1 renames). Ambiguous cases (multiple columns with identical expressions) are handled but the pairing is deterministic rather than optimal.

Backend-only change. The `"renamed"` status will be surfaced in the UI in a future Phase 3 PR. No user-visible change yet.


even-wei:
Closing — we are not pursuing semantic SQL equivalence / rename detection further at this time. (DRC-3185)

PR #1285, opened 2026-04-09
fix(cll): include adapter_type in cache key to prevent cross-database poisoning

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Bug fix + refactor

Fixes cross-database cache poisoning in the CLL SQLite cache and redesigns the cache key for better invalidation.

**Bug:** The CLL cache key did not include the database adapter type. When a user switched adapters (e.g. Snowflake → DuckDB), the cache silently returned lineage computed under the wrong dialect — Snowflake uppercases all identifiers while DuckDB preserves case, causing column mapping failures.

**Redesign:** While investigating, we found the cache key used raw SQL strings and parent node IDs as proxies. Replaced with a more principled design using dbt-computed checksums:

| Before | After |
|---|---|
| `sha256(node_id + raw_code + parent_ids + column_names)` | `sha256(adapter_type + checksum + parent_checksums + column_names)` |

Key improvements:
- **`adapter_type`** — prevents cross-database poisoning (the original bug)
- **`checksum`** — uses dbt's pre-computed `sha256(raw_code)` instead of re-hashing large SQL strings
- **`parent_checksums`** — cascading invalidation: if any parent's SQL changes, children recompute (previously only tracked parent IDs, not content)
- **`node_id` removed from content key** — already included in the DB lookup key (`CllCache.make_node_key`), was redundant
- **Jinja fallback warning** — `logger.warning` when `compiled_code` is missing and the fragile Jinja rendering path is used

**Verified with jaffle-shop-expand (1060 models):**
1. DuckDB cold cache → 1060 computed (194s)
2. DuckDB warm cache → All 1060 cached (0.2s)
3. Switched manifest to `adapter_type: snowflake` → **0 cache hits**, 973 recomputed (fix confirmed)

Resolves DRC-3199

Backward compatible — old cache entries (hashed with the old key format) become natural cache misses and recompute on first access. No migration or `recce cache clear` needed.

Sources, exposures, and metrics have no `checksum` in the manifest (no SQL), so their node ID is used as a stable placeholder.

Users who switch between database adapters (e.g. DuckDB ↔ Snowflake) with CLL cache enabled will no longer get silently incorrect lineage from stale cache entries. Additionally, changes to parent model SQL now correctly invalidate downstream cache entries.


wcchang1115:
## Code Review — PR #1285

### Summary

Solid fix for a real cache poisoning bug. The cache key redesign (adapter_type + checksums instead of raw_code + node IDs) is well-motivated and the test coverage is thorough. Two findings — one is a latent collision in the hash function (pre-existing, low practical risk), the other is code duplication that could drift.

### Findings

#### [Warning] Missing list boundary separator allows theoretical hash collisions
**File:** `recce/adapter/dbt_adapter/__init__.py:1021-1026`
**Issue:** The parent_checksums and column_names lists are hashed with the same `value \x00` pattern, with no separator between the two lists. This means `parent_checksums=["a","b"], column_names=[]` produces the same hash as `parent_checksums=["a"], column_names=["b"]`:

```python
for p in sorted(parent_checksums):
    h.update(p.encode("utf-8"))
    h.update(b"\x00")
# ← no boundary marker here
for c in sorted(column_names):
    h.update(c.encode("utf-8"))
    h.update(b"\x00")
```

Verified with a quick test — both cases produce `bb7f88...`.

**Practical risk: very low.** Parent checksums are 64-char SHA256 hex strings (or node IDs like `source.pkg.table`) while column names are short identifiers (`id`, `name`). These domains never overlap in practice.

**Suggestion:** Insert a sentinel separator between the two lists: `h.update(b"\x01")` before the column loop. Note: this is a pre-existing pattern from the original code, so fixing it would invalidate all existing caches. Fine to defer.

#### [Warning] `_get_checksum` logic duplicated between adapter and CLI
**File:** `recce/adapter/dbt_adapter/__init__.py:1065-1071` and `recce/cli.py:408-413`
**Issue:** The `_get_checksum` nested function is copy-pasted between `build_full_cll_map()` and the CLI `init` command. Both do the same thing: look up `manifest.nodes[nid].checksum.checksum`, fall back to `nid`. If the checksum extraction logic changes (e.g., handling a new dbt manifest format), it needs to be updated in two places.

**Suggestion:** Extract to a static method on `DbtAdapter` (next to `_make_node_content_key`) that takes `manifest` and `nid` as parameters. Both call sites become one-liners.

### Verdict

Approved — no critical issues. The key generation redesign is correct and well-tested. The two findings above are both low-priority improvements that can be addressed in a follow-up. The cache key now properly includes adapter_type, uses dbt checksums for cascading invalidation, and the test suite covers all the important differentiation cases.

wcchang1115 (approved):
LGTM, thanks!

even-wei:
@wcchang1115 Both findings addressed in d3a8bb47:

**1. List boundary sentinel** — Added `h.update(b"\x01")` between the parent_checksums and column_names loops. Since this PR already changes the cache key format, no extra cache invalidation cost. Added a test that verifies the exact collision case (`parent_checksums=["a","b"], column_names=[]` vs `parent_checksums=["a"], column_names=["b"]`).

**2. Extracted `_get_node_checksum`** — Promoted to `DbtAdapter._get_node_checksum(manifest, nid)` static method. Both `build_full_cll_map` and the CLI `init` command now call it instead of duplicated nested functions.

PR #1315, opened 2026-04-15
fix(lineage): make impacted-column snapshot trigger sidebar re-render

danyelf:
## Summary

- Promotes `impactedNodeIdsRef` / `impactedColumnIdsRef` from `useRef` to `useState` in `LineageViewOss` so downstream memos — notably `SchemaView`'s `impactedColumns` — re-render when whole-graph impact CLL finishes populating the snapshot.
- Previously a ref mutation meant `SchemaView` latched onto the initial empty `Set()`; if the user opened the schema sidebar while the impact CLL was in flight, amber highlights never appeared once the data resolved.
- Behavior preserved: the snapshot still "freezes" across column navigation (a column-scoped fetch only writes when the snapshot is still empty or when impact analysis completes without a `column` scope).

## Context

Race diagnosed via Playwright + injected logs:
- `impactedColumnIdsRef` starts empty; writes at `LineageViewOss.tsx:625` / `:915` are guarded by `!cllInput?.column`.
- `SchemaView.tsx:232` reads `lineageViewContext.impactedColumnIds` (which pointed at `impactedColumnIdsRef.current`) inside a `useMemo`. Because refs don't trigger re-renders, the memo never re-ran when impact CLL populated the set after an open sidebar.

With a 5 s delay simulated on the impact-analysis fetch (before fix):
- `SV:frozen-size=0` @ +1354ms (sidebar blank)
- `ref-update` @ +6010ms (ref filled, but no re-render)
- Sidebar stays blank indefinitely.

After fix (verified against Snowflake-backed jaffle_shop, `change_analysis=true`, 46 impacted columns):
- State setter fires when impact CLL resolves → context Provider re-renders → `SchemaView`'s `useMemo` re-runs → amber paints on `orders.STATUS` and all downstream-impacted columns.

## Test plan

- [x] `pnpm type:check` clean
- [x] `pnpm lint:fix` clean
- [x] `pnpm test` — 3664 passed, 5 skipped
- [x] Manual verification: Snowflake `jaffle_shop`, `--new-cll-experience --impact-at-startup`, sidebar shows amber on impacted columns (`orders.STATUS`, `customers.FULL_NAME` downstream, etc.) both on patient load and during a simulated 5 s CLL delay.

## User-facing changes

- Schema sidebar now reliably shows amber highlights on columns impacted by upstream breaking changes, even when opened mid-fetch.

### DIFF
--- a/js/packages/ui/src/components/lineage/LineageViewOss.tsx
+++ b/js/packages/ui/src/components/lineage/LineageViewOss.tsx
@@ -475,8 +475,14 @@ export function PrivateLineageView(
 
   // Guard: auto-trigger impact analysis only once per mount
   const impactAtStartupFired = useRef(false);
-  const impactedNodeIdsRef = useRef<Set<string>>(new Set());
-  const impactedColumnIdsRef = useRef<Set<string>>(new Set());
+  // State (not refs) so downstream memos re-render when impact CLL populates the
+  // snapshot; a ref mutation would leave SchemaView latched on the initial empty Set.
+  const [impactedNodeIds, setImpactedNodeIds] = useState<Set<string>>(
+    new Set(),
+  );
+  const [impactedColumnIds, setImpactedColumnIds] = useState<Set<string>>(
+    new Set(),
+  );
 
   // biome-ignore lint/correctness/useExhaustiveDependencies: Intentionally only run when lineageGraph changes (initial load/refetch).
   useLayoutEffect(() => {
@@ -621,8 +627,8 @@ export function PrivateLineageView(
       // Snapshot impacted sets during impact analysis so they stay stable
       // when the user clicks a column (which returns column-scoped CLL data).
       if (impacted && cllInput?.change_analysis && !cllInput?.column) {
-        impactedNodeIdsRef.current = impacted.nodeIds;
-        impactedColumnIdsRef.current = impacted.columnIds;
+        setImpactedNodeIds(impacted.nodeIds);
+        setImpactedColumnIds(impacted.columnIds);
       }
 
       // Track lineage view render
@@ -911,8 +917,8 @@ export function PrivateLineageView(
 
     // Snapshot impacted node and column IDs during impact analysis (see layout effect).
     if (impacted && !cllInput2?.column) {
-      impactedNodeIdsRef.current = impacted.nodeIds;
-      impactedColumnIdsRef.current = impacted.columnIds;
+      setImpactedNodeIds(impacted.nodeIds);
+      setImpactedColumnIds(impacted.columnIds);
     }
 
     // Track lineage view render
@@ -1141,8 +1147,8 @@ export function PrivateLineageView(
     selectParentNodes,
     selectChildNodes,
     deselect,
-    impactedNodeIds: impactedNodeIdsRef.current,
-    impactedColumnIds: impactedColumnIdsRef.current,
+    impactedNodeIds,
+    impactedColumnIds,
     isNodeHighlighted: (nodeId: string) => highlighted.has(nodeId),
     isNodeSelected: (nodeId: string) => selectedNodeIds.has(nodeId),
     isEdgeHighlighted: (source, target) => {
--- a/js/packages/ui/src/components/lineage/LineageViewOss.tsx
+++ b/js/packages/ui/src/components/lineage/LineageViewOss.tsx
@@ -105,6 +105,7 @@ import {
 import {
   useLineageCopyToClipboard,
   useNavToCheck,
+  usePublishedImpactSets,
   useResizeObserver,
   useTrackLineageRender,
 } from "./hooks";
@@ -475,14 +476,11 @@ export function PrivateLineageView(
 
   // Guard: auto-trigger impact analysis only once per mount
   const impactAtStartupFired = useRef(false);
-  // State (not refs) so downstream memos re-render when impact CLL populates the
-  // snapshot; a ref mutation would leave SchemaView latched on the initial empty Set.
-  const [impactedNodeIds, setImpactedNodeIds] = useState<Set<string>>(
-    new Set(),
-  );
-  const [impactedColumnIds, setImpactedColumnIds] = useState<Set<string>>(
-    new Set(),
-  );
+  const {
+    impactedNodeIds,
+    impactedColumnIds,
+    publish: publishImpactSets,
+  } = usePublishedImpactSets();
 
   // biome-ignore lint/correctness/useExhaustiveDependencies: Intentionally only run when lineageGraph changes (initial load/refetch).
   useLayoutEffect(() => {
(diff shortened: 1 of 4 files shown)


danyelf:
## Code Review — PR 1315

### Summary
Tight, well-scoped fix that does correctly address the described race. Verdict: **yes, this really solves it** — not just paper over it.

### Does it actually fix the race?

Yes. The data flow is:
- `SchemaView.tsx:232` memo: `useMemo(..., [newCllExperience, lineageViewContext?.impactedColumnIds])`
- Before this PR, `impactedColumnIds` came from `impactedColumnIdsRef.current`. Mutating a ref does not trigger a re-render of `LineageViewOss`, so the context value that `SchemaView` received never changed. Even on an unrelated re-render, the memo's dep was still the identity-stable `Set` that `LineageViewOss` closed over at that moment — so once `SchemaView` latched the empty Set on first read, it never saw a populated one.
- After this PR, `setImpactedColumnIds(impacted.columnIds)` triggers a re-render with a new `Set` identity → context Provider delivers new value → `SchemaView`'s memo re-runs. This is the React-idiomatic way to express "snapshot state", and the freeze semantics still hold because both writers are guarded by `!cllInput?.column` (`LineageViewOss.tsx:629, 919`).

### Findings

#### [Warning] `contextValue` is not memoized
**File:** `js/packages/ui/src/components/lineage/LineageViewOss.tsx:1137`
**Issue:** The `contextValue` object is rebuilt on every render of `LineageViewOss`. Every consumer re-renders whenever the parent does, regardless of what actually changed. Pre-existing, but worth noting because this PR now relies on the Set identity propagating through context — it works, but would be more robust with a memoized context value (or a split context).
**Suggestion:** Follow-up, not blocking.

#### [Warning] Guard asymmetry between the two write sites
**File:** `js/packages/ui/src/components/lineage/LineageViewOss.tsx:629` vs `:919`
**Issue:** Layout-effect writer requires `cllInput?.change_analysis && !cllInput?.column`. `handleViewOptionsChanged` writer only requires `!cllInput2?.column`. This is pre-existing, but promoting the snapshot to observable state means any asymmetry is now visible to consumers. Worth confirming this is intentional (I suspect the layout-effect guard is the correct one — the second site should probably also check `change_analysis`).
**Suggestion:** Align the guards, or add a comment explaining why they differ.

#### [Nit] No automated regression test
**File:** n/a
**Issue:** The race is verified only via manual Playwright with injected 5s delay. Easy to reintroduce (someone refactors back to `useRef` thinking "it never changes across re-renders"). A test rendering `LineageViewOss` + `SchemaView`, resolving impact CLL asynchronously, and asserting the memo re-runs with the populated set would pin the behavior.
**Suggestion:** Optional but high-value given the subtle nature of the bug.

### Non-issues checked
- Initial `new Set()` state is fine; no stale-data flash.
- Column-scoped fetches (`cllInput?.column` truthy) correctly skip both setters — freeze preserved.
- Both `setState` calls batch into a single render.
- `setState` inside `useLayoutEffect`'s async body is safe here.
- Type errors in `useTrackLineageRender.ts` and `createResultView.test.tsx` from `pnpm type:check` are **pre-existing** (confirmed: `git diff origin/main` shows no touches). Not introduced by this PR.

### Verdict
**Approved.** Fix is correct, minimal, and targeted. The warnings are about adjacent code quality, not the correctness of this change. The diagnosis in the PR body matches reality; `useState` is the right fix. ✅

even-wei:
# Code Review: PR #1315

**Reviewer:** Claude Code (structured multi-pass)
**Files reviewed:** 1 (`LineageViewOss.tsx`)
**Categories:** Frontend (business logic + UI state)
**Passes run:** A, B, C, D, E, F, G, H

---

## Validation Results

The `useRef` → `useState` migration is the correct fix. Key verification points:

- **State setters are stable** — `setImpactedNodeIds` / `setImpactedColumnIds` don't need to be in the `useLayoutEffect` dependency array (React guarantees setter identity stability). The effect only *writes* to these state values, never *reads* them, so no stale closure risk.
- **Guard preserved** — `!cllInput?.column` guard at both call sites (`:631` and `:920`) ensures the snapshot only updates during impact analysis, not column-scoped CLL. Behavior unchanged.
- **`useState(new Set())` initializer** runs once on mount, then `setImpactedNodeIds(impacted.nodeIds)` replaces with a new Set identity on each impact completion — correct identity-based change detection for downstream memos.

### Pass B: Security — PASS

Pure UI state management change. No user input, API calls, or data exposure affected.

- `LineageViewContextType` (`:270-272`) declares `impactedNodeIds: Set<string>` and `impactedColumnIds: Set<string>` — types match the new `useState<Set<string>>` declarations.
- `SchemaView.tsx:234` reads `lineageViewContext?.impactedColumnIds` in a `useMemo` with it as a dependency — will now correctly re-evaluate when state changes propagate through the provider.
- `GraphNodeOss.tsx:300-306` destructures `impactedNodeIds` from context — also correctly re-renders.
- Context provider (`:1150-1151`) now passes state values directly instead of `.current` — consistent.

No new error paths. The change is purely about render timing. The `impacted &&` null guard at both sites protects against undefined results.

No new tests, which is reasonable here. This race condition (CLL resolving after sidebar is open) is timing-dependent and hard to cover with deterministic unit tests. The PR description documents thorough manual verification (Playwright + injected logs, 5s simulated delay, Snowflake-backed jaffle_shop). Existing suite passes (3664 tests).

All 4 ref access sites (2 reads, 2 writes) are migrated. `impactedNodeIdsRef` and `impactedColumnIdsRef` are fully removed — no stale references. Migration is complete and symmetric across both call sites (layout effect and `refreshLayout`).

Converting to state adds re-renders when impact analysis completes, but this is: (a) infrequent (once per impact analysis, guarded by `!column`), and (b) necessary for correctness. The re-render propagation is exactly what was missing.

This fix directly addresses the async race: CLL impact analysis resolving *after* the sidebar opens. With refs, the mutation was invisible to React's render cycle. With state, the Context Provider re-renders → `SchemaView`'s `useMemo` re-runs → amber highlights appear. The core race is resolved.

### Verification Results

- `pnpm type:check`: 1 pre-existing error in `csv/__tests__/excel.test.ts` (also present on `main`). **No new errors from this PR.**
- `pnpm lint`: ESLint not configured (Biome used instead per AGENTS.md). No issues from the diff.

---

## Verdict: GO ✅

All 8 validation passes clean. This is a minimal, focused, and correct fix for a real race condition. The `useRef` → `useState` conversion is the idiomatic React solution: refs don't participate in the render cycle, so downstream memos couldn't observe the mutation. State setters create new identities that propagate through Context, which is exactly what's needed here.

No blockers, no issues, no notes. Clean merge.

even-wei (approved):
Claude Code Review: No critical issues found. Clean ref→state migration that correctly fixes the sidebar race condition.

PR #1319, opened 2026-04-16
feat: make raw_code usage defensive for on-demand fetch

even-wei:
## Summary

Make all `raw_code` consumers handle the case where it's missing from lineage data, preparing for Phase I field stripping (DRC-3255).

- `NodeView.tsx`: Replace `raw_code` comparison with `changeStatus === "modified"` for code-change dot
- `NodeSqlView.tsx`: Show "No code available" when `raw_code` is undefined (both base and current)
- `SandboxViewOss.tsx` / `SandboxView.tsx`: Use `"-- No code available"` fallback
- `utils.ts`: Added TODO comment for checksum fallback removal (DRC-3254)

**Backward-compatible:** Works with both old `/info` (raw_code present) and new `/info` (raw_code stripped).

## Test plan

- [x] All 3681 tests pass
- [x] Lint clean, type check clean
- [ ] Manual: verify Code tab shows code when raw_code is present (current behavior)

Resolves DRC-3256

### DIFF
--- a/js/packages/ui/src/components/lineage/NodeSqlView.tsx
+++ b/js/packages/ui/src/components/lineage/NodeSqlView.tsx
@@ -125,6 +125,10 @@ export const NodeSqlView = ({
   const modelName =
     node.data.data.base?.name ?? node.data.data.current?.name ?? "";
 
+  // Defensive: show "No code available" only when raw_code is missing from
+  // all sources (forward-compatibility for when backend strips raw_code).
+  const hasCode = original !== undefined || modified !== undefined;
+
   return (
     <Box
       className="no-track-pii-safe"
@@ -136,7 +140,19 @@ export const NodeSqlView = ({
         setIsHovered(false);
       }}
     >
-      {isSingleEnv ? (
+      {!hasCode ? (
+        <Box
+          sx={{
+            display: "flex",
+            alignItems: "center",
+            justifyContent: "center",
+            height: "100%",
+            color: "text.secondary",
+          }}
+        >
+          No code available
+        </Box>
+      ) : isSingleEnv ? (
         <CodeEditor
           language="sql"
           value={original ?? ""}
@@ -196,7 +212,19 @@ export const NodeSqlView = ({
           </IconButton>
         </DialogTitle>
         <DialogContent>
-          {isSingleEnv ? (
+          {!hasCode ? (
+            <Box
+              sx={{
+                display: "flex",
+                alignItems: "center",
+                justifyContent: "center",
+                height: "100%",
+                color: "text.secondary",
+              }}
+            >
+              No code available
+            </Box>
+          ) : isSingleEnv ? (
             <CodeEditor
               language="sql"
               value={original ?? ""}
--- a/js/packages/ui/src/components/lineage/NodeView.tsx
+++ b/js/packages/ui/src/components/lineage/NodeView.tsx
@@ -601,11 +601,7 @@ export function NodeView<TNode extends NodeViewNodeData>({
   const { base, current } = node.data.data;
   const hasSchemaChanges =
     !isSingleEnv && isSchemaChanged(base?.columns, current?.columns) === true;
-  const hasCodeChanges =
-    !isSingleEnv &&
-    base?.raw_code != null &&
-    current?.raw_code != null &&
-    base.raw_code !== current.raw_code;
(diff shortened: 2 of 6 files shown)


even-wei:
# Code Review: PR #1319

**Branch:** `feature/drc-3256-on-demand-raw-code` → `main`
**Files reviewed:** 6
**Categories:** Frontend (lineage components), Tests, Business logic (utils)
**Passes run:** A, C, D, E, F

## Validation Results

### Pass A: Correctness & Logic — PASS (with NOTE)

**`NodeView.tsx` — `hasCodeChanges` substitution is semantically correct.**
The old logic compared `raw_code` strings directly; the new logic uses `changeStatus === "modified"` which is derived from dbt checksums (themselves computed from `raw_code`). All edge cases align:
- Added/removed nodes: both old and new return `false` ✓
- Unchanged nodes (same checksum): both return `false` ✓
- Changed nodes: checksum difference ↔ raw_code difference ✓

**`NodeSqlView.tsx` — `hasCode` guard is correct for the `undefined` case.**
`original !== undefined || modified !== undefined` properly detects when at least one side has code.

**NOTE:** The `hasCode` check uses strict `!== undefined`, but the Python backend serializes `None` as JSON `null` (Pydantic `BaseModel` default behavior). At runtime, `raw_code` could be `null` rather than `undefined` — and `null !== undefined` evaluates to `true` in JavaScript. If Phase I (DRC-3255) strips `raw_code` by setting it to `None` on the Python side (rather than deleting the key), this check would pass incorrectly, showing an empty editor instead of "No code available".

**Recommendation:** Consider using `original != null` (loose equality, catches both `null` and `undefined`) for maximum robustness:
```ts
const hasCode = original != null || modified != null;
```
*Severity:* **NOTE** — depends on the Phase I implementation approach. If stripping means deleting the field entirely, current code is fine.

**`SandboxView.tsx` / `SandboxViewOss.tsx` — fallback to `"-- No code available"` is appropriate.** The SQL comment prefix (`--`) is correct since the Sandbox editor is editable and the value is used as an initial SQL template.

### Pass B: Security — PASS

No security-relevant changes. No PII exposure, injection vectors, or auth changes.

- `changeStatus` field is well-established across the codebase (30+ references). Type is `string | undefined` on `NodeViewNodeData`, and `"modified"` is a known value set in `utils.ts:buildLineageGraph`.
- `NodeSqlViewNodeData.data.base?.raw_code` type (`string | undefined`) is consistent with the API types in `info.ts` and `cll.ts`.
- The "No code available" Box JSX is duplicated in `NodeSqlView.tsx` (inline view at ~L141 and dialog at ~L218) — this is acceptable given they're in separate render branches and extracting a component would add more complexity than it saves for 2 instances.

- `original ?? ""` and `modified ?? ""` correctly coalesce missing values for the CodeEditor/DiffEditor.
- Guard for non-model/non-snapshot resource types (`return "Not available"`) is still in place and runs before the `hasCode` check.
- `modelName` fallback chain (`base?.name ?? current?.name ?? ""`) handles all combinations.

### Pass E: Test Coverage — PASS (with NOTE)

The existing test was correctly updated: `"handles both base and current missing gracefully"` now asserts `"No code available"` text instead of an empty editor.

**NOTE:** The test uses `base: undefined, current: undefined` (entire objects missing), but the more realistic Phase I scenario is base/current objects present with `raw_code` field stripped:
```ts
data: {
  base: { name: "test_model" },      // no raw_code
  current: { name: "test_model" },    // no raw_code
}
```
This would also pass the current `hasCode` check, but having an explicit test for it would strengthen the contract. *Severity:* **NOTE** — nice-to-have, not blocking.

- The `utils.ts` TODO comment correctly references DRC-3254 as the issue for removing the client-side checksum fallback.
- No stale references introduced. All `raw_code` consumers outside the changed files are either type definitions (should stay), storybook fixtures (harmless), or backend code (not in scope for this PR).
- PR description accurately reflects all changes.

### Verification Results

Quality checks not run locally (draft PR, review-only pass). PR description confirms:
- All 3681 tests pass ✓
- Lint clean, type check clean ✓

## Verdict: GO

Zero BLOCKERs, zero ISSUEs. The defensive approach is well-reasoned: it decouples the UI from `raw_code` availability while maintaining backward compatibility. The two NOTEs (null vs undefined handling, and test scenario coverage) are worth considering before marking ready, but neither blocks merge.

PR #1324, opened 2026-04-17
fix(mcp): auto-approve checks after successful run in create_check

iamcxa:
## Summary

`_tool_create_check` runs checks but never updates `is_checked`, causing PR comment checklist to show ⏳ Pending for all checks even when runs succeed.

## Root Cause

`execute_preset_checks` (run.py:176) calls `run_should_be_approved()` / `schema_diff_should_be_approved()` after runs. The MCP server's `_tool_create_check` was missing this step.

## Fix

After run completes, call the same approval functions and update `is_checked` before `export_persistent_state()`. 14 lines, zero new dependencies.

## Test Plan

- [x] `pytest tests/test_mcp_server.py` — 114/114 pass
- [x] Manual: `/datarecce-local-dev` on jaffle_shop_golden PR#26 → checklist shows ✅ Passed instead of ⏳

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


tests/test_mcp_server.py
```diff
@@ -388,13 +388,17 @@ async def test_tool_create_check_idempotent_update(self, mcp_server):
...
+        first_call = mock_check_dao.update_check_by_id.call_args_list[0]
+        assert first_call[0][0] == check_id
+        patch_in = first_call[0][1]
         assert patch_in.name == "Updated name"
         assert patch_in.description == "Updated description"
-        assert patch_in.is_checked is None  # Not touching approval
+        # Second call: auto-approve (is_checked=True) after run succeeded
+        second_call = mock_check_dao.update_check_by_id.call_args_list[1]
+        assert second_call[0][0] == check_id
+        assert second_call[0][1].is_checked is True
```
iamcxa, line 407: **Finding (MEDIUM)** — Missing auto-approve test for newly-created checks.

`test_tool_create_check_basic` (the new-check path, ~L307-349) exercises the same `if run_executed and not run_error` auto-approve code but never asserts that `update_check_by_id` was called with `is_checked=True`. The idempotent-update test here correctly verifies auto-approve, but the create-new-check path is untested.

Suggestion: add an assertion in `test_tool_create_check_basic` like:
```python
mock_check_dao.update_check_by_id.assert_called_once_with(
    check_id, PatchCheckIn(is_checked=True)
)
```

iamcxa (commented):
## Review Summary

The auto-approve logic itself looks correct and well-placed. Two advisory observations below:

### Advisory A — Semantic gap with `run_should_be_approved`

`recce/mcp_server.py` ~L1791-1794: `execute_preset_checks` in `run.py` uses `run_should_be_approved(run)` (which inspects result content), while this PR uses a blanket `is_checked=True` for any successful run. The PR body documents this as a "PM decision" which is valid, but consider adding a brief code comment referencing the deliberate difference from `run_should_be_approved` so future readers don't try to "fix" the inconsistency.

### Advisory B — Missing auto-approve coverage in `test_tool_create_check_metadata_run_for_schema_diff`

`tests/test_mcp_server.py` ~L404-448: The metadata-path test also doesn't verify auto-approve (`is_checked=True`) after a successful run. Lower priority since Finding 1 (inline) covers the primary gap, but worth noting for completeness.

tests/test_mcp_server.py
iamcxa: Fixed in 049c54c4. Added auto-approve assertion to `test_tool_create_check_basic` — verifies `update_check_by_id` is called once with `is_checked=True` after a successful run.

Also addressed Advisory B in the same commit: added the same assertion to `test_tool_create_check_metadata_run_for_schema_diff`.

iamcxa:
All review feedback addressed in 049c54c4:

- **Finding 1** (inline): Added auto-approve assertion to `test_tool_create_check_basic` — thread resolved.
- **Advisory A**: Added code comment in `mcp_server.py` clarifying the deliberate difference from `run_should_be_approved()`.
- **Advisory B**: Added auto-approve assertion to `test_tool_create_check_metadata_run_for_schema_diff`.

All 114 MCP tests pass.

wcchang1115:
## Code Review — PR #1324

### Summary
Small, focused fix: adds auto-approval (`is_checked=True`) inside `_tool_create_check` after a successful run so MCP-created checks render as ✅ in the PR checklist. Minimal diff, tests and lint pass locally (`pytest tests/test_mcp_server.py` → 114/114, `make flake8` clean).

### Findings

#### [Warning] Approval semantics diverge from CLI (`execute_preset_checks`)
**File:** `recce/mcp_server.py:1791-1798`
**Issue:** The new block blanket-approves any successful run — including `ROW_COUNT_DIFF` runs where base/curr row counts differ, and `SCHEMA_DIFF` runs with actual column changes. The CLI path in `recce/run.py:162-176` deliberately gates approval on content:
- `run_should_be_approved()` (run.py:112-126) only approves `ROW_COUNT_DIFF` when base == curr.
- `schema_diff_should_be_approved()` (run.py:67-109) only approves `SCHEMA_DIFF` when the lineage diff is empty.

The code comment calls this intentional per PM decision, so this is not a bug — but it means `Passed` in the MCP checklist carries weaker semantics than `Passed` from `recce run` preset checks (ran without error vs. actually matched). Worth double-checking this is the intent (and documenting it in the MCP tool description so agent users don't assume "Passed" implies no diff).

#### [Warning] Auto-approve overrides prior user "un-approval" silently
**File:** `recce/mcp_server.py:1791-1798`
**Issue:** In the idempotent path (lines 1745-1748), if a user has manually unchecked a previously approved check (set `is_checked=False`), re-invoking `create_check` with matching `(type, params)` will silently flip it back to `True` whenever the re-run succeeds. There is no "user-unchecked" signal preserved.
**Suggestion:** Consider only auto-approving when `created=True` or when the existing check's `is_checked` was not explicitly `False`. Minimally, surface this in the tool description so an agent knows re-running overrides manual state.

#### [Warning] Test gap — failure path does not assert absence of approval call
**File:** `tests/test_mcp_server.py:458-489` (`test_tool_create_check_run_failure`)
**Issue:** The test verifies `run_executed is False` and `run_error` is populated, but does not assert `mock_check_dao.update_check_by_id.assert_not_called()`. Given the new branch is specifically gated by `if run_executed and not run_error`, the negative case is the one worth locking down. If someone later drops the guard, this test would still pass.
**Suggestion:** Add `mock_check_dao.update_check_by_id.assert_not_called()` to the failure test.

#### [Nit] Code comment references PM decision without a link
**File:** `recce/mcp_server.py:1791-1796`
**Issue:** The comment says "intentional per PM decision" but does not point to the ticket/thread. A future reader hitting the divergence in #1 above has no way to verify the decision still stands.
**Suggestion:** Add the Linear issue ID or Slack thread reference.

### Verification Evidence
- `pytest tests/test_mcp_server.py` → 114 passed
- `make flake8` → clean
- Manually traced both branches (idempotent update and newly-created) through `update_check_by_id`; both correctly reach the auto-approve path.

### Verdict
✅ Approved with minor nits. The fix is correct for the stated goal, minimal in surface area, and well-tested for the happy paths. The semantic divergence from `run.py` is the one thing worth confirming is intentional before merging.

wcchang1115 (approved):
Claude Code Review: No critical issues found. Minor warnings and a test gap posted as a review comment.

PR #1329, opened 2026-04-21
fix(profile): handle empty tables in profile_diff

danyelf:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

bug

Fixes `profile_diff` when a model has zero rows on one side. The generated SQL divided by `count(*)`, so an empty table raised a database "division by zero" error (Snowflake 100051), which Recce surfaced as the confusing message "No profile diff result due to the model is empty."

An empty model on one side is useful diagnostic information — the user should see `row_count` go from K to 0, not an opaque error.

Changes in `recce/tasks/profile.py`:
- Wrap the two `count(*)` denominators (`not_null_proportion`, `distinct_proportion`) in `nullif(..., 0)` so empty tables yield `NULL` proportions instead of dividing by zero.
- Wrap `is_unique` in `case when count(*) = 0 then null else ... end` so it doesn't vacuously return `TRUE` on zero rows.

Adds two tests in `tests/tasks/test_profile.py`: `test_profile_diff_current_empty` and `test_profile_diff_both_empty`.

None linked.

- Discovered against Snowflake on `order_returns` (181 base → 0 current). Verified manually on Snowflake jaffle_shop with this branch.
- The Snowflake 100051 error handler at `recce/tasks/profile.py` becomes a safety net after this change — kept as-is.
- All 8 tests in `tests/tasks/test_profile.py` pass against DuckDB (6 existing + 2 new).

Yes — profile diff now renders row counts and NULL proportions when a model is empty on one side, instead of failing with "No profile diff result due to the model is empty."

```release-note
Fix profile diff to handle empty tables: emit NULL proportions instead of failing with a divide-by-zero error when a model has zero rows on one side.
```

### DIFF
--- a/recce/tasks/profile.py
+++ b/recce/tasks/profile.py
@@ -41,18 +41,25 @@ PROFILE_COLUMN_JINJA_TEMPLATE = r"""
 
 {# General Agg ------------------------------------------- #}
 {%- set agg_row_count = 'cast(count(*) as ' ~ dbt.type_bigint() ~ ')' -%}
+{# Wrap count(*) denominators in nullif so empty tables yield NULL proportions
+   instead of raising a database "division by zero" error. #}
 {%- set agg_not_null_proportion =
         'sum(case when ' ~ adapter.quote(column_name) ~ ' is null '
         ~ 'then 0 '
         ~ 'else 1 end) / '
-        ~ 'cast(count(*) as ' ~ dbt.type_numeric() ~ ')'
+        ~ 'nullif(cast(count(*) as ' ~ dbt.type_numeric() ~ '), 0)'
 -%}
 {%- set agg_distinct_proportion =
         'count(distinct ' ~ adapter.quote(column_name) ~') / '
-        ~ 'cast(count(*) as ' ~ dbt.type_numeric() ~ ')'
+        ~ 'nullif(cast(count(*) as ' ~ dbt.type_numeric() ~ '), 0)'
 -%}
 {%- set agg_distinct_count = 'count(distinct ' ~ adapter.quote(column_name) ~ ')' -%}
-{%- set agg_is_unique =      'count(distinct ' ~ adapter.quote(column_name) ~ ') = count(*)' -%}
+{# is_unique is undefined on an empty table — emit null rather than vacuously
+   "0 = 0 = true". #}
+{%- set agg_is_unique =
+        'case when count(*) = 0 then null '
+        ~ 'else count(distinct ' ~ adapter.quote(column_name) ~ ') = count(*) end'
+-%}
 {%- set agg_min =            'cast(null as ' ~ dbt.type_string() ~ ')' -%}
 {%- set agg_max =            'cast(null as ' ~ dbt.type_string() ~ ')' -%}
 {%- set agg_avg =            'cast(null as ' ~ dbt.type_numeric() ~ ')' -%}
--- a/tests/tasks/test_profile.py
+++ b/tests/tasks/test_profile.py
@@ -56,6 +56,49 @@ def test_profile_diff_with_selected_columns(dbt_test_helper):
     assert len(run_result.base.data) == 2
 
 
+# Empty table has no rows — the SQL used to divide by count(*) and fail with
+# "division by zero". After the nullif() fix, an empty side produces one row
+# per column with row_count=0 and NULL proportions.
+csv_data_empty = """
+        customer_id,name,age
+        """
+
+
+def test_profile_diff_current_empty(dbt_test_helper):
+    dbt_test_helper.create_model("customers", csv_data_base, csv_data_empty)
+    params = dict(model="customers")
+    task = ProfileDiffTask(params)
+    run_result = task.execute()
+
+    # Still returns a row per column on both sides.
+    assert len(run_result.base.data) == 3
+    assert len(run_result.current.data) == 3
+
+    # Row counts are populated: base has 3, current has 0.
+    def _row_count(df, col_name):
+        col_names = [c.name for c in df.columns]
+        idx_name = col_names.index("column_name")
+        idx_rc = col_names.index("row_count")
+        for row in df.data:
+            if row[idx_name] == col_name:
+                return row[idx_rc]
+        return None
+
+    assert _row_count(run_result.base, "customer_id") == 3
+    assert _row_count(run_result.current, "customer_id") == 0
+
+
+def test_profile_diff_both_empty(dbt_test_helper):
+    dbt_test_helper.create_model("customers", csv_data_empty, csv_data_empty)
+    params = dict(model="customers")
+    task = ProfileDiffTask(params)
+    run_result = task.execute()
+
+    # No crash; one row per column on each side.
+    assert len(run_result.base.data) == 3
+    assert len(run_result.current.data) == 3
+
+
 def test_validator():
     from recce.tasks.profile import ProfileCheckValidator
 


danyelf:
# Code Review: PR #1329

**Files reviewed:** 2 (`recce/tasks/profile.py`, `tests/tasks/test_profile.py`)
**Categories:** Business logic + Tests
**Passes run:** A, C, D, E, F, G

## Validation Results

Trace through the SQL on an empty table:
- `nullif(cast(count(*) as NUMERIC), 0)` → `nullif(0, 0)` → `NULL` → both proportion columns become `NULL` instead of dividing by zero. ✓
- `case when count(*) = 0 then null else count(distinct x) = count(*) end` → `NULL` on empty, correctly avoiding the vacuous `0 = 0 = TRUE` result. ✓
- Non-empty paths are unchanged (denominator is non-zero, so `nullif(n, 0) = n`).

`is_unique` transitions from always-boolean to nullable-boolean on the empty-table path. This nullability is **not new to downstream consumers**:
- `recce/tasks/profile.py:72` already sets `agg_is_unique = 'null'` for struct types.
- `recce/tasks/profile.py:116` already sets `agg_is_unique = 'null'` for BigQuery array types.

So `DataFrame.from_agate()`, `ProfileDiffResultDiffer`, and the frontend (`ProfileResultView.tsx`) must already tolerate a null `is_unique`. No contract change.

The Snowflake 100051 error handler at `recce/tasks/profile.py:245-247` is correctly retained as a defense-in-depth safety net (confirmed in PR description). Good call — the SQL fix is the primary defense, but handler catches any dialect-specific edge cases the template might miss.

### Pass E: Test Coverage & Quality — PASS (with one minor note)
Two new tests exercise the empty-table path:
- `test_profile_diff_current_empty` asserts `base.data`/`current.data` lengths and row counts (3 → 0). ✓
- `test_profile_diff_both_empty` asserts no crash + one row per column. ✓

The existing `test_profile_column_jinja_template` already parses the rendered SQL against 10 dialects via `sqlglot`, so the new `nullif(...)` and `case when ... end` expressions are syntactically validated on all supported platforms. ✓

**NOTE:** Neither new test asserts that `not_null_proportion`, `distinct_proportion`, or `is_unique` are actually `NULL` on the empty side — they only confirm the query doesn't crash and row counts are present. If the SQL ever regressed to a literal `0` (e.g., someone reverts the `case when`), row-count assertions wouldn't catch it. Not blocking — the change is tight enough that a regression would be obvious — but a one-liner assertion against those three columns on the empty side would tighten the test.

- Template-only change; no caller signatures modified.
- No breaking behavior change for non-empty tables.
- PR description accurately describes the fix.

`nullif()` wrapping and `case when count(*) = 0` add negligible overhead (the aggregates run once per column, not per row).

### Verification Results
Ran locally on `fix/profile-empty-table` worktree:
```
tests/tasks/test_profile.py::test_profile PASSED
tests/tasks/test_profile.py::test_profile_with_selected_columns PASSED
tests/tasks/test_profile.py::test_profile_diff PASSED
tests/tasks/test_profile.py::test_profile_diff_with_selected_columns PASSED
tests/tasks/test_profile.py::test_profile_diff_current_empty PASSED
tests/tasks/test_profile.py::test_profile_diff_both_empty PASSED
tests/tasks/test_profile.py::test_validator PASSED
tests/tasks/test_profile.py::test_profile_column_jinja_template PASSED
============================== 8 passed in 2.99s ===============================
```
`flake8 recce/tasks/profile.py tests/tasks/test_profile.py` — clean.

CI status at review time: CodeQL, Flake8, DCO, all smoke-test matrices PASSING. Python/dbt matrix jobs still in progress but no failures so far.

## Verdict: GO

Clean, minimal, well-scoped bug fix. The diagnosis is right (SQL divide-by-zero, not an empty-model error), the fix is at the right layer (template, not Python), and the Snowflake-specific error handler is correctly preserved as a safety net rather than replaced. Existing cross-dialect SQL parsing test automatically covers the new syntax.

### Notes
1. Consider adding assertions that `not_null_proportion`, `distinct_proportion`, and `is_unique` are `None` on the empty side in `test_profile_diff_current_empty` — tightens the test against future regressions where the denominator fix gets partially reverted.

even-wei:
# Code Review: PR #1329

**Files reviewed:** 2 (`recce/tasks/profile.py`, `tests/tasks/test_profile.py`)
**Categories:** Business logic (Jinja-templated SQL) + Tests
**Passes run:** A, C, D, E, F, G

Reviewer focus (per request): correctness of the Jinja template.

## Validation Results

Traced the Jinja template end-to-end for `column_name="x"`, `column_type="integer"`. All three modified `{%- set ... -%}` blocks concatenate into well-formed SQL:

- `agg_not_null_proportion` →
  `sum(case when "x" is null then 0 else 1 end) / nullif(cast(count(*) as numeric), 0)` ✓
- `agg_distinct_proportion` →
  `count(distinct "x") / nullif(cast(count(*) as numeric), 0)` ✓
- `agg_is_unique` →
  `case when count(*) = 0 then null else count(distinct "x") = count(*) end` ✓

Whitespace control markers (`{%-` / `-%}`) and trailing spaces inside the string literals are preserved, so the concatenated SQL is syntactically correct and idiomatic.

Semantics on an empty table (`count(*) = 0`):
- `sum(case …) / nullif(0, 0)` → `NULL / NULL` → `NULL` ✓
- `count(distinct x) / nullif(0, 0)` → `0 / NULL` → `NULL` ✓
- `case when count(*) = 0 then null …` → `NULL` ✓ (fixes the prior vacuous `0 = 0 = TRUE`)

Semantics on a non-empty table:
- Unchanged behavior: `nullif(N, 0)` where `N > 0` returns `N`, so division is identical to before.
- `is_unique` falls through the `case` to the original `count(distinct x) = count(*)` — unchanged semantics ✓

- `dbt.type_numeric()` is a dbt macro that resolves per-adapter (e.g., `numeric` on Snowflake, `DECIMAL` on BigQuery) — already used elsewhere in this template, consistent here.
- `adapter.quote(column_name)` usage is consistent with existing `agg_distinct_count` / `agg_min` / `agg_max` patterns.
- The struct-type override (`{%- set agg_is_unique = 'null' -%}`) and BigQuery-array override later in the template both rebind `agg_is_unique` wholesale — the new multi-line value is safely replaced without leakage.

- The Snowflake `100051` handler in `_profile_column` is intentionally retained as a safety net, as noted in the PR description.
- All three fix points are defensively wrapped: numerators that produce `NULL` on empty (`sum()` returns `NULL`; `count(distinct)` returns `0`) interact correctly with `nullif()` denominators.
- No new exception paths introduced; no resource-leak or cancellation concerns.

### Pass E: Test Coverage & Quality — PASS with notes

Tests `test_profile_diff_current_empty` and `test_profile_diff_both_empty` exercise both code paths (one-empty + both-empty) via DuckDB. They confirm no crash and that `row_count` is populated as expected. See Notes below for strengthening suggestions already raised by Copilot.

- No signature changes, no callers to update.
- Behavior change is documented in the PR description and matches the `release-note` block.

`nullif(cast(count(*) as numeric), 0)` adds one function call per aggregate; negligible. The new `case when count(*) = 0 then null else … end` evaluates `count(*)` twice, but query planners will deduplicate. No N+1 or unbounded-scan concerns.

### Verification Results

- **CI:** 19/19 required checks passing; the only `FAIL` in `gh pr checks` is `notify-staging-failure` with conclusion `skipping` (not a real failure — it's a conditional notifier).
- **Codecov:** comment posted; no regression flagged.

## Verdict: GO

No blockers. No issues.

### Notes

These are non-blocking observations. Items 1 and 2 echo findings already raised by Copilot — included here for completeness so the test suite most precisely reflects the intent of the fix.

1. **Test assertions could target the fix semantics directly** (`tests/tasks/test_profile.py:88`).
   The new tests verify `row_count == 0` but don't assert the three values that were actually fixed: `not_null_proportion`, `distinct_proportion`, `is_unique` should all be `NULL` on the empty side. Without these, a future regression where (say) `is_unique` reverts to `TRUE` on empty tables would still pass the suite.

2. **Missing symmetric case: base empty + current non-empty** (`tests/tasks/test_profile.py:99`).
   `test_profile_diff_current_empty` covers the `current=empty` side; no test covers the mirror case. The `ProfileDiffTask.execute()` loop runs the same SQL generator for both sides, so symmetry is almost certainly preserved — but a two-line test would lock it in.

3. **Snowflake coverage is manual-only.** The originating bug was on Snowflake (error 100051); CI tests only cover DuckDB. The PR description confirms manual verification on Snowflake jaffle_shop. Acceptable for a bug fix, but worth noting that future changes to this template won't have automated Snowflake coverage.

Neither the main fix nor the Jinja template has any correctness concerns. Approving.

even-wei (approved):
Claude Code Review: No critical issues found. Jinja template correctness verified — all three modified set blocks concatenate into well-formed SQL and handle the empty-table case correctly.

PR #1331, opened 2026-04-22
refactor(api): strip raw_code from /api/cll payload

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

refactor / payload optimization

Strips `raw_code` from the `/api/cll` response payload. `CllNode.raw_code` is populated from the dbt manifest for server-side CLL computation (read in `DbtAdapter.get_cll_cached` on two paths) but is read by zero frontend consumers off a CLL-sourced node — frontend only reads `raw_code` from `/api/models/{id}` responses (NodeSqlView, SandboxView, SandboxViewOss).

This mirrors the precedent set by `/info` (lineage diff), which already strips `raw_code`, `config`, `columns`, `checksum` in `recce-cloud-infra`'s `api_server/core/lineage/diff.py`.

Implementation: mark `raw_code: Optional[str] = Field(default=None, exclude=True)` on `CllNode`. `exclude=True` applies to every `model_dump`, so the field is dropped from both the FastAPI response and the SQLite CLL cache payload, while the in-memory attribute remains available for the two server-side reads.

Cloud's `/cll` endpoint is a passthrough (S3 map fast path + OSS proxy fallback), so fixing OSS fixes both.

N/A — payload optimization task.

- Two existing assertions in `tests/util/test_cll_cache.py` asserted `raw_code` round-trips through the CLL cache. Updated to assert `raw_code is None` after deserialization — this is the new intended contract (the cache stores post-computation `CllData` and no consumer reads raw_code from cached entries).
- New test `test_cll_response_excludes_raw_code` serializes via `CllOutput.model_dump_json()` (exactly what FastAPI uses with `response_model=CllOutput`) and asserts no node carries `raw_code`. Also positively asserts the in-memory attribute is still populated.

```release-note
NONE
```

### DIFF
--- a/recce/models/types.py
+++ b/recce/models/types.py
@@ -230,7 +230,11 @@ class CllNode(BaseModel):
     name: str
     package_name: str
     resource_type: str
-    raw_code: Optional[str] = None
+    # raw_code is populated from the manifest for server-side CLL computation
+    # (see DbtAdapter.get_cll_cached) but is excluded from serialization — no
+    # frontend consumer reads it off a CLL-sourced node, and it bloats the
+    # /api/cll payload significantly.
+    raw_code: Optional[str] = Field(default=None, exclude=True)
     source_name: Optional[str] = None
 
     # change analysis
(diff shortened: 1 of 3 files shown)


gcko (approved):
:shipit:

PR #1335, opened 2026-04-23
feat(cli): emit per_node.db in recce init --cloud [PR 2/3]

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`feat` — wires the emitter into `recce init --cloud` and adds the cache.db handling.

**Stacked on top of #1334** (PR 1/3 — the pure emitter module). Please review #1334 first; this PR shows the merge diff against that branch.

Wires `PerNodeDbWriter` from #1334 into `recce init --cloud`:

- After `build_full_cll_map()` succeeds, emits `per_node.db` for both `base` and `current` envs from the loaded manifests + catalogs
- Uploads `per_node.db` via a new `per_node_db_url` presigned-URL key (graceful-degraded: missing key logs a warning, doesn't fail the command)
- Cloud mode's `cll_cache.db` now lives in `tempfile.mkdtemp(prefix="recce-cll-")` instead of `~/.recce/cll_cache.db`. Purpose: keep the warm-cache perf (`build_full_cll_map` re-uses previous CLL slices) while guaranteeing the cache DB never ends up on S3 or on the ECS task host after shutdown
- **Removed the cloud-mode `cll_cache_upload_url` upload path** entirely (cache.db is not a cloud artifact)
- `shutil.rmtree` cleanup of the tempdir on success via `try/finally`

Local (non-cloud) `recce init` behavior is unchanged: still writes `~/.recce/cll_cache.db` (or whatever `--cache-db` points to), still doesn't emit `per_node.db`.

Part of [DRC-3295](https://linear.app/recce/issue/DRC-3295) (epic [DRC-3294](https://linear.app/recce/issue/DRC-3294), project "Lineage API Data Transfer Optimization" Phase III).

- Cloud-side prerequisite: the API server needs to add `per_node_db_url` to the response of `get_upload_urls_by_session_id` (same presigned-URL pattern as `cll_map_url`, content type `application/octet-stream`). CLI handles missing key gracefully, so this PR can land independently of that work.
- 7 new tests in `tests/test_cli_per_node_db.py`:
  - Cloud happy path (PUTs verified)
  - Missing-key graceful degradation
  - Scratch-dir cleanup after success
  - Local-mode non-regression
  - **Contract parity vs. `DbtAdapter.get_model()`** on the `tests/manifest.json` + `catalog.json` jaffle_shop fixture
  - Mode-switching safety
- One test in `tests/test_cli_cache.py` retargeted (`test_init_cloud_upload_partial_failure_shows_warning`) since the `cll_cache_url` upload branch is now deleted
- Full suite: 1106 passed, 5 pre-existing SPA failures (unchanged from #1334 — `test_spa_route_*` requires a built frontend that's not in the worktree)
- Review focus: correctness of the cloud / local branch split around `cache_db` path; `try/finally` cleanup semantics; contract-test coverage

**Cloud users (implicit):** running `recce init --cloud` now uploads an extra `per_node.db` artifact and no longer uploads `cll_cache.db`. Consumers in Recce Cloud will begin reading `per_node.db` once the Cloud-side URL key is added. Nothing breaks in the meantime — the `cll_cache.db` upload path was already only read via the `cll_cache_url` key, which Cloud will continue to provide transparently during migration (CLI just ignores it).

**Local users (no change):** local `recce init` behavior is unchanged.

NONE (user-visible), but the cloud artifact set does change — covered by the cross-repo coordination note above.

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -1,6 +1,6 @@
 import os
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 import click
 
@@ -318,6 +318,7 @@ def init(cache_db, **kwargs):
 
     import json
     import logging
+    import shutil
     import tempfile
     import time
 
@@ -325,9 +326,15 @@ def init(cache_db, **kwargs):
     from rich.console import Console
     from rich.progress import Progress
 
+    from recce import __version__
     from recce.adapter.dbt_adapter import DbtAdapter
     from recce.core import load_context
     from recce.util.cll import _DEFAULT_DB_PATH, CllCache, get_cll_cache, set_cll_cache
+    from recce.util.per_node_db import SCHEMA_VERSION as PER_NODE_DB_SCHEMA_VERSION
+    from recce.util.per_node_db import (
+        PerNodeDbWriter,
+        extract_rows_from_artifacts,
+    )
 
     logger = logging.getLogger("recce")
     console = Console()
@@ -343,9 +350,15 @@ def init(cache_db, **kwargs):
     cloud_client = None
     cloud_org_id = None
     cloud_project_id = None
+    # In cloud mode, cll_cache.db and per_node.db are written to a tempdir that
+    # is scoped to this invocation. cll_cache.db keeps its warm-cache perf but
+    # never leaves the container; per_node.db is uploaded then the dir is GC'd.
+    cloud_scratch_dir: Optional[Path] = None
 
     if is_cloud:
         from recce.util.recce_cloud import RecceCloud, RecceCloudException
+
+        cloud_scratch_dir = Path(tempfile.mkdtemp(prefix="recce-cll-"))
 
         cloud_token = kwargs.get("cloud_token") or kwargs.get("api_token")
         if not cloud_token:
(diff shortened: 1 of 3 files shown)


even-wei:
**Heads-up — pushing a revert commit shortly.**

I had this PR suppress the `cll_cache.db` upload in cloud mode based on a misread of the design. `cll_cache.db` is actually load-bearing: it enables cross-session warm-cache reuse for OSS users and cloud sessions alike (downloaded at session init, uploaded at completion, reused by `build_full_cll_map` to skip already-computed CLL slices).

Incoming fix:
- `cache_db` default path restored to `~/.recce/cll_cache.db` (or `--cache-db` override)
- `cll_cache_upload_url` upload block restored
- `per_node.db` stays tempdir-scoped (it IS genuinely a throwaway per-task artifact)
- Tests updated to assert `cll_cache.db` IS uploaded in cloud mode

Fix commit inbound. Please hold review until it lands — or review just the `per_node.db` emission path if it helps.

even-wei:
Consolidated into #1334 for commit-by-commit review. PR 2's two commits (`cb8d74f1` CLI wire-in + `eff86d6e` cache.db fix) are in #1334 now via fast-forward. GitHub auto-marked this as merged when the source branch was absorbed — that's accurate from its perspective. Deleting the source branch.

PR #1336, opened 2026-04-23
feat(cli): emit info.json and lineage_diff.json in recce init --cloud [DRC-3296]

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

feat

Adds `info.json` and `lineage_diff.json` to the cloud-mode precompute artifacts emitted by `recce init --cloud`. These serve Cloud's `/info` and `/select` endpoints directly from S3 without re-computing lineage from raw dbt artifacts on every request.

Both files are derived purely from the adapter's already-loaded manifests and catalogs:
- `info.json` — the artifact-sourced portion of the `/info` response (adapter_type + merged lineage), built via `build_merged_lineage` (same helper the OSS `/api/info` endpoint uses). Cloud-specific fields (org_id, project_id, cloud_mode, pull_request, etc.) are injected by the Cloud API at request time.
- `lineage_diff.json` — the raw `LineageDiff` (base / current / diff) emitted by `adapter.get_lineage_diff()`.

Both are raw JSON (no compression, per scope decision). No new dependencies; no SQL execution beyond what `recce init` already does.

**Upload contract**: `info.json` uploaded via `info_url`, `lineage_diff.json` via `lineage_diff_url` keys on `get_upload_urls_by_session_id`. **Graceful degradation**: if either key is missing in the upload URL response, we log a warning and continue (exit 0) — keeps old CLI + new Cloud (and vice-versa) compatible during the rollout.

**Scratch dir**: cloud mode uses a `recce-metadata-*` tempdir (distinct prefix from A1's `recce-cll-*`), cleaned up via `shutil.rmtree` on successful upload, preserved on failure for debugging.

Resolves DRC-3296 — https://linear.app/recce/issue/DRC-3296

- Parallel to A1 (#1334). Expect a small `recce/cli.py` rebase conflict in the cloud-upload block when one merges first — trivial to resolve (both PRs append upload branches side by side).
- **Cloud-side follow-up**: `info_url` and `lineage_diff_url` keys need to be added to `get_upload_urls_by_session_id` in the cloud backend. This PR ships the CLI-side emit+upload logic with graceful missing-key handling; the cloud-side work is tracked separately and not in this PR's scope.
- The emitter (`recce/util/info_emitter.py`) reuses `build_merged_lineage` and `adapter.get_lineage_diff()` so the JSON shapes stay in sync with the live-instance `/api/info` and `/api/select` responses.

**Tests**:
- Unit tests (`tests/test_info_emitter.py`, 9 tests) — adapter-type resolution, payload shape/contract, atomic write, .tmp cleanup on failure, both-files write, parent-dir creation, pydantic round-trip.
- CLI integration tests (`tests/test_cli_info_emitter.py`, 6 tests) — happy path both-URLs-present, missing info_url warn+continue, missing lineage_diff_url warn+continue, both-URLs-missing warn+no-upload, local-mode non-regression (no emit, no upload), metadata-scratch prefix distinctness + cleanup.
- Full suite: 1093 passed / 18 deselected. Only pre-existing `test_spa_route_*` failures remain (SPA build missing in OSS test env, unchanged from baseline).

NONE

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -330,6 +330,7 @@ def init(cache_db, **kwargs):
     from recce.adapter.dbt_adapter import DbtAdapter
     from recce.core import load_context
     from recce.util.cll import _DEFAULT_DB_PATH, CllCache, get_cll_cache, set_cll_cache
+    from recce.util.info_emitter import emit_info_and_lineage_diff
     from recce.util.per_node_db import SCHEMA_VERSION as PER_NODE_DB_SCHEMA_VERSION
     from recce.util.per_node_db import (
         PerNodeDbWriter,
@@ -350,7 +351,6 @@ def init(cache_db, **kwargs):
     cloud_client = None
     cloud_org_id = None
     cloud_project_id = None
-
     if is_cloud:
         from recce.util.recce_cloud import RecceCloud, RecceCloudException
 
@@ -680,12 +680,38 @@ def init(cache_db, **kwargs):
 
     # In cloud mode, emit per_node.db — a pure-artifact SQLite that Cloud
     # streams to serve lineage without proxying to an ephemeral Recce instance.
-    # The scratch dir is always cleaned up, even on upload failure, so
-    # long-lived Cloud deploys don't accumulate recce-per-node-* directories
-    # in /tmp on retries.
+    # Also emit info.json + lineage_diff.json — pre-computed artifacts that
+    # serve Cloud's /info and /select endpoints without re-computing lineage
+    # from raw dbt artifacts on every request.
+    # Both scratch dirs are always cleaned up, even on upload failure, so
+    # long-lived Cloud deploys don't accumulate recce-per-node-* or
+    # recce-metadata-* directories in /tmp on retries.
     if is_cloud:
         per_node_scratch = Path(tempfile.mkdtemp(prefix="recce-per-node-"))
+        metadata_scratch = Path(tempfile.mkdtemp(prefix="recce-metadata-"))
         try:
+            # Emit info.json + lineage_diff.json up front — these are pure
+            # artifact derivations and don't depend on the upload URL keys.
+            info_path: Optional[Path] = metadata_scratch / "info.json"
+            lineage_diff_path: Optional[Path] = metadata_scratch / "lineage_diff.json"
+            console.print("\n[bold]Emitting lineage metadata...[/bold]")
+            t_meta_start = time.perf_counter()
+            try:
+                emit_info_and_lineage_diff(dbt_adapter, info_path, lineage_diff_path)
+                meta_elapsed = time.perf_counter() - t_meta_start
+                info_size_kb = info_path.stat().st_size / 1024
+                lineage_diff_size_kb = lineage_diff_path.stat().st_size / 1024
+                console.print(
+                    f"  Metadata saved to [bold]{metadata_scratch}[/bold] "
+                    f"(info.json {info_size_kb:.1f} KB, lineage_diff.json {lineage_diff_size_kb:.1f} KB, "
+                    f"{meta_elapsed:.1f}s)"
+                )
+            except Exception as e:
+                logger.warning("[recce init] Failed to emit metadata artifacts: %s", e)
+                console.print(f"  [[yellow]Warning[/yellow]] Failed to emit metadata artifacts: {e}")
+                info_path = None
+                lineage_diff_path = None
+
             if cloud_client:
                 console.print("\n[bold]Uploading results to Cloud...[/bold]")
                 upload_failures: list[str] = []
@@ -847,6 +873,43 @@ def init(cache_db, **kwargs):
                     elif not cll_cache_upload_url:
                         logger.debug("No cll_cache_url in upload URLs — cache upload not supported yet")
 
+                    # Upload info.json and lineage_diff.json. Graceful
+                    # degradation: if Cloud hasn't added the info_url /
+                    # lineage_diff_url keys yet, log a warning and continue —
+                    # keeps old CLI + new Cloud (and vice versa) compatible.
+                    metadata_uploads = [
+                        ("info.json", info_path, "info_url"),
+                        ("lineage_diff.json", lineage_diff_path, "lineage_diff_url"),
+                    ]
+                    for display_name, local_path, url_key in metadata_uploads:
+                        metadata_upload_url = upload_urls.get(url_key)
+                        if metadata_upload_url and local_path is not None and local_path.is_file():
+                            try:
+                                with open(local_path, "rb") as f:
+                                    resp = requests.put(
+                                        metadata_upload_url,
+                                        data=f,
+                                        headers={"Content-Type": "application/json"},
+                                        timeout=_UPLOAD_TIMEOUT,
+                                    )
+                                if resp.status_code in (200, 204):
+                                    size_kb = local_path.stat().st_size / 1024
+                                    console.print(f"  Uploaded {display_name} ({size_kb:.1f} KB)")
+                                else:
+                                    upload_failures.append(display_name)
+                                    console.print(
+                                        f"  [[yellow]Warning[/yellow]] Failed to upload {display_name}: "
+                                        f"HTTP {resp.status_code}"
+                                    )
+                            except requests.RequestException as e:
+                                upload_failures.append(display_name)
+                                console.print(f"  [[yellow]Warning[/yellow]] Failed to upload {display_name}: {e}")
+                        elif not metadata_upload_url:
+                            console.print(
+                                f"  [[yellow]Warning[/yellow]] No {url_key} in upload URLs "
+                                "(Cloud server may need update)"
+                            )
+
                     if upload_failures:
                         console.print(
                             f"[bold yellow]Cloud upload completed with warnings[/bold yellow] "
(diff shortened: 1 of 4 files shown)


even-wei:
# Code Review: PR #1336

**Files reviewed:** 4 (`recce/util/info_emitter.py`, `recce/cli.py`, `tests/test_info_emitter.py`, `tests/test_cli_info_emitter.py`)
**Categories:** Business logic, Tests

## Validation Results

### Pass A: Correctness & Logic — PASS (minor)
- Emitter correctly reuses `build_merged_lineage` (same helper used by the server `/api/info` endpoint at `recce/server.py:597`), and `model_dump(exclude_none=True, by_alias=True)` matches the server's serialization exactly — wire-format drift risk is contained.
- `_resolve_adapter_type` preference order (manifest metadata → `adapter.type()` → None) is sensible; falls back cleanly when manifests are absent.
- `NodeDiff` / `LineageDiff` contract from `recce/models/types.py` verified — lineage_diff.json round-trip test is valid.

- Upload-URL keys `info_url` / `lineage_diff_url` match sibling PR `recce-cloud-infra#1243` (verified against `api_server/apis/sessions_api.py`). Contract is aligned.
- `adapter.get_lineage_diff()` signature matches `BaseAdapter.get_lineage_diff` in `recce/adapter/base.py:23`.
- Test mocks (`_make_mock_adapter`, `_make_adapter`) correctly mirror the real adapter attribute surface used by the emitter.

### Pass D: Error Handling & Edge Cases — FAIL

**ISSUE** — `recce/cli.py:709-713` combined with `cli.py:884-911` — Partial-emit failure silently produces a misleading success message. If `emit_info_and_lineage_diff` raises after writing `info.json` but before `lineage_diff.json` (e.g., disk full on the second write, or any exception during the second pydantic `model_dump`), both `info_path` and `lineage_diff_path` are reset to `None`. Later, in the upload loop, when `local_path is None` but `metadata_upload_url` is truthy, neither branch of the `if/elif` fires — the skip is silent and nothing is appended to `upload_failures`. The terminal summary then prints `Cloud upload complete.` even though metadata was never uploaded. The emit-time warning is the only signal, and a user or operator scanning only the final status line will miss it. Fix options: (a) append both filenames to `upload_failures` inside the emit-except block, or (b) add an `else` arm covering "URL present but local file missing" that warns and appends.

**NOTE** — `recce/util/info_emitter.py:92` — `json.dumps(..., default=str)` silently coerces unexpected non-JSON-serializable types (`datetime`, `UUID`, `Decimal`, etc.) to their `str()` form. Since `LineageDiff.base`/`current` are untyped `dict`, anything deep in `manifest_metadata` could be stringified without warning. For the current contract (round-trip to `LineageDiff.model_validate`, which also accepts dicts-of-anything) this is harmless, but it hides future data-type regressions rather than surfacing them.

### Pass E: Test Coverage & Quality — PASS (one gap)

**NOTE** — No test exercises the partial-emit failure path (first file written, second write raises). The existing `test_cleans_up_tmp_on_failure` covers `_write_json_atomic` internals, not the two-file sequencing in `emit_info_and_lineage_diff`. Combined with the Pass D issue above, this is the highest-value missing test: mock a second-write failure and assert the final CLI summary reports a partial failure rather than `Cloud upload complete.`.

- Happy-path, missing-info_url, missing-lineage_diff_url, both-missing, local-mode non-regression, scratch-dir prefix + cleanup: all covered. 15/15 pass locally.
- Round-trip test through `LineageDiff.model_validate` is a strong contract test — good.

- Rebase onto A1 (#1334) merged cleanly: `per_node_scratch` and `metadata_scratch` coexist in one `try/finally` block with independent cleanup. Scratch prefixes (`recce-per-node-`, `recce-metadata-`) do not collide. Test `test_scratch_dir_has_distinct_prefix_and_is_cleaned_up` pins this.
- Control flow change is structurally sound: outer `if is_cloud:` wraps emit + upload under a single `try/finally` for cleanup, with `if cloud_client:` guarding the upload block and `if upload_urls is not None:` guarding the URL-dependent section. No code path leaks scratch dirs.
- `_UPLOAD_TIMEOUT = 600s` applied consistently to all four upload calls.

- No SQL execution, no N+1. Single `model_dump` call per payload (two payloads total). File writes are atomic via sibling `.tmp` + `replace`.
- No streaming needed at current artifact scale (KB-range JSON).

- All synchronous code paths. No asyncio, no threads.

### Verification Results
- `uv run pytest tests/test_info_emitter.py tests/test_cli_info_emitter.py` — 15 passed / 0 failed.
- `uv run ruff check` on changed files — clean.
- `uv run black --check --line-length 120` (repo's actual formatter) — clean.

## Verdict: NEEDS-CHANGES

### Issues
1. `recce/cli.py:709-713` + `cli.py:884-911` — Partial emit failure silently produces a false "Cloud upload complete." terminal message because the `None` local_path path neither uploads nor appends to `upload_failures`. Either append on emit-except, or add an explicit missing-file branch in the upload loop.

### Notes
1. `recce/util/info_emitter.py:92` — `default=str` in `json.dumps` silently stringifies unexpected types; harmless today but masks future regressions in manifest_metadata typing.
2. `tests/test_cli_info_emitter.py` — No test for partial-emit failure path (info.json written, lineage_diff.json write raises). Adding this would pin the Issue #1 fix.
3. The metadata_uploads loop at `cli.py:880-911` could be extracted to a small helper (`_upload_artifact(...)`) — the same PUT-and-handle-response pattern repeats three times in this function and diverges only in display name, URL key, and content type. Non-blocking; just easier to maintain as Phase III adds more artifacts.

## What I could not verify
- Actual behavior against a real Cloud staging server with the sibling PRs (#1243, #1244) deployed — this review is static only.
- End-to-end S3 PUT semantics (e.g., whether the Cloud signed URL enforces `Content-Type: application/json` server-side or accepts any).

- Secrets or PII in emitted JSON: none (payload is derived from manifest/catalog structure, not query results or auth data).
- Mock drift: all mocks match real signatures (`get_upload_urls_by_session_id`, `get_lineage_diff`, `build_full_cll_map`, `adapter.type()`).
- Resource leaks: file handles use `with open(...)` in all four upload paths; scratch dirs cleaned up in `finally`.
- Cross-rebase breakage with A1's per_node.db flow: coexistence is correct; per_node emit is gated on `per_node_db_url` presence and uses a distinct scratch dir.
- Off-by-one / boundary issues in the emit-vs-upload gate: none — control flow is linear and correctly guarded.
- Wire-format drift between CLI emit and server `/api/info`: identical `model_dump(exclude_none=True, by_alias=True)` path; contract is pinned by reuse of `build_merged_lineage`.

even-wei (commented):
Claude Code Review: 1 ISSUE + 3 NOTEs (NEEDS-CHANGES). Partial-emit failure silently reports 'Cloud upload complete.' — see review comment for details.

gcko:
## Updated Review — PR #1336

### Summary
Re-review after `a24c1cf6 fix(cli): drop misleading scratch path from emit success messages`. The first nit from the prior review (misleading "Metadata saved to {scratch_dir}" path) is fixed — both the metadata and per_node.db success messages now print `…emitted (size, elapsed)` without referencing the temp dir that's about to be deleted in `finally`. Verification re-run on `a24c1cf6`: `make flake8` clean; `pytest tests/test_info_emitter.py tests/test_cli_info_emitter.py` → 16 passed.

### Findings

#### [Resolved] Misleading "Metadata saved to …" path is deleted on the way out
**File:** `recce/cli.py:705`
**Status:** Fixed in `a24c1cf6`. Message now reads `Metadata emitted (info.json X.X KB, lineage_diff.json X.X KB, X.Xs)`. Same fix applied to the `per_node.db` emit message at `recce/cli.py:777`.

#### [Unaddressed — low priority] Two `tempfile.mkdtemp` calls when sometimes neither is used
**File:** `recce/cli.py:690-691`
**Issue:** Still allocates `per_node_scratch` and `metadata_scratch` unconditionally before checking which upload URLs are present. Wasted I/O on the rollout edge where Cloud hasn't shipped the new URL keys yet. Carried over from the prior review — restating for completeness, not blocking.
**Suggestion:** Defer `mkdtemp` until after `get_upload_urls_by_session_id` returns. Low priority — emit cost is negligible vs. the surrounding upload work.

### Verdict
Approved. The actionable nit is resolved, the remaining nit is genuinely low-priority and can be deferred or skipped. Ship it.

gcko (approved):

even-wei:
@gcko Pushed a24c1cf addressing review nits.

**Nit 1 (misleading "Metadata saved to {metadata_scratch}")** — fixed. Dropped the path from the success message; kept the load-bearing info (file sizes + elapsed). Applied the same fix to the `per_node.db` print one block down — same pattern (`per_node_scratch` is also cleaned up unconditionally in `finally`), so left to itself it would have read inconsistently against the new metadata message.

**Nit 2 (eager `tempfile.mkdtemp` for `metadata_scratch` / `per_node_scratch` before knowing which artifacts get uploaded)** — agreed in principle, but holding off here. The cost is two empty-dir mkdtemp calls plus, in the rollout-edge case, two small JSON writes that don't get uploaded — well under any user-perceptible threshold. Restructuring the control flow to defer `mkdtemp` until after `get_upload_urls_by_session_id` returns is a non-trivial refactor that would also reshuffle the surrounding `try/finally` cleanup logic. Better as a follow-up once Phase III lands more artifacts and the pattern repeats.

Verification: 16/16 tests pass (`test_info_emitter.py` + `test_cli_info_emitter.py`), black/flake8 clean.

gcko (approved):
Claude Code Review: Updated — prior nit fixed in a24c1cf6, no critical issues.

PR #1337, opened 2026-04-24
test(cloud): add staging recce-cloud CLI smoke test

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

test / chore — new CI smoke test for the standalone `recce-cloud` CLI.

Adds a staging-only smoke test for the standalone `recce-cloud` CLI covering the `RECCE_API_TOKEN` workflow end to end, including the `--session-base` (isolated base) path. Closes two gaps: the legacy `smoke-test-cloud` job only exercises the older `recce cloud …` subcommands, and `--session-base` has no coverage today.

Runs on the existing nightly cron via `schedule` when `environment=staging`. Prod variant and GHA auto-detect path will be follow-up PRs once staging proves stable.

N/A — infra improvement.

Two prerequisites that are non-obvious:

- **Token role:** `RECCE_API_TOKEN_STAGING` must have **MANAGER** role, not just MEMBER. `recce-cloud delete` (the teardown step) requires MANAGER+ on the backend — without it the smoke leaks a session per run. Audited the token's blast radius; it's limited to the dedicated test project and the script only deletes sessions it creates itself.
- **Seeded base session:** the staging test project must have a session promoted to base in the UI (one-time, already done). Step 11's prod-session download reads `list --type prod` and fails if empty.

Slack notification on failure was intentionally deferred until we've seen a few clean nightly runs.

None


even-wei:
# Code Review: PR #1337

**Files reviewed:** 2
**Categories:** Config/infra (workflow YAML), Tests (bash smoke test)
**Passes run:** A, B, C, D, E, F

## Validation Results

Traced every command and branch in `integration_tests/dbt/smoke_test_recce_cloud.sh`:

- `set -euo pipefail` at top; strict-mode errors propagate cleanly.
- Cleanup trap (lines 41–49) captures `$?` **before** doing any work, then exits with the original code — correct pattern. `|| true` on the delete call prevents teardown failures from masking the real error.
- `SESSION_ID=""` on line 180 cleanly prevents the EXIT trap from double-deleting on the success path.
- Negative check (lines 133–144) correctly toggles `set +e`/`set -e` around the call that is expected to fail, capturing the exit code before re-enabling strict mode.
- `jq -e` assertions on `doctor.pre.json`, `has_isolated_base`, and the `all(.[]; .id != $i)` post-delete check all have specific, fail-loud semantics.

### Pass B: Security — PASS

- `actions/checkout@v4` uses `persist-credentials: false` — good.
- `RECCE_API_TOKEN_STAGING` never touches stdout; not echoed, not logged.
- Staging job only triggers via `schedule` or `workflow_dispatch` (the latter requires write perms upstream) — so the lack of an `authorize` gate on staging is consistent with the existing `smoke-test-cloud-staging` job.
- PR description documents the MANAGER-role requirement and the blast-radius audit on the token. Operationally responsible.

Every `recce-cloud` command and flag used in the script was verified against `recce_cloud/recce_cloud/cli.py`:

| Script usage | CLI verification |
|---|---|
| `version` | cli.py:70 |
| `init --org --project --status` | cli.py:180-203 |
| `doctor --json` | `commands/diagnostics.py:147-154` |
| `upload --session-name --yes --target-path --session-base --session-id --type --dry-run` | cli.py:487-533 |
| `list --type --json` | cli.py:956 |
| `download --session-id --target-path --force --dry-run` | cli.py:1122-1151 |
| `delete --session-id --force --dry-run` | cli.py:1336-1358 |

All match.

- Every script-created session is tied to a `trap cleanup EXIT` that deletes it even on failure mid-script.
- Empty-`SESSION_ID` guards on lines 113 and 158 fail loudly with actionable messages.
- The `|| true` on line 45 ensures teardown failures don't clobber the original exit code.
- Edge: if the `upload` on line 111 **partially succeeds** server-side but fails client-side before the `list` on line 112 assigns `SESSION_ID`, the session would leak. In practice `recce-cloud upload` returns non-zero only after the session is fully created or not at all, so this is a theoretical concern, not an observed one.

- Assertions are specific (`test -s` for size>0, `jq -e` for structural invariants, explicit exit-code check on the negative case) — no vacuous "did not throw" tests.
- The negative test (`--session-base + --type prod` must be rejected) is a real contract lock-in — future refactors that loosen validation will fail here.

- New file and new CI job, no existing code paths modified.
- PR description accurately describes scope ("staging-only", "prod variant is follow-up").
- The new `smoke-test-recce-cloud-staging` job is intentionally **not** wired into `notify-staging-failure` (which only `needs: smoke-test-cloud-staging`), per the PR's "Slack notification deferred" note.

### Verification Results

- `bash -n` syntax check: **PASS**
- shellcheck: not available locally, not run

## Verdict: GO

No blockers, no issues. The script has been engineered with the right primitives (strict mode, cleanup trap, exit-code-preserving teardown, and specific structural assertions). Every CLI flag used matches the current implementation.

### Notes

1. **`~/.recce/profile.yml` overwrite on local runs** (`smoke_test_recce_cloud.sh:37-39`). If a developer runs this locally, their own Recce profile is silently overwritten. The existing `smoke_test_cloud.sh` has the same pattern, so this isn't new — but a short "intended for CI" comment near the top of the script would help future readers.
2. **Asymmetric artifact checks.** Line 151-152 checks both `manifest.json` and `catalog.json` for the downloaded PR session; line 164 checks only `manifest.json` for the prod session. Either intentional (prod sessions may legitimately lack `catalog.json`) or an oversight. If intentional, a one-line comment would help; if not, add the missing `test -s target-prod/catalog.json`.
3. **Append-without-newline risk** (line 81). `echo "where customer_id > 0" >> models/customers.sql` assumes the upstream `chore/smoke-test-cloud` branch's `customers.sql` ends in a newline. If it doesn't, the append concatenates onto the final line, producing invalid SQL. Very low probability given git's usual behavior, but a defensive `printf '\n' >> models/customers.sql` before the append would eliminate the fragility.
4. **Minor style — unquoted constants** (line 62): `git clone --depth 1 --branch $GIT_BRANCH $GIT_REPO`. Constants, so no injection vector, but consistency with the rest of the script would quote them.
5. **Fixture setup duplication** with `smoke_test_cloud.sh` (profile.yml bootstrap, jaffle_shop clone). Minor — since the two scripts test different CLIs, keeping them independent is reasonable. Worth revisiting once the prod variant lands and there are three copies.
6. **Staging job auth gating.** The new `smoke-test-recce-cloud-staging` job follows the existing `smoke-test-cloud-staging` pattern — no explicit `authorize` check, relying on the fact that `environment == 'staging'` can only come from `schedule` or `workflow_dispatch` (write-perm-gated by GitHub). If a fourth trigger type is ever added to `resolve-environment` that routes to staging, this assumption needs to be re-examined. Not a current issue.

even-wei (approved):
Claude Code Review: No critical issues found. See comment for detailed analysis and minor observations.

Issue #26, opened 2023-11-27
[RC-19] [Feature] Support dbt command options `--target`, `--profile`, `--profile-dirs`


popcornylu:
### Describe the feature

Currently, we use the default target and profile in project file to find the credentials
https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles

This feature request is to add option in the recce command to specify the target/profile to use in recce

```
recce server --target mytarget --profile myprofile --profiles-dir /path/to/profile
```

<sub>[RC-19](https://linear.app/piperider/issue/RC-19/[feature]-support-dbt-command-options-target-profile-profile-dirs)</sub>

PR #1342, opened 2026-04-24
fix(mcp): DRC-3307 fix submit_run race condition + run_check auto-approve

iamcxa:
## Problem

PR comment Checklist showed wrong status for two reasons:

1. **`submit_run` race condition**: `update_run_result` was `async` and scheduled via `run_coroutine_threadsafe`. After `await future`, `run.status` was still `RUNNING` — auto-approve gate (`run.status == FINISHED`) never triggered.
2. **`run_check` missing auto-approve**: `_tool_run_check` had no auto-approve logic. Preset checks (created via API before the agent runs) stayed ⏳ Unapproved forever.

Note: PR #1324 added auto-approve to `_tool_create_check` but it was defeated by the race condition — `run_executed = run.status == RunStatus.FINISHED` evaluated to `False` because `run.status` was still `RUNNING`.

### Fix 1: submit_run race condition (run_func.py)

`update_run_result` changed from `async def` + `run_coroutine_threadsafe` to plain `def` called synchronously inside the executor thread `fn()`. Now `run.status` and `run.result` are set BEFORE the future resolves. This fixes ALL 6 callers of `submit_run`:

| Caller | File | Benefit |
|--------|------|---------|
| REST API (run) | `run_api.py` | `run.result` populated |
| REST API (check rerun) | `check_api.py` | `run.result` populated |
| MCP run_check | `mcp_server.py` | Auto-approve works |
| MCP create_check | `mcp_server.py` | Auto-approve works (fixes #1324) |
| Preset checks | `run.py` | `run_should_be_approved()` sees result |
| State checks | `run.py` | Run completes before next iteration |

### Fix 2: run_check auto-approve (mcp_server.py)

Added auto-approve to `_tool_run_check` matching `_tool_create_check` policy:
- Metadata checks (lineage_diff, schema_diff): unconditional auto-approve after success
- Task checks: conditional on `run.status == FINISHED and not run.error`

### Tests

- Removed `asyncio.sleep(0.1)` workarounds (no longer needed with sync update)
- Added `assert run.status == RunStatus.FINISHED` after `await future`
- Added `update_check_by_id(is_checked=True)` assertions to all 3 run_check tests (row_count_diff, lineage_diff, schema_diff)

## Testing

- 143 tests pass (29 run_func + 114 mcp_server)
- Break-point probe verified at Level B (unit tests exercise exact break-point)

## Staging Verification

Local integration test (Docker image with this fix) on jaffle_shop_golden PR #26:

| Metric | Before | After |
|--------|--------|-------|
| ✅ Approved | 1/10 (10%) | **8/10 (80%)** |
| ⏳ Unapproved | 8/10 | **0/10** |

## Related

- Companion PR: DataRecce/recce-cloud-infra#1245 (consumer-side defense-in-depth)
- Supersedes: #1324 (auto-approve was defeated by race condition)
- DRC-3307


recce/mcp_server.py
```diff
@@ -1704,15 +1728,26 @@ async def _tool_run_check(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
                     result=result,
                     triggered_by=triggered_by,
                 )
+                # Auto-approve metadata checks (always succeed if no exception)
+                check_dao.update_check_by_id(check_id, PatchCheckIn(is_checked=True))
```
iamcxa: **HIGH (cloud exception leaks past MCP error contract).** This `update_check_by_id` call sits inside the metadata branch's `try / except RecceException` (covering line 1719-1735). In cloud mode, `CheckDAO.update_check_by_id` re-raises `RecceCloudException`, which is defined at `recce/util/recce_cloud.py:31` as `class RecceCloudException(Exception)` — **not** a subclass of `RecceException`. A cloud-side failure during auto-approve therefore propagates raw to the MCP caller instead of being wrapped in `ValueError`, breaking the consistent error contract this branch was designed to provide.

The same applies at line 1750 (task branch) — outside an explicit `try`, but still in a function that wraps `RecceException` into `ValueError` further down (line 1753).

Fix options:
1. Move `update_check_by_id` outside the `try` (mirrors `_tool_create_check:1832-1833`, which structures auto-approve as a separate, unguarded call).
2. Broaden the catch: `except (RecceException, RecceCloudException) as e: raise ValueError(str(e)) from e`.

recce/apis/run_func.py
```diff
@@ -191,10 +204,10 @@ def fn():
                     logger.warning(f"Failed to serialize task.params: {e}")
                     updated_params = None
 
-            asyncio.run_coroutine_threadsafe(update_run_result(run, result, None, updated_params), loop)
+            update_run_result(run, result, None, updated_params)
```
iamcxa, line 242: **HIGH (new cross-thread race window introduced by the fix).** This direct call (and its sibling on the failure path 1 line below) moves the `update_run_result` execution from event-loop-thread (via `run_coroutine_threadsafe`) into the executor thread. The PR fixes one race (`status` stays RUNNING after `await future`) but introduces a smaller one inside `update_run_result` body:

- Success path stores `run.result` BEFORE `run.status = RunStatus.FINISHED`.
- Failure path stores `run.error` BEFORE `run.status = FAILED`.

CPython's GIL makes each individual store atomic, but the multi-store sequence is no longer serialized with async-loop readers. Observable inconsistencies:

- `wait_run_handler` (`run_api.py:85`) terminates its wait loop on `run.result` non-None and returns the run object. A reader interleaving between the two stores sees `result != None` with `status == RUNNING`.
- `cancel_run` (`run_func.py:225-235`) writes `run.status = CANCELLED` from the loop while the executor thread may be mid-`update_run_result`. The success path unconditionally overwrites with FINISHED, regressing the cancel sentinel. (Pre-PR also had this on success but was at least serialized on the loop, making it deterministic in tests.)

This is exactly the surface the consumer-side `recce-cloud-infra#1245` `_auto_approve_successful_checks` is defending against. Worth either:
1. Inverting store order in `update_run_result` (set `status` first, then `result`/`error`) so readers never see "result-present + status-RUNNING".
2. Skipping the FINISHED assignment when `status == CANCELLED` to preserve the cancel sentinel.
3. At minimum, document the CPython GIL assumption explicitly in the docstring.

recce/mcp_server.py
```diff
@@ -1704,15 +1728,26 @@ async def _tool_run_check(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
...
+                check_id=check_id,
+                triggered_by=triggered_by,
             )
             run.result = await future
+            # Auto-approve on successful run — same policy as _tool_create_check.
+            # PM decision: Passed = Approved (a check that ran successfully is
+            # considered reviewed by the agent). Without this, preset checks
+            # (created via API before the agent runs) stay ⏳ Unapproved forever.
+            if run.status == RunStatus.FINISHED and not run.error:
+                check_dao.update_check_by_id(check_id, PatchCheckIn(is_checked=True))
```
iamcxa: **MEDIUM (persistence asymmetry vs `_tool_create_check`).** `_tool_create_check` calls `await asyncio.get_event_loop().run_in_executor(None, export_persistent_state)` after auto-approve (line 1836-1837). `_tool_run_check` does NOT — neither here nor in the metadata branch (line 1732). The PR claims "matches `_tool_create_check` policy" but persistence behavior differs.

Impact:
- **Cloud mode**: `update_check_by_id` writes to the cloud API directly, so persistence isn't lost.
- **Local mode**: `is_checked=True` is updated only in the in-memory `CheckDAO` list. If the process exits between this auto-approve and the MCP-shutdown export (e.g., SIGTERM, OOM, crash), the approval is lost from disk. The REST `PATCH /checks/{check_id}` endpoint correctly registers `export_persistent_state` as a background task (`check_api.py:246`); the MCP path skips this.

Fix: add `await asyncio.get_event_loop().run_in_executor(None, export_persistent_state)` after both auto-approve sites (here + line 1732). The `export_persistent_state` symbol is already imported in `_tool_create_check`'s lazy-import block.

recce/mcp_server.py
```diff
@@ -1704,15 +1728,26 @@ async def _tool_run_check(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
...
         try:
             run, future = submit_run(
-                check.type, params=check.params or {}, check_id=check_id, triggered_by=triggered_by
+                check.type,
+                params=check.params or {},
+                check_id=check_id,
+                triggered_by=triggered_by,
             )
             run.result = await future
+            # Auto-approve on successful run — same policy as _tool_create_check.
```
iamcxa: **MEDIUM (claim is misleading).** "same policy as `_tool_create_check`" is wrong on persistence — only the *gate* is equivalent, not the operational policy. See finding #3.

Suggest:
```python
# Auto-approve on successful run — same auto-approve gate as _tool_create_check
# (line 1832: run_executed and not run_error). PM decision (DRC-3307): Passed = Approved.
# Note: unlike _tool_create_check, this does NOT call export_persistent_state — state
# is exported on MCP shutdown via the state_loader.export() call in run().
```
This way the divergence is explicit and a future reader won't add `export_persistent_state` here under a false equivalence assumption.

recce/mcp_server.py
```diff
@@ -1704,15 +1728,26 @@ async def _tool_run_check(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
                     result=result,
                     triggered_by=triggered_by,
                 )
+                # Auto-approve metadata checks (always succeed if no exception)
```
iamcxa: **MEDIUM (unconditional auto-approve on metadata branch — empty result still approves).** The comment "always succeed if no exception" describes the *gate* but not what "success" means. If `_tool_lineage_diff` or `_tool_schema_diff` returns an empty/degenerate result (e.g., `{nodes: [], edges: []}`) without raising — a valid path when no nodes changed — the check is auto-approved despite producing zero evidence.

The task branch (line 1749) gates explicitly on `run.status == FINISHED`. Metadata branch has no equivalent content guard.

Two paths:
1. Mirror task-branch intent and require a non-empty result:
   ```python
   if result:  # non-empty diff
       check_dao.update_check_by_id(check_id, PatchCheckIn(is_checked=True))
   ```
2. Document that empty result IS valid evidence ("zero changes confirms logic-only modification") — in which case the comment should say so explicitly so the policy is unambiguous.

recce/mcp_server.py
iamcxa: **MEDIUM (cloud audit attribution gap).** `CheckDAO.update_check_by_id` in cloud mode reads `get_current_cloud_user()` to populate `acting_user_id` for audit attribution at the cloud API layer. `get_current_cloud_user()` reads from a `ContextVar` set only in:
- HTTP middleware `extract_cloud_user_context` (per HTTP request)
- WebSocket message handler `_handle_cloud_user_context` (per WS frame)

The MCP server does not call `set_current_cloud_user()`. When `_tool_run_check` (and `_tool_create_check`) fires `update_check_by_id`, the ContextVar returns `None`, and the cloud API records every MCP-triggered auto-approval as anonymous.

In shared-cloud deployments where multiple users share one recce process, the approval audit trail can't identify which user's MCP session triggered each auto-approve.

Mitigations:
1. Extract user identity from MCP transport (e.g., SSE auth headers) and call `set_current_cloud_user()` before `update_check_by_id`.
2. At minimum, log the `triggered_by` field at the auto-approve site so server logs preserve some attribution: `logger.info(f"Auto-approved check {check_id} (triggered_by={triggered_by})")`.

This is a pre-existing pattern in `_tool_create_check` too — not introduced by this PR — but worth flagging since this PR widens the surface (run_check now also auto-approves).

iamcxa (commented):
@iamcxa — review of DRC-3307 root-cause fix.

Overall: race-condition fix is correct in spirit and the 80% jaffle_shop_golden integration result is solid C-level evidence. Surfacing 12 inline findings; 2 HIGH worth resolving before merge (cloud exception leak + new cross-thread race window introduced by the sync fix), and a persistence-asymmetry MEDIUM that breaks the "matches `_tool_create_check` policy" claim. Defaulting to **COMMENT** because the core fix is sound but several edges need attention.

## Verification Matrix (from PR body)

| # | Concern | Status |
|---|---------|--------|
| V1 | `submit_run` race fixed — sync `update_run_result` | Confirmed at unit + integration. **Introduces new smaller race**: result-before-status store ordering visible to async-loop readers (#2) |
| V2 | All 6 `submit_run` callers benefit | Confirmed; readers may now see transient `RUNNING + result-present` inconsistency (#2) |
| V3 | `_tool_run_check` auto-approve matches `_tool_create_check` policy | Gate matches; **persistence behavior differs** — `_tool_create_check` calls `export_persistent_state`, `_tool_run_check` does not (#3, #4) |
| V4 | Removed `asyncio.sleep(0.1)` workarounds | Confirmed; only 1 of 3 removed-sleep tests added a regression assertion (advisory B) |

## Break-point Coverage

```
BP1 update_run_result sync (run_func.py:152-178):    B (regression test) + C (jaffle_shop_golden 80%)
BP2 _tool_run_check metadata auto-approve:           B (2 success tests) — no negative test (#11)
BP3 _tool_run_check task auto-approve:               B (1 success test) — no negative test for error path
```

**Residual uncertainty:**
- Cross-thread race window (NEW): result-before-status store ordering creates sub-µs window where async-loop readers see `result is not None` while `status == RUNNING`. Symptom likely undetectable in practice but a regression of the previous loop-serialized model. Fix: invert the store order, OR document GIL/CPython assumption (#2).
- `RecceCloudException` not subclass of `RecceException` — cloud-mode `update_check_by_id` failure surfaces unwrapped to MCP caller (#1).
- Persistence asymmetry: local-mode SIGTERM after `run_check` before any other operation → auto-approve lost (#3).
- Cloud-mode `acting_user_id=None` for all MCP-triggered auto-approvals — shared-instance audit attribution gap (#6).

**Recommended human probe before merge:**
1. Decide store-order inversion in `update_run_result` (status before result/error) OR add explicit GIL/CPython note in docstring.
2. Move `update_check_by_id` outside the metadata `try` (matches `_tool_create_check`'s structure) OR broaden `except` to include `RecceCloudException`.
3. Decide whether to add `export_persistent_state` to `_tool_run_check` for parity with `_tool_create_check`, OR clarify the comment that persistence is intentionally deferred to MCP shutdown.
4. Add `mock_check_dao.update_check_by_id.assert_not_called()` to the existing RecceException-path test in `test_mcp_server.py` (cheap regression guard, #11).

## Advisory (not posted as inline comments)

| # | File:Line | Root | Summary | Suggested Action |
|---|-----------|------|---------|------------------|
| A | run_func.py:155-163 | DOC | DRC-3307 historical narrative is **load-bearing** while consumer recce-cloud-infra#1245 defense exists — its `derive_check_run_status` workaround cites this docstring as the canonical justification | Add sunset note: "Safe to remove once recce-cloud-infra defensive guards in `derive_check_run_status` are removed" (#9) |
| B | tests/apis/test_run_func.py:376, 395 | DOC | 2 of 3 removed-sleep tests don't add regression assertion (`run.status == FINISHED`); only `test_normalized_params_propagate_to_run` does | Decide: add regression assertion to all 3, OR document why one canonical assertion is sufficient |
| C | — | NEW | Cross-repo defense-in-depth pattern: root-cause repo's docstring outlives its own fix because consumer-side defense cites it as canonical justification | Captured in `recce-cloud-infra/.claude/review-lessons.md` (just created in companion review) |

## Verification Summary

| Check | Result |
|-------|--------|
| ruff (run_func.py + mcp_server.py) | clean |
| Pre-scan persistence asymmetry | 1 finding (export_persistent_state missing in _tool_run_check) |
| code-reviewer | 2 HIGH + 2 LOW |
| comment-analyzer | 1 critical + 5 improvements + 4 rot risks |
| ToB security (standard) | 2 MEDIUM + 2 LOW |
| Unit tests | not executed locally — PR claims 143 pass |

recce/mcp_server.py
```diff
@@ -1704,15 +1728,26 @@ async def _tool_run_check(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
...
+                check.type,
+                params=check.params or {},
+                check_id=check_id,
+                triggered_by=triggered_by,
             )
             run.result = await future
+            # Auto-approve on successful run — same policy as _tool_create_check.
+            # PM decision: Passed = Approved (a check that ran successfully is
+            # considered reviewed by the agent). Without this, preset checks
+            # (created via API before the agent runs) stay ⏳ Unapproved forever.
```
iamcxa: **LOW (consumer-side context bleed).** "preset checks (created via API before the agent runs) stay ⏳ Unapproved forever" — the "preset checks" concept and the ⏳ icon both originate from `recce-cloud-infra` (the agent + PR comment system). The OSS `recce` repo doesn't model "preset checks" — checks are just checks here. This comment leaks consumer-side context into a library-side file and will go stale silently if `recce-cloud-infra` renames the concept or changes the icon.

Suggest either:
```python
# Without this, checks created externally (not via the agent) never get marked
# approved even after a successful run.
```
or mark explicitly as cross-repo motivation:
```python
# Cross-repo motivation (DataRecce/recce-cloud-infra): preset checks created via
# API before the agent runs would otherwise stay Unapproved forever.
```

recce/mcp_server.py
```diff
@@ -1704,15 +1728,26 @@ async def _tool_run_check(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
...
             run, future = submit_run(
-                check.type, params=check.params or {}, check_id=check_id, triggered_by=triggered_by
+                check.type,
+                params=check.params or {},
+                check_id=check_id,
+                triggered_by=triggered_by,
             )
             run.result = await future
+            # Auto-approve on successful run — same policy as _tool_create_check.
+            # PM decision: Passed = Approved (a check that ran successfully is
```
iamcxa: **LOW (PM decision attribution).** "PM decision: Passed = Approved" lacks ground-truth link. Future maintainers questioning whether this is intentional have no path to verify. Add `# See: DRC-3307` so the rationale is traceable when Linear archives change.

recce/apis/run_func.py
```diff
@@ -146,8 +152,15 @@ def progress_listener(message=None, percentage=None):
 
     task.progress_listener = progress_listener
 
-    async def update_run_result(run, result, error, updated_params=None):
-        """Update run with result, error, and optionally updated params."""
+    def update_run_result(run, result, error, updated_params=None):
+        """Update run with result, error, and optionally updated params.
```
iamcxa, line 156: **LOW (load-bearing historical narrative — add sunset condition).** This "Previously this was async + scheduled via run_coroutine_threadsafe..." docstring is genuinely useful right now: the consumer-side `recce-cloud-infra/api_server/apis/recce_task_func.py` (`derive_check_run_status` race-workaround) explicitly cites this exact behavior as the reason for its defense-in-depth, and #1245's review found three places downstream that reference this fix. So the narrative is canonical justification for code that lives in another repo.

But after 1-2 releases when DRC-3307 archives, this becomes archeology unless tied to a sunset condition. Suggest appending:

```
DRC-3307 historical context (safe to remove once recce-cloud-infra defensive
guards in derive_check_run_status are removed — those guards depend on this
rationale being preserved here as the canonical justification).
```

This way a future maintainer knows when it's safe to prune.

tests/test_mcp_server.py
```diff
@@ -576,6 +579,11 @@ async def test_tool_run_check_row_count_diff(self, mcp_server):
         assert "check_id" in result
         assert result["check_id"] == str(check_id)
 
+        # Verify auto-approve: successful run → is_checked=True (DRC-3307)
```
iamcxa: **LOW (negative-path test gap).** This success-path assertion confirms `update_check_by_id` is called when `run.status == FINISHED`. The metadata branch in `mcp_server.py:1731-1735` is wrapped in `try/except RecceException`. If `_tool_lineage_diff` raises a `RecceException`, control jumps to `raise ValueError(str(e))` and `update_check_by_id` should NOT be called.

There is no test asserting this. Cheap addition to the existing `test_run_check_metadata_branch_recce_exception` (or equivalent) test: `mock_check_dao.update_check_by_id.assert_not_called()`. Without it, a future regression that moves auto-approve before the result computation would pass tests.

iamcxa: **NIT (DRC-3307 ticket tag).** This tag (and the same on lines 630, 678) ambiguously could mean "introduced by DRC-3307" or "regression guard for DRC-3307". When DRC-3307 archives in Linear, the meaning fades. Suggest tightening to:

```python
# Regression: successful run must auto-approve (DRC-3307 root cause)
```

or consolidating to a single place — the test name + assertion already convey intent.

tests/apis/test_run_func.py
```diff
@@ -341,16 +341,20 @@ async def test_normalized_params_propagate_to_run(self, mock_context, mock_task_
...
 
-            # The update_run_result coroutine is scheduled via
-            # asyncio.run_coroutine_threadsafe from the executor thread.
-            # Its callback chain (call_soon_threadsafe → create Task →
-            # execute coroutine) spans multiple event loop iterations,
-            # so a brief sleep lets the loop fully process it.
-            await asyncio.sleep(0.1)
+            # update_run_result is now called synchronously inside the
+            # executor thread (DRC-3307), so run state is fully updated
+            # by the time the future resolves — no sleep needed.
```
iamcxa: **NIT (forward-looking framing).** The current comment frames the change as "no sleep needed" — historically accurate but doesn't tell a future reader what BREAKS if someone changes the implementation. Suggest reframing as a regression assertion:

```python
# Regression guard for DRC-3307: run.status MUST be set before the future
# resolves. If this fails, update_run_result is being scheduled async again
# instead of called inline.
```

This stays useful even after the original race is forgotten.

recce/mcp_server.py
iamcxa: Resolved by 8385e0a7 — went with option 1 (move outside try, mirrors `_tool_create_check`). The metadata branch's `try/except RecceException` now only wraps the result-computation step; `update_check_by_id` runs unguarded so a `RecceCloudException` propagates raw to the caller instead of being silently absorbed. Same restructure on the task branch — auto-approve is now a single shared block after both branches join.

recce/apis/run_func.py
iamcxa: Resolved by 8385e0a7 — applied both mitigations:
1. Inverted store order in `update_run_result`: `run.status` is now written BEFORE `run.result`/`run.error`, so async readers never see "result-present + status-RUNNING".
2. Both success and failure paths now skip the status overwrite when `run.status == CANCELLED`, preserving the cancel sentinel set by `cancel_run` from the loop thread.
3. Docstring updated to declare the cross-thread store-ordering invariant explicitly (CPython GIL atomic per attribute, but multi-store sequences are not serialized with the loop).

New test `test_cancel_sentinel_preserved_on_success` exercises the newly-guarded success-path branch (also closes a coverage gap).

recce/mcp_server.py
iamcxa: Resolved by 8385e0a7 — added `await loop.run_in_executor(None, export_persistent_state)` after the auto-approve. Now matches `_tool_create_check` policy on persistence as well as on the gate, so local-mode `is_checked` survives process exit. The `export_persistent_state` import was hoisted from `_tool_create_check`'s lazy-import block to `_tool_run_check`'s top-of-function imports.

recce/mcp_server.py
iamcxa: Resolved by 8385e0a7 via the structural fix (#3). Persistence now actually matches `_tool_create_check`, so the comment is no longer misleading — the policies are aligned in code rather than just in claim.

recce/mcp_server.py
iamcxa: Resolved by 8385e0a7 via path 2 — documented explicitly that empty result IS valid evidence for metadata-only types (zero changes confirms the upstream PR did not affect lineage/schema). The auto-approve comment block now states this directly so the policy is unambiguous.

recce/mcp_server.py
iamcxa: Partially addressed in 8385e0a7 — added `logger.info(f"Auto-approved check {check_id} (triggered_by=...)")` so server logs preserve attribution at the auto-approve site. The full ContextVar-injection mitigation (extract user from MCP transport, call `set_current_cloud_user()`) is intentionally deferred — it's pre-existing scope shared with `_tool_create_check` and warrants a separate ticket rather than being bundled into the DRC-3307 race-condition fix.

recce/mcp_server.py
iamcxa: Resolved by 8385e0a7 — comment block reframed to drop the consumer-side concepts ("preset checks" / ⏳ icon). Now reads as a library-side rationale: "Auto-approve on successful run — same gate as `_tool_create_check`. PM decision: Passed = Approved (See: DRC-3307). For metadata-only types, an empty result IS valid evidence: zero changes confirms the upstream PR did not affect lineage/schema."

recce/mcp_server.py
iamcxa: Resolved by 8385e0a7 — added `(See: DRC-3307)` inline with the PM-decision line so the rationale is traceable when Linear archives change.

recce/apis/run_func.py
iamcxa: Resolved by 8385e0a7 — appended a sunset condition tied to the consumer-side defensive guards: "safe to remove (and this docstring archeology along with it) once the recce-cloud-infra defensive guards in `derive_check_run_status` are removed — those guards depend on this rationale being preserved here as the canonical justification." Future maintainer now knows the prune signal.

tests/test_mcp_server.py
iamcxa: Resolved by 8385e0a7 — added `test_tool_run_check_metadata_branch_recce_exception_no_auto_approve` which forces `_tool_lineage_diff` to raise `RecceException`, asserts `ValueError` is raised, AND asserts both `update_check_by_id.assert_not_called()` and `mock_export.assert_not_called()`. A future regression that moves auto-approve before the result computation will now fail this test. Also closes a codecov-flagged coverage gap on the metadata-branch error path.

iamcxa: Resolved by 8385e0a7 — tightened the three tags to "Regression: ... must auto-approve (DRC-3307 root cause)" so the meaning survives Linear-archive of the ticket. The test name + assertion convey intent on their own; the comment now adds the *failure-mode* context that an unfamiliar reader needs.

tests/apis/test_run_func.py
iamcxa: Resolved by 8385e0a7 — reframed as a forward-looking regression assertion: "Regression guard for DRC-3307 root cause: run.status MUST be set BEFORE the future resolves. If this assertion needs a sleep/poll to pass, update_run_result is being scheduled async again (e.g., via run_coroutine_threadsafe) instead of called synchronously inside the executor thread." Tells a future reader exactly what implementation change would break this.

iamcxa:
> Surfacing 12 inline findings; 2 HIGH worth resolving before merge (cloud exception leak + new cross-thread race window introduced by the sync fix), and a persistence-asymmetry MEDIUM that breaks the "matches `_tool_create_check` policy" claim.

All 12 inline threads addressed in commit `8385e0a7` (pushed). Per-thread replies posted. Summary by severity:

**HIGH (2) — both resolved**
- Cloud exception leak: `update_check_by_id` moved OUTSIDE the metadata-branch `try/except RecceException` (mirrors `_tool_create_check` structure), so `RecceCloudException` no longer escapes unwrapped.
- Cross-thread race window: `update_run_result` now writes `run.status` BEFORE `run.result`/`run.error` (no "result-present + status-RUNNING"), and both success and failure paths skip the status overwrite when `run.status == CANCELLED` (preserves the cancel sentinel from `cancel_run`). Docstring declares the cross-thread store-ordering invariant explicitly.

**MEDIUM (4) — all resolved**
- Persistence asymmetry: added `await loop.run_in_executor(None, export_persistent_state)` after auto-approve. Now genuinely matches `_tool_create_check` policy.
- Misleading "same policy" claim: closed by the structural fix above.
- Empty-result auto-approve: documented explicitly that empty result IS valid evidence for metadata-only types (zero changes confirms the upstream PR did not affect lineage/schema).
- Cloud audit attribution: added `logger.info(... triggered_by=...)` at the auto-approve site for log-side attribution. Full ContextVar injection deferred — pre-existing scope shared with `_tool_create_check`, warrants a separate ticket.

**LOW + NIT (5+1) — all resolved**
- Comment block reframed to drop consumer-side concepts ("preset checks" / ⏳).
- `(See: DRC-3307)` added inline with PM-decision attribution.
- Sunset condition added to historical-narrative docstring (tied to recce-cloud-infra defensive-guard removal).
- 3 DRC-3307 test tags tightened to "Regression: ... (DRC-3307 root cause)".
- "no sleep needed" reframed as forward-looking regression assertion.

**Coverage**: codecov-flagged 4 missing patch lines closed:
- `mcp_server.py` (2 lines): new `test_tool_run_check_metadata_branch_recce_exception_no_auto_approve` exercises the negative path.
- `run_func.py` (2 lines): new `test_cancel_sentinel_preserved_on_success` (CANCELLED-collision branch) + `test_repr_fallback_when_str_error_is_none_string` (`repr(error)` fallback when `str(e) == "None"`).

Tests: 146 passing (was 115 + 31 unrelated suite additions during this fix). Manual review handoff per your request — no re-review tagging.

recce/apis/run_func.py
```diff
@@ -146,21 +152,54 @@ def progress_listener(message=None, percentage=None):
...
-        if result is not None:
-            run.result = result
-            run.status = RunStatus.FINISHED
         if updated_params is not None:
             # Merge updated params (preserves any fields not in updated_params)
             run.params.update(updated_params)
+        if result is not None:
+            # Status BEFORE result: see "Cross-thread store ordering" above.
+            if run.status != RunStatus.CANCELLED:
+                run.status = RunStatus.FINISHED
```
iamcxa, line 204: **LOW (residual GIL race not covered by test).** The cancel-sentinel guard is good, but it's a check-then-assign sequence — not atomic. Race scenario:

1. Executor thread: `if run.status != RunStatus.CANCELLED` — passes (status is RUNNING).
2. GIL releases; loop thread runs `cancel_run`, sets `run.status = CANCELLED`.
3. Executor resumes: `run.status = RunStatus.FINISHED` — overwrites CANCELLED.

`test_cancel_sentinel_preserved_on_success` covers the cancel-BEFORE-executor-enters case (sets `run.status = CANCELLED` *then* `await asyncio.wrap_future(future)`), which exercises the check passing through. The check-then-assign GIL window is theoretically sub-µs and unlikely in practice, but a future maintainer reading the test may assume cancellation is *fully* race-safe.

Two paths:
1. Document in the docstring that the cancel guard is best-effort against the GIL window — "defends against macroscopic races (cancel before executor enters); a sub-µs window between check and assign remains, accepted as practically unobservable."
2. If full safety matters, wrap the `if` + `=` pair in a `threading.Lock` (or use `threading.Event` as the cancel sentinel).

recce/apis/run_func.py
```diff
@@ -146,21 +152,54 @@ def progress_listener(message=None, percentage=None):
...
+
+        Cross-thread store ordering: this runs in the executor thread while
+        async callers read run.status / run.result from the event-loop
+        thread. CPython's GIL makes each individual attribute store atomic,
+        but multi-store sequences are not serialized with the loop.
+        Mitigations applied here:
+
+        - Status is written BEFORE result/error so a reader who sees a
+          terminal status can trust that it is observing the final state
+          on the next read of result/error (no "result-present +
```
iamcxa: **NIT (claim is true for completion-signal readers, not snapshot readers).** "No `result-present + status-RUNNING` window" is correct under the new ordering, but technically a *snapshot* reader (one that grabs `status` and `result` in a single render — e.g., `list_run_handler` returning the run object as-is) can briefly observe `(status=FINISHED, result=None)` between line 195 and line 196. The wait-for-completion readers (`wait_run_handler`) are the ones the doc actually protects, and that's the load-bearing case.

Minor wording tighten:

```
Status is written BEFORE result/error so a completion-signal reader
(e.g., wait_run_handler polling on result) never observes the inverse
window ("result-present + status-RUNNING"). A snapshot reader may
briefly see (FINISHED, result=None) for the sub-µs gap between the two
stores, accepted as practically unobservable.
```

Doesn't change behavior; just makes the docstring's promise match what the code actually guarantees.

iamcxa (commented):
@iamcxa — re-review after `8385e0a7`.

**Excellent close-out — would APPROVE if you weren't the author** (GitHub blocks self-approve). All 12 v1 inline findings + 2 of 3 advisory items addressed; 10 fully fixed, 2 with documented partial trade-offs. Sign-off equivalent: ready for human reviewer + merge. The new tests (`test_cancel_sentinel_preserved_on_success`, `test_repr_fallback_when_str_error_is_none_string`, `test_tool_run_check_metadata_branch_recce_exception_no_auto_approve`) noticeably expand coverage beyond what v1 review asked for.

## v1 → v2 Closure

| # | v1 finding | Status |
|---|-----------|--------|
| 1 | RecceCloudException not caught (HIGH) | ✅ FIXED — auto-approve moved OUTSIDE the `try / except RecceException` blocks; explicit comment explains the trade-off |
| 2 | Cross-thread store ordering (HIGH) | ✅ FIXED — status BEFORE result/error; CANCELLED preserved; new regression test |
| 3 | `export_persistent_state` missing (MEDIUM) | ✅ FIXED — `await ... run_in_executor(None, export_persistent_state)` added |
| 4 | "Same policy" claim misleading (MEDIUM) | ✅ FIXED — rewritten as "same gate as `_tool_create_check` (line 1832)" + persistence parity stated |
| 5 | Metadata empty result auto-approve (MEDIUM) | ✅ FIXED via doc — "empty result IS valid evidence: zero changes confirms..." |
| 6 | `acting_user_id=None` audit gap (MEDIUM) | ⚠️ PARTIAL — `logger.info(triggered_by=...)` added (the minimum I suggested); cloud-mode user-context propagation not done (acceptable, pre-existing pattern) |
| 7 | "preset checks" consumer-side context (LOW) | ✅ FIXED — phrasing removed |
| 8 | PM decision ticket link (LOW) | ✅ FIXED — `(See: DRC-3307)` added |
| 9 | Load-bearing docstring sunset (LOW) | ✅ FIXED — explicit sunset-condition paragraph added |
| 10 | DRC-3307 tag fade in tests (NIT) | ⚠️ Partial — wording reframed as regression-assertion, tag retained (acceptable) |
| 11 | Negative-path test gap (LOW) | ✅ FIXED — `test_tool_run_check_metadata_branch_recce_exception_no_auto_approve` added |
| 12 | Forward-looking framing in test comment (NIT) | ✅ FIXED — reframed as regression guard |
| Adv A | Sunset for historical narrative | ✅ FIXED via #9 |
| Adv B | Regression assertion in 2/3 removed-sleep tests | ⚠️ Partial — replaced with NEW dedicated tests (cancel sentinel + repr fallback) instead; net coverage improved |
| Adv C | Cross-repo review-lessons | Captured in companion repo |

## 2 minor follow-ups

Neither is blocking. Both are about the new cross-thread race window's residual edges, posted as inline LOW + NIT.

## Verification Matrix (final)

| # | Concern | Status |
|---|---------|--------|
| V1 | `submit_run` race fixed | ✅ Confirmed; ordering refined; cancel sentinel preserved |
| V2 | All 6 callers benefit | ✅ Confirmed; ordering protects completion-signal readers |
| V3 | `_tool_run_check` matches `_tool_create_check` policy | ✅ Confirmed; gate + persistence + comment all aligned |
| V4 | Removed sleep workarounds | ✅ Confirmed + new regression tests added |

recce/apis/run_func.py
iamcxa: Resolved by e175c783 via path 1 — docstring now explicitly states the guard is check-then-assign (not atomic), defends against the macroscopic race (cancel before executor enters), and notes the sub-µs GIL window between check and overwrite as accepted/practically unobservable. Also references `threading.Lock` / `threading.Event` as the escape hatch for any future caller that needs full atomicity. Going with doc-only — adding a Lock for a sub-µs window that no caller currently exercises is over-engineering for the actual workload.

recce/apis/run_func.py
iamcxa: Resolved by e175c783 — adopted your wording almost verbatim. The completion-signal vs snapshot-reader distinction is now spelled out: the store-order guarantee protects `wait_run_handler`-style polling on `result is not None`, while snapshot readers may briefly see (FINISHED, result=None) and are expected to re-poll. Doc now matches the actual guarantee.

iamcxa:
> **Excellent close-out — would APPROVE if you weren't the author** (GitHub blocks self-approve). All 12 v1 inline findings + 2 of 3 advisory items addressed.

Two v2 LOW/NIT threads addressed in `e175c783` (docstring-only refinements):

- Cancel-sentinel guard now documents check-then-assign as best-effort against the macroscopic race; sub-µs GIL window between check and overwrite called out as practically unobservable, with `threading.Lock` / `threading.Event` noted as the escape hatch for callers that need full atomicity.
- Store-order claim tightened: the guarantee protects completion-signal readers (e.g., `wait_run_handler` polling `result is not None`); snapshot readers may briefly see `(FINISHED, result=None)` for the sub-µs gap and are expected to re-poll. Doc now matches the actual guarantee.

No behavioral change. Tests still 31/31 on `test_run_func.py`.

Per your sign-off equivalent — ready for human reviewer + merge.

iamcxa (commented):
@iamcxa — final sign-off after `e175c783` + `9eb96f17`.

**All clear — would APPROVE if you weren't the author** (GitHub blocks self-approve). v2 findings closed; bonus project-knowledge capture in `CLAUDE.md`. Sign-off equivalent: ready for human reviewer + merge.

## v2 → v3 Closure

| v2 finding | v3 status |
|-----------|-----------|
| LOW (cancel sentinel sub-µs GIL race) | ✅ Addressed via doc — "best-effort against the macroscopic race... use `threading.Lock` or `threading.Event` if a future caller needs full atomicity" |
| NIT (snapshot reader docstring nuance) | ✅ Addressed — completion-signal vs snapshot reader distinction documented; sub-µs `(FINISHED, result=None)` window acknowledged |

## Bonus

`9eb96f17` adds a *Cloud vs Local Mode Exception Types* section to `CLAUDE.md` capturing the cloud-vs-local exception class divergence rule (from v1 finding #1, RecceCloudException). This is exactly the project-knowledge promotion I would have proposed in D2 — captured proactively without prompting. 👍

## Closure Summary

| Round | Commit | Inline | Body |
|-------|--------|--------|------|
| v1 | `76eac07` | 12 (2 HIGH + 4 MEDIUM + 6 LOW/NIT) | full review |
| v2 | `8385e0a7` | 2 (1 LOW + 1 NIT) | closure + sign-off |
| v3 | `e175c783` | 0 | final sign-off |

## Verification Matrix (final)

| # | Concern | Status |
|---|---------|--------|
| V1 | `submit_run` race fixed | ✅ sync `update_run_result`, status-before-result, cancel sentinel preserved |
| V2 | All 6 callers benefit | ✅ ordering protects completion-signal readers |
| V3 | `_tool_run_check` matches `_tool_create_check` policy | ✅ gate + persistence + comment all aligned |
| V4 | Removed sleep workarounds | ✅ regression tests added (cancel sentinel + repr fallback + negative-path metadata) |

Ready to ship.

wcchang1115:
## Code Review — PR #1342 (DRC-3307) — Adversarial Re-review

### Summary
Re-reviewed every change with the assumption it's wrong until proven otherwise. **All changes prove valid under scrutiny.** No critical, high, or medium-severity bugs found. Two test gaps remain.

---

## Adversarial proofs of validity

### Claim 1: `update_run_result` async→sync is safe across ALL 6 callers

**Suspect:** Removing `async` and `run_coroutine_threadsafe` could break callers that depended on async scheduling timing or expected the old "result-set-before-status" ordering.

**Proof:**
- **Memory ordering:** After `loop.run_in_executor(None, fn)`, the future resolves only when `fn()` returns. By Python's asyncio spec, `await future` resumes via `call_soon_threadsafe`, which acts as a memory barrier — all writes from the executor thread (including `run.status`, `run.result`, `run.error`) are visible to the awaiter.
- **All 6 callers verified:**

  | Caller | File:Line | Reads after `await future` | Safe |
  |---|---|---|---|
  | `create_run_handler` | `run_api.py:66` | None (returns `run` to FastAPI) | ✅ |
  | `rerun_check_handler` | `check_api.py:182` | None (returns `run`) | ✅ |
  | `wait_run_handler` | `run_api.py:85` | Polls `run.result is None and run.error is None` — only sees one field | ✅ |
  | `_tool_run_check` (task branch) | `mcp_server.py:1746` | `run.status == FINISHED and not run.error` | ✅ |
  | `_tool_create_check` | `mcp_server.py:1833` | `run.status == FINISHED` | ✅ |
  | `execute_preset_checks` | `run.py:176` | `run_should_be_approved(run)` reads `run.error`, `run.result` | ✅ |
  | `execute_state_checks` | `run.py:258` | None | ✅ |

- **Status-BEFORE-result inversion proof:** New order is `status = FINISHED; run.result = result`. A reader observing `run.result is not None` is guaranteed (under GIL serialization at executor thread's GIL release point) to also see `run.status == FINISHED`. The reverse window from the OLD code (`run.result` set, but status still `RUNNING`) was the actual DRC-3307 bug — fixed.
- **`updated_params` merge moved to top:** Previously between result-write and status-set on success. Now first. This means when an external observer sees `status == FINISHED`, `params` is also already merged. Strictly safer.
- **Loop captured correctly:** `loop = asyncio.get_running_loop()` at line 147 is still used at line 256 for `run_in_executor`. No dangling references.

**Verdict:** ✅ Proven safe.

### Claim 2: Cancel-sentinel preservation widening is correct

**Suspect:** The OLD success path had NO `if run.status != CANCELLED` guard. Adding it changes behavior — cancelled runs that finish anyway now stay `CANCELLED` instead of becoming `FINISHED`. Could a caller depend on the old behavior?

**Proof:**
- Searched all callers of `cancel_run` — only `cancel_run_handler` in `run_api.py:73`. It does not poll for "cancelled but actually finished."
- No frontend code reads cancelled-then-finished state as a special condition (frontend renders by status enum).
- Test `test_cancel_sentinel_preserved_on_success` explicitly locks in the new behavior.
- `_tool_run_check` task branch: `run_succeeded = run.status == FINISHED and not run.error` — `CANCELLED` correctly maps to `False` (no auto-approve).

**Verdict:** ✅ Proven correct. Behavior change is intentional and tested.

### Claim 3: `run_succeeded` flag covers all control-flow paths through `_tool_run_check`

**Suspect:** The flag could be left in a wrong state along some control-flow path.

**Proof — exhaustive path enumeration:**

| Path | `run_succeeded` final | Auto-approve | Correct? |
|---|---|---|---|
| Metadata branch, success | `True` | Fires | ✅ |
| Metadata branch, RecceException | n/a (re-raised as ValueError) | Skipped | ✅ |
| Metadata branch, RecceCloudException | n/a (propagates, since it's not `RecceException`) | Skipped (run_succeeded never set True) | ✅ |
| Task branch, `submit_run` raises RecceException | n/a (re-raised) | Skipped | ✅ |
| Task branch, `submit_run` raises RecceCloudException | n/a (propagates) | Skipped | ✅ |
| Task branch, run finishes successfully | `True` | Fires | ✅ |
| Task branch, run fails (`status=FAILED`) | `False` | Skipped | ✅ |
| Task branch, run cancelled | `False` (CANCELLED ≠ FINISHED) | Skipped | ✅ |

**Cloud-vs-local exception handling:** Both `update_check_by_id` and `export_persistent_state` are placed OUTSIDE the `try / except RecceException` blocks. Confirmed: `RecceCloudException` extends `Exception`, NOT `RecceException` (`recce/util/recce_cloud.py:31`). The CLAUDE.md addition documenting this is accurate.

**Verdict:** ✅ Proven correct.

### Claim 4: `await asyncio.get_event_loop().run_in_executor(None, export_persistent_state)` cost is acceptable

**Suspect:** Could be expensive enough to regress agent latency.

**Proof:**
- **Local mode:** `state_loader.export()` writes a JSON file. Sub-ms cost.
- **Cloud mode:** HTTPS upload of full RecceState. Cost = O(checks + runs + artifacts). For typical sessions (≤100 checks/runs), this is small (~100KB-few MB upload).
- **Non-blocking:** Wrapped in `run_in_executor`, so event loop continues servicing other coroutines.
- **Idempotent and safe under failure:** If `export_persistent_state` fails after `update_check_by_id` succeeds, cloud is source of truth and remains consistent. Local state file is stale but recoverable on next session.
- **Same pattern as `_tool_create_check:1847`** — consistent.

**Verdict:** ✅ Cost is acceptable. Non-blocking. Could be debounced as a future optimization but not required.

### Claim 5: Test removal of `await asyncio.sleep(0.1)` is correct

**Suspect:** The sleep might have been masking a real race condition that still exists.

**Proof:**
- OLD code: `update_run_result` was async + scheduled via `run_coroutine_threadsafe` from the executor thread. The scheduled coroutine ran in the event loop AFTER `await future` resumed (because `run_coroutine_threadsafe` posts to the loop, which only processes it on the next iteration). The sleep gave the loop time to process the coroutine.
- NEW code: `update_run_result` is called synchronously inside `fn()` BEFORE `fn()` returns. The future resolves only after `fn()` returns. Therefore by the time `await future` resumes, all writes are committed.
- **Regression guard:** `assert run.status == RunStatus.FINISHED` in `test_normalized_params_propagate_to_run` will fail (without the sleep) if anyone reverts to async scheduling.

**Verdict:** ✅ Sleep correctly removed. Regression guard in place.

### Claim 6: CLAUDE.md "Cloud vs Local Mode Exception Types" rule is accurate

**Suspect:** The rule could be wrong about the inheritance hierarchy.

**Proof:**
- Verified `recce/exceptions.py:1`: `class RecceException(Exception)` — base is plain `Exception`.
- Verified `recce/util/recce_cloud.py:31`: `class RecceCloudException(Exception)` — base is plain `Exception`, NOT `RecceException`.
- Therefore `try / except RecceException` does not catch `RecceCloudException`. Documentation accurate.

**Verdict:** ✅ Accurate.

---

## Findings

### Finding 1 — [Low priority, test gap] Task-branch negative paths not asserted
**File:** `tests/test_mcp_server.py`

**Issue:** No test asserts that auto-approve does NOT fire when:
- `mock_run.status = RunStatus.FAILED`, OR
- `mock_run.error = "some error"` while status is FINISHED

**Proof of gap:** If the gate `run.status == RunStatus.FINISHED and not run.error` is accidentally weakened to `True` (or only one of the two clauses), the existing tests still pass because they only exercise the success case.

**Suggestion:** Add two tests mirroring `test_tool_run_check_metadata_branch_recce_exception_no_auto_approve` for the task branch (one for FAILED status, one for non-None error). 5–10 lines each.

### Finding 2 — [Low priority, test robustness] `test_cancel_sentinel_preserved_on_success` has timing dependency
**File:** `tests/apis/test_run_func.py:399-422`

**Issue:** The test sets `run.status = RunStatus.CANCELLED` AFTER `submit_run` returns but BEFORE awaiting the future. This relies on the test thread winning the race against the executor's `fn()` execution. With a fast mock task, the executor could have already completed `update_run_result(run, result, ...)` before the test sets `CANCELLED` — in which case the assertion passes for the wrong reason (the test just observes its own write).

**Proof:** No `threading.Event` or barrier blocks `fn()` from running ahead of the test's CANCELLED write. The test happens to work because Python's executor submission has small overhead, but this is implementation-detail timing.

**Suggestion:** Inject a `threading.Event` into the mock task's `execute()` and have the test set CANCELLED, then signal the event, before awaiting the future. Deterministic.

### Finding 3 — [Low priority, cleanup] Redundant `run.result = await future`
**File:** `recce/apis/run_api.py:66`, `recce/apis/check_api.py:182`, `recce/mcp_server.py:1745`

**Issue:** With `update_run_result` now setting `run.result` synchronously inside `fn()` before the future resolves, these three lines overwrite `run.result` with the same value (or with `None` on failure, where it was already `None`).

**Suggestion:** Replace with bare `await future`. Not a bug — pure cleanup.

### Finding 4 — [Low priority, cosmetic] Brittle line reference in comment
**File:** `recce/mcp_server.py:1751-1752`

**Issue:** Comment says "_tool_create_check (line 1832: ...)" but actual line is 1843. Line numbers rot.

**Suggestion:** Drop the line number; symbolic reference suffices.

---

## Tests verification
- `pytest tests/apis/test_run_func.py tests/test_mcp_server.py` → **146 passed**, 0 failed
- `flake8 recce/apis/run_func.py recce/mcp_server.py` → clean

---

## Verdict

✅ **Approve.**

Every suspicious change has a defensible justification. The race-condition fix is correctly implemented; the auto-approve flag covers all control-flow paths; the cloud-exception placement matches the documented rule. No CRITICAL or HIGH-severity issues. Findings 1–4 are LOW-priority polish that can land in a follow-up.

wcchang1115 (approved):

PR #1346, opened 2026-04-29
fix(lineage): dedupe /api/models/{id} fetch on node click

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

bug fix

`NodeViewOss` (the panel that opens on lineage node click) was firing `GET /api/models/{id}` **twice** in parallel for the same model. Two co-located fetchers raced for the same payload:

1. `NodeViewOss.tsx:117` — `useQuery({ queryKey: ["modelDetail", node.id], staleTime: 5 * 60 * 1000 })`
2. `useModelColumns(node.data.name)` — a hand-rolled `useEffect` + `useState` flow that called `getModelInfo` directly, bypassing TanStack Query (no cache key, no in-flight dedupe).

This rewrites `useModelColumns` to use `useQuery` with the same `["modelDetail", node.id]` key and the same 5-minute `staleTime`. Two consumers of the same key on the same render tree → one network request.

`NodeSqlViewOss`, `SandboxViewOss`, `SchemaDiffView`, and `SchemaSummary` already use the same `["modelDetail", id]` key, so this dedupes for **all five** consumers — not just `NodeViewOss`.

Also drops dead exports (`EMPTY_COLUMNS`, `extractColumns`) — grep found zero external consumers.

Resolves DRC-3343

- The hook's public API (`{ columns, primaryKey, isLoading, error }`) is preserved. Field shapes unchanged.
- One subtle behavior tightening: when the hook is called with no `apiClient` resolvable, the old code returned `isLoading=true` forever (the `fetchData` early-return left state stuck); the new code returns `isLoading=false` via TanStack's `enabled` flag. That's strictly an improvement, not a regression.
- New regression test `js/packages/ui/src/hooks/__tests__/useModelColumns.test.tsx`: under one `QueryClient`, `useModelColumns(model)` plus a sibling `useQuery({queryKey: ["modelDetail", id]})` together → exactly **one** `getModelInfo` call.
- Verification run: \`pnpm lint:fix\`, \`pnpm type:check\`, \`pnpm test\` (156 files, 3703 pass, 0 fail).

NONE

### DIFF
--- a/js/packages/ui/src/hooks/__tests__/useModelColumns.test.tsx
+++ b/js/packages/ui/src/hooks/__tests__/useModelColumns.test.tsx
@@ -1,0 +1,181 @@
+/**
+ * @file useModelColumns.test.tsx
+ * @description Tests for the useModelColumns hook (DRC-3343).
+ *
+ * Verifies:
+ * - The hook resolves columns + primary_key from the /api/models/{id} payload.
+ * - When useModelColumns and a sibling useQuery({queryKey: ["modelDetail", id]})
+ *   mount under one QueryClient, only ONE network request fires
+ *   (regression guard for DRC-3343, where the panel double-fetched).
+ * - The hook returns isLoading=false when no model resolves (matches the
+ *   prior hook contract).
+ */
+
+import { vi } from "vitest";
+
+// ============================================================================
+// Mocks - MUST be set up before imports
+// ============================================================================
+
+const mockApiClient = {
+  get: vi.fn(),
+  post: vi.fn(),
+  put: vi.fn(),
+  delete: vi.fn(),
+};
+
+const mockGetModelInfo = vi.fn();
+
+vi.mock("../../api", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../api")>();
+  return {
+    ...actual,
+    getModelInfo: (...args: unknown[]) => mockGetModelInfo(...args),
+  };
+});
+
+const mockUseLineageGraphContext = vi.fn();
+vi.mock("../../contexts/lineage", () => ({
+  useLineageGraphContext: () => mockUseLineageGraphContext(),
+}));
+
+vi.mock("../../providers", () => ({
+  useApiConfigOptional: () => ({ apiClient: mockApiClient }),
+}));
+
+// ============================================================================
+// Imports
+// ============================================================================
+
+import {
+  QueryClient,
+  QueryClientProvider,
+  useQuery,
+} from "@tanstack/react-query";
+import { renderHook, waitFor } from "@testing-library/react";
+import type { ReactNode } from "react";
+import { useModelColumns } from "../useModelColumns";
+
+// ============================================================================
+// Test Setup
+// ============================================================================
+
+const NODE_ID = "model.test.my_model";
+const MODEL_NAME = "my_model";
+
+const lineageNode = {
+  id: NODE_ID,
+  data: { id: NODE_ID, name: MODEL_NAME },
+};
+
+const createWrapper = () => {
+  const queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false } },
+  });
+  return ({ children }: { children: ReactNode }) => (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+  );
+};
+
+const fakeModelInfo = {
+  model: {
+    base: {
+      columns: { id: { name: "id", type: "INT" } },
+      primary_key: "id",
+    },
+    current: {
+      columns: {
+        id: { name: "id", type: "INT" },
+        name: { name: "name", type: "TEXT" },
+      },
+      primary_key: "id",
+    },
+  },
+};
+
+describe("useModelColumns", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockGetModelInfo.mockReset();
+    mockUseLineageGraphContext.mockReturnValue({
+      lineageGraph: { nodes: [lineageNode] },
+    });
+  });
+
+  it("resolves columns and primaryKey from /api/models/{id}", async () => {
+    mockGetModelInfo.mockResolvedValue(fakeModelInfo);
+
+    const { result } = renderHook(() => useModelColumns(MODEL_NAME), {
+      wrapper: createWrapper(),
+    });
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(mockGetModelInfo).toHaveBeenCalledTimes(1);
+    expect(mockGetModelInfo).toHaveBeenCalledWith(NODE_ID, mockApiClient);
+    expect(result.current.columns.map((c) => c.name)).toEqual(["id", "name"]);
+    expect(result.current.primaryKey).toBe("id");
+    expect(result.current.error).toBe(null);
+  });
+
+  it("dedupes the /api/models/{id} fetch when a sibling useQuery shares the cache key (DRC-3343)", async () => {
+    mockGetModelInfo.mockResolvedValue(fakeModelInfo);
+
+    // One QueryClient shared by both hooks — the dedupe under test.
+    const queryClient = new QueryClient({
+      defaultOptions: { queries: { retry: false } },
+    });
+    const wrapper = ({ children }: { children: ReactNode }) => (
+      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+    );
+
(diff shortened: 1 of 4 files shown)


gcko:
# Code Review: PR #1346

**Reviewed at:** `61079a72` (fix/drc-3343-dedupe-model-info-fetch)
**Files reviewed:** 4
**Categories:** Frontend hook + tests + barrel exports

## Validation Results

- `_.find(lineageGraph?.nodes, {data:{name: model}})` works on both `Record<string, LineageGraphNode>` (production type per `contexts/lineage/types.ts:87`) and an array (test fixture). lodash iterates values for objects, elements for arrays — same predicate semantics.
- `enabled = !!apiClient && !!node?.id` correctly gates the query: undefined model → no node found → `node?.id` falsy → `enabled=false`. The defensive throw inside `queryFn` is unreachable when `enabled` is honored, but harmless.
- `useMemo` dep `[data]` returns a stable `{columns, primaryKey}` object across renders for the same `data` reference (TanStack v5 returns the same reference from cache hits). The DRC-3260 infinite-loop hazard from `[]` literals in render is not reintroduced — the literal lives inside the memoized branch.
- Error normalization `queryError instanceof Error ? queryError : new Error(String(queryError))` is correct.
- The early-return path `if (!modelInfo?.base.columns || !modelInfo.current.columns)` mirrors the old hook's check. `!modelInfo?.base.columns` is true only when `columns` is `undefined`/`null` — `!{}` evaluates to `false`, so empty-columns objects fall through to the union path.

- `getModelInfo(model: string, client: ApiClient): Promise<ModelInfoResult>` (api/info.ts:229) — call site matches.
- `ModelInfoResult.model.{base,current}: ModelEnvDetail` with `columns?: Record<string, NodeColumnData>` and `primary_key?: string` — `Object.values(modelInfo.base.columns)` and `modelInfo.current.primary_key` are correct.
- `useApiConfigOptional(): ApiContextValue | null` (providers/contexts/ApiContext.tsx) — `?.apiClient` chain is correct.
- All 5 consumers verified using `["modelDetail", <id>]` with `staleTime: 5 * 60 * 1000`:
  - `NodeViewOss.tsx:136`, `NodeSqlViewOss.tsx:41`, `SchemaSummary.tsx:27` — `["modelDetail", node.id]`
  - `SchemaDiffView.tsx:205` — `["modelDetail", selectedNode?.id]`
  - `SandboxViewOss.tsx:66` — `["modelDetail", current?.id ?? ""]` (uses `?? ""` fallback; harmless mismatch with the others when id exists)
- `extractColumns` is dropped from both barrel exports (`hooks/index.ts`, `index.ts`); zero external consumers found across `js/packages`, `js/app`, and `js/src`.

- `enabled` flag prevents the queryFn from running with undefined apiClient or node id; no runtime crash from the defensive throw.
- `error` is always an `Error` instance or `null`; consumers can rely on `error?.message` access.
- The `if (!modelInfo?.base.columns || !modelInfo.current.columns)` guard handles partial responses (added/removed models with one-sided columns).

- 4 new tests in `useModelColumns.test.tsx` cover: happy path resolution, regression dedupe under shared QueryClient, undefined-model short-circuit, error surface.
- The dedupe regression test (lines 123-154) is the load-bearing one for DRC-3343 — asserts `mockGetModelInfo).toHaveBeenCalledTimes(1)` when `useModelColumns` and a sibling `useQuery` with matching key mount under the same QueryClient. This will catch any future drift in queryKey/staleTime alignment.
- Test fixture uses `nodes: [lineageNode]` (array) but production type is `Record<string, LineageGraphNode>` — see Notes.
- Missing test: explicit assertion of `isLoading=false` when `useApiConfigOptional()` returns `null` (the documented "improvement" over the prior contract). Currently the test always provides a working `apiClient` mock.

- `extractColumns` removal: zero external consumers under `js/`. Safe deletion. `EMPTY_COLUMNS` was a module-level non-exported const — see Notes.
- Public hook return type `UseModelColumnsReturn = { columns, primaryKey, isLoading, error }` preserved — verified against all callers (`TopKDiffForm`, `ValueDiffForm`, `ProfileDiffForm`, `HistogramDiffForm`, `LineageViewContextMenuOss`, `NodeViewOss`).
- One subtle behavior change in `primaryKey` resolution that the PR description doesn't call out — see Notes.

- Dedupe is the entire point of the PR. With one QueryClient and matching `["modelDetail", id]` keys, TanStack Query collapses concurrent in-flight requests and reuses cached responses for `staleTime`.
- `useMemo` keyed on `data` keeps `columns`/`primaryKey` reference-stable across renders, reducing downstream re-renders compared to the old `useState`-driven implementation.

- No new fire-and-forget tasks. `queryFn` returns the awaitable from `getModelInfo` — TanStack manages the lifecycle.
- No race against the prior render's `useState`/`useEffect` interplay (the DRC-3260-era stale-state hazard).

### Verification Results

- `pnpm type:check` — passing (tsc --noEmit, 0 errors)
- `pnpm lint` — passing (Biome, 610 files, 0 errors)
- `pnpm test` — passing (156 files, 3703 passing / 5 skipped / 0 failed)

## Verdict: GO

No BLOCKERs, no ISSUEs.

### Notes

1. **Undocumented behavior change: `primaryKey` is now returned even when one env's columns are missing.** Old code's `try` block did `setPrimaryKey(modelInfo.current.primary_key)` only after the `!modelInfo.base.columns || !modelInfo.current.columns` early-return; in the early-return branch `primaryKey` retained its prior state value (typically `undefined`). New code at `useModelColumns.tsx:113-118` returns `primaryKey: modelInfo?.current.primary_key` from the same branch, so an added/removed model whose `current.primary_key` is set will now produce a non-undefined `primaryKey`. This is a strict improvement (the user can use the primary key for an added model's query/value diff), but the PR description's claim of "field shapes unchanged" understates it — the *value semantics* changed for a real edge case. Worth a line in the PR description so consumers like `LineageViewContextMenuOss.tsx:70` and `NodeViewOss.tsx:130` (which feed `primaryKey` into `setPrimaryKeys`) know the new contract.

2. **PR description: `EMPTY_COLUMNS` was never exported.** The PR description says "drops dead exports (`EMPTY_COLUMNS`, `extractColumns`)". Only `extractColumns` was exported (verified against the diff and both barrels); `EMPTY_COLUMNS` was a module-level non-exported const. Description nit, not a code issue.

3. **Test fixture shape mismatch.** `useModelColumns.test.tsx:101` mocks `lineageGraph: { nodes: [lineageNode] }` (array) while the production type is `Record<string, LineageGraphNode>`. lodash `_.find` happens to work on both, so the test passes — but a regression where the hook switches to `Object.values(lineageGraph.nodes)` or a direct key lookup wouldn't be caught. A faithful `Record<id, node>` mock would close that gap.

4. **Pre-existing: name-collision risk in node resolution.** `_.find(lineageGraph?.nodes, {data:{name: model}})` returns the first node by iteration order. If two nodes share the same `data.name` (e.g., dbt models in different packages), `useModelColumns(name)` could resolve to a different `node.id` than `NodeViewOss` displays — defeating the dedupe this PR is establishing. Not introduced by this PR; flagging because it's the underlying mechanism the dedupe depends on.

5. **Cache-hit removes the `isLoading=true` flash for warm renders.** Old code always passed through `isLoading=true` for one render before resolving; new code returns `isLoading=false` immediately on cache hit. Consumers (`TopKDiffForm`, `ValueDiffForm`, `ProfileDiffForm`, `HistogramDiffForm`) that conditionally render loading skeletons will skip that flash on warm reopens. This is the intended dedupe behavior and arguably a UX win, but it is a behavior change worth noting.

## What I could not verify

- End-to-end browser confirmation that `/api/models/{id}` fires once on a real node click (the regression test under one QueryClient is the proxy). Did not run `recce server`.
- Whether any recce-cloud-infra or downstream package consumes `extractColumns` outside this monorepo (grep covered the OSS repo only).

- queryKey/staleTime drift across the 5 dedupe consumers (all aligned).
- Stale references back to `EMPTY_COLUMNS` or `extractColumns` in `js/`.
- Mock-vs-production signature drift on `getModelInfo`.
- Hook return-type breaking change for the 6 callers.
- Infinite-loop hazard from a `[]` literal in render (the DRC-3260 trap).
- Missing-await or fire-and-forget Promise leaks.
- TanStack v5 API misuse (`enabled`, `staleTime`, queryKey shape are all correct).

gcko (approved):
Claude Code Review: GO. No critical issues found. See PR comment for full review (5 NOTEs, no BLOCKERs/ISSUEs).

PR #1348, opened 2026-04-30
fix(cll): dedupe per_node.db columns case-insensitively (catalog wins)

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` — bug fix in the cloud-precompute per-node SQLite emitter.

The per-session `per_node.db` emitter unioned `catalog.json` and `manifest.json` column names with a **case-sensitive** seen-set. Snowflake's `catalog.json` reports identifiers UPPERCASE (warehouse truth); `schema.yml` typically declares them lowercase. The same logical column ended up in the `columns` table twice — once as `ID`, once as `id`.

Cloud's `/sessions/{id}/models/{model_id}` reader (`per_node_db_reader.py`) builds `columns_info[col_name]` keyed on raw name, so the side panel renders both casings as separate columns.

Verified in two production sessions:

| Session | Adapter | (node, env) groups affected | Total groups | % affected |
|---|---|---:|---:|---:|
| `122/59/.../1d050866-…` (snaptravel) | snowflake | 1,379 | 4,830 | **28.5%** |
| `6/163/.../d33a8778-…` (jaffle-shop-expand) | snowflake | 1,048 | 2,120 | **49.4%** |

Concrete example: `model.snaptravel.fact_bookings` (env=base) had both `ADDONS_AMOUNT_LOCAL` (from catalog) and `addons_amount_local` (from manifest) in the precomputed columns table.

**Fix**: dedup by `lower(col_name)` while preserving original casing. Catalog rows are written first (their casing + dtype win); manifest-only columns still come through with their declared casing. This mirrors local-mode `DbtAdapter.get_model()` which queries the warehouse via `get_columns()` — a single source of truth.

Resolves DRC-3349
Refs DRC-3295 (per_node.db emitter)

- 56 LOC total (12 prod + 44 test).
- TDD: added regression test `test_catalog_wins_over_manifest_when_column_casing_differs` first, confirmed red, then applied fix → green.
- Existing scale tests against jaffle-shop-expand artifacts still pass (`tests/test_cli_per_node_db.py`).
- Existing S3 `per_node.db` artifacts still contain duplicate rows; no backfill needed — they re-emit cleanly on the next session upload after this ships.

```release-note
Fixes column-list duplication on Snowflake projects in the Recce Cloud lineage side panel — the same column no longer appears twice (uppercase + lowercase).
```


even-wei:
# Code Review: PR #1348

**Files reviewed:** 2 (`recce/util/per_node_db.py`, `tests/test_per_node_db.py`)
**Categories:** Business logic
**Passes run:** A, C, D, E, F, G
**HEAD reviewed:** `e74a802d`

## Validation Results

### Pass A: Correctness & Logic — FAIL

The catalog/manifest dedup is now case-insensitive — that part works. But the same **catalog-uppercase vs manifest-lowercase** mismatch this PR is fighting also breaks **`_pick_primary_key`** (`recce/util/per_node_db.py:307-323`, called from L384-L385), and that path was not updated.

**Trace** on a Snowflake project (catalog `ID`, schema.yml `id`, dbt-generated test name `unique_<model>_id`):

1. After this fix, `ordered_cols = ["ID"]` (catalog wins, manifest `id` deduped).
2. `_derive_tests` parses the test name and produces `NodeTestRow(column_name="id", ...)` — manifest casing, unchanged by this PR.
3. `unique_cols = {"id"}` at L384.
4. `_pick_primary_key(["ID"], {"id"})` checks `"ID" in {"id"}` → False → returns `None`.

**Empirical confirmation** on real Snowflake `jaffle-shop-expand` artifacts (`adapter_type: "snowflake"`, 1949 nodes, 275 nodes that have unique tests):

```
Total nodes: 1949
Nodes with primary_key set: 0   # was non-zero before this PR
```

Before this PR, the manifest-cased `id` row survived in `ordered_cols`, so `_pick_primary_key(["ID", "id"], {"id"})` matched on the second iteration and returned `"id"`. After this PR, no Snowflake node ever gets a `primary_key`. Every Snowflake project regresses.

**Downstream effect (cloud reader, `api_server/core/recce_context/per_node_db_reader.py`):**

- L213-L222 builds `tests_by_column` keyed on raw `node_tests.column_name` — lowercase on Snowflake.
- L237-L245 iterates `column_rows` (now uppercase only after this PR) and looks up `tests_by_column.get(col_name, set())` — `col_name="ID"` against `{"id": {...}}` → no match. `not_null` and `unique` flags are silently dropped from every column on every Snowflake model.
- L249-L251 reads `node_row["primary_key"]`, which is now `None` on every Snowflake node, so the API response never includes `"primary_key"`.

Net trade-off: the PR removes a visible bug (duplicate columns in the side panel) and replaces it with a **silent** one (missing PK marker, missing not_null/unique markers on every Snowflake model). Silent regressions are worse — users can't see them to report them.

The minimal local fix is to canonicalize in `_pick_primary_key` too:

```python
def _pick_primary_key(ordered_column_names, unique_column_names):
    canonical_unique = {c.lower() for c in unique_column_names}
    for col in ordered_column_names:
        if col.lower() in canonical_unique:
            return col
    return None
```

Even better: also write `node_tests.column_name` using the catalog casing when one exists, so the cloud reader's lookup just works without a separate change. Otherwise `recce-cloud-infra/api_server/core/recce_context/per_node_db_reader.py` needs a parallel fix to do case-insensitive joins.

### Pass C: Cross-Reference Consistency — FAIL

Same root cause as Pass A. Column identity lives in three places that must agree on case:

1. `columns.column_name` — now catalog casing (UPPERCASE on Snowflake) after this PR.
2. `node_tests.column_name` — manifest casing (lowercase from test-name parsing) — unchanged.
3. `nodes.primary_key` — derived from (1) ∩ (2) — now broken on Snowflake.

The PR fixes (1) but leaves (3) and any joiner of (1) ↔ (2) inconsistent.

### Pass D: Error Handling — PASS

`col_name.lower()` is safe — JSON dict keys are always strings. No new exception paths.

### Pass E: Test Coverage — FAIL

- The new test `test_catalog_wins_over_manifest_when_column_casing_differs` (`tests/test_per_node_db.py:324-365`) covers column dedup but **omits any unique tests**, so it never triggers `_pick_primary_key` and misses the regression above.
- The pre-existing `test_primary_key_follows_catalog_column_order_with_multiple_unique_tests` uses lowercase consistently in both catalog and manifest, so it can't catch the cross-casing scenario either.
- `tests/test_cli_per_node_db.py::TestContractWithGetModel::test_primary_key_and_tests_derived_from_manifest_only` (L878-L899) DOES assert `WHERE primary_key IS NOT NULL` returns at least one row on real artifacts — and that assertion would FAIL on a real Snowflake fixture. It currently passes only because `tests/fixtures/jaffle-shop-expand/{manifest,catalog}.json` aren't checked into the repo, so the fixture loader at L800-L801 calls `pytest.skip(...)`. The PR description's claim that "Existing scale tests against jaffle-shop-expand artifacts still pass" is technically true but misleading — the most relevant assertion is silently skipped.

A regression test that pins the fix should:
- Match the existing test's shape but use catalog UPPERCASE + manifest/test lowercase.
- Assert `primary_key == "ID"` (catalog casing) and that `unique` markers are recoverable.

- Function signature unchanged, comment block correctly explains the behavior change.
- One minor side-effect of sharing `seen_canonical` across both loops: if a single catalog ever had internal case duplicates (`{"ID": ..., "id": ...}` in the same node), the previous code kept both; this code keeps only the first. Practically irrelevant — warehouses don't return case-duplicate columns from `INFORMATION_SCHEMA`. Not a finding.

One `.lower()` per column. Negligible at any realistic scale.

### Verification Results

- `uv run ruff check recce/util/per_node_db.py tests/test_per_node_db.py` → clean.
- `uv run pytest tests/test_per_node_db.py` → 13/13 pass.
- `uv run pytest tests/test_cli_per_node_db.py` → 11 pass, 5 skipped (scale tests skipped without `RECCE_SCALE_FIXTURE_DIR`; contract tests skipped because `tests/fixtures/jaffle-shop-expand/` is absent).
- Empirical run via `extract_rows_from_artifacts` on a real Snowflake jaffle-shop-expand snapshot: 0/1949 nodes get `primary_key` set.

## Verdict: NO-GO

### Blockers

1. **`_pick_primary_key` regresses to `None` on every Snowflake model** — `recce/util/per_node_db.py:320-321`. The case-sensitive `if col in unique_column_names` check fails to match catalog-uppercase against test-derived-lowercase. Empirically: 0/1949 nodes get a primary_key on the Snowflake jaffle-shop-expand fixture. Fix `_pick_primary_key` to compare case-insensitively (or normalize `node_tests.column_name` to catalog casing at write time).
2. **Cloud reader's per-column test markers silently drop on Snowflake** — `recce-cloud-infra/api_server/core/recce_context/per_node_db_reader.py:213-245`. `tests_by_column.get(col_name, set())` joins lowercase keys against now-uppercase column names. Either fix is acceptable: (a) join case-insensitively in the reader, or (b) write `node_tests.column_name` with catalog casing in this PR so the existing reader code keeps working.

### Issues

1. **The new test for the case-mismatch scenario doesn't exercise primary_key or test-marker recovery** — `tests/test_per_node_db.py:324-365`. Add a unique/not_null test on the catalog-uppercase columns and assert `primary_key == "ID"` and that `node_tests.column_name` joins back to a column row.
2. **`TestContractWithGetModel::test_primary_key_and_tests_derived_from_manifest_only` would fail on real Snowflake fixtures**, but the fixtures aren't checked in so it skips silently. Consider checking in a small Snowflake-cased fixture (or at least one synthetic node with mixed casing) so this regression class is caught in CI.

### Notes

1. PR description says "Existing scale tests against jaffle-shop-expand artifacts still pass." That's true but the most relevant assertion is being skipped for missing fixtures — worth qualifying or removing that line.

## What I could not verify

- Whether the cloud reader has an in-flight branch I'm not seeing that already does case-insensitive joins. The local checkout of `recce-cloud-infra` has zero `.lower()`/`.upper()` calls in `per_node_db_reader.py`, and the file's last commit on `main` is `dc861c4f feat(api): flip /models/{id} to S3-first with per_node.db backend` — the case-sensitive shape this PR's description quotes.
- Whether the frontend renders `primary_key` and `not_null`/`unique` markers prominently enough that operators will notice them disappearing on Snowflake projects after deploy. If the answer is "no one looks at them" then the trade-off is less bad — but the data contract is still broken.

- Injection / auth / PII concerns — none; the change is pure data manipulation.
- Performance regression at scale — extra `.lower()` calls are negligible.
- Boundary issues with empty/None catalog or manifest — handled the same as before.
- Issues with `seen_canonical` being shared across catalog and manifest loops — that's the correct design, what makes the dedup work across sources.

even-wei:
Thanks — both blockers were real. Fixed in `53759bd2`.

## Blocker 1 (Pass A) — `_pick_primary_key` regresses to `None` on Snowflake — FIXED

Verified by reading `recce/util/per_node_db.py:307-323` and tracing on the local jaffle-shop-expand Snowflake fixture: 0/1949 nodes had `primary_key` set after the round-1 commit. The reviewer's empirical claim was correct.

**Fix taken:** instead of patching `_pick_primary_key` in isolation, normalize `NodeTestRow.column_name` to the casing we just stored in `column_rows` for the same node. That keeps the primary_key derivation working *and* fixes Blocker 2 below in one shot — `node_tests.column_name` and `columns.column_name` now agree on case at write time, so no downstream consumer needs a parallel case-insensitive change.

After `53759bd2`, against the same Snowflake fixture:

```
Total nodes: 1949
Nodes with primary_key set: 275           # matches the 275 unique tests, 1:1
Total tests: 786 (unique: 275)
Tests joinable to columns by exact name: 786/786
```

## Blocker 2 (Pass A) — Cloud reader silently drops `not_null`/`unique` markers on Snowflake — FIXED

Verified by reading `recce-cloud-infra/api_server/core/recce_context/per_node_db_reader.py:213-245`. `tests_by_column` is keyed on raw `node_tests.column_name`; `_build_model_payload` does `tests_by_column.get(col_name)` against `column_rows[i].column_name`. Both come from the per_node.db DRC-3349 emitter, so once those two columns agree on case (the round-2 fix above), the existing reader code joins correctly. **No change required in recce-cloud-infra.**

I prefer this over fixing the reader because it keeps a single source of truth for column identity in the precompute artifact: `columns.column_name == node_tests.column_name == nodes.primary_key`. Anything else that joins these tables in the future inherits the invariant for free.

## Pass C — Cross-Reference Consistency — FIXED

Same root cause as Pass A. After `53759bd2`, all three locations (column_rows, node_tests, primary_key) use the same casing per node — catalog-uppercase on Snowflake when catalog is present, manifest casing for manifest-only columns.

## Pass E — Test gap — FIXED

`test_catalog_wins_over_manifest_when_column_casing_differs` now adds `unique_m_id` and `not_null_m_amount` tests on the catalog-uppercase columns and asserts:

- `node_tests.column_name == "ID"` and `"AMOUNT"` (test casing remapped to catalog casing).
- `primary_key == "ID"` (would have been `None` under round-1).

So this test would have caught the regression before round-2.

## Issue 2 / Note 1 — fixture-skipped contract test + PR description claim — Acknowledged, scope-deferred

The reviewer is right that `tests/test_cli_per_node_db.py::TestContractWithGetModel::test_primary_key_and_tests_derived_from_manifest_only` is silently skipped due to the missing `tests/fixtures/jaffle-shop-expand/` directory and that the PR description's "scale tests still pass" line was misleading. Adding a small Snowflake-cased fixture to that directory is a worthwhile follow-up but lives outside this fix's scope — I'll file a separate Linear issue for it under DRC-3295. The new unit test added here covers the regression class deterministically without external fixtures.

## Verification re-run

- `uv run pytest tests/test_per_node_db.py tests/test_cli_per_node_db.py` → 24 passed, 5 skipped (unchanged from round-1; skipped tests still need fixtures that aren't checked in).
- `make format && make flake8` → clean.
- Empirical re-run on real Snowflake artifacts: see Blocker 1 numbers above.

## What I could not address

The reviewer's "Notes" mention checking whether the frontend renders `primary_key` and the test markers prominently — I didn't dig into the frontend in this round; the data contract is now correct, so any rendering issue is a separate concern.

even-wei:
## Round-3: catalog is authoritative when present (`05df3532`)

**Why:** During verification on staging session `d307961c-…`, found 80 lowercase rows surviving round-2 dedup. They turned out to be phantom schema.yml columns — declared in `schema.yml` but never actually selected by the model SQL. Concrete case: `model.jaffle_shop.dq_missing_invoices` declares `order_status` in schema.yml, but its compiled SQL doesn't select it — catalog (warehouse truth) correctly omits it.

Surfacing those phantoms produces two real bugs the round-2 fix didn't address:

1. **False schema-change alarms** when comparing two sessions where one has the phantom column declared and the other doesn't.
2. **SQL errors** when a user tries to profile/query a phantom column — it doesn't exist in the warehouse.

This is also a contract divergence with local-mode `DbtAdapter.get_model()`, which queries the warehouse via `get_columns()` and naturally never returns phantoms.

**Fix:** When catalog has an entry for a node, treat catalog as the authoritative column list. Drop the manifest-overflow path entirely for that node (and drop any `node_tests` rows that referenced now-dropped columns — they'd be join orphans). Manifest fallback still applies for nodes the catalog has nothing for (model never built, or catalog absent entirely).

**Verification on session `d307961c-…` (jaffle-shop-expand, Snowflake):**

| Metric | Before fix | Round-2 | Round-3 |
|---|---:|---:|---:|
| Total columns | 26,290 | 20,228 | **20,148** |
| Case-duplicate (node, env) groups | 2,048 / 2,120 | 0 / 2,120 | 0 / 2,120 |
| Phantom (manifest-only) columns | — | 80 | **0** |
| Lowercase rows | 6,144 | 82 | 2¹ |
| Nodes with `primary_key` | 550 | 550 | 550 |
| `node_tests` rows joinable | 1,572 / 1,572 | 1,572 / 1,572 | 1,572 / 1,572 |

¹ The 2 remaining "lowercase" rows are catalog entries for columns literally named `'0'` — an unaliased numeric literal in the SQL (`select 0` rather than `select 0 as foo`). Catalog has them, so we keep them. Not a regression.

**Test added:** `test_manifest_fallback_when_catalog_has_no_entry_for_node` pins the legitimate fallback path. The existing case-mismatch test was renamed to `test_catalog_is_authoritative_when_present` and extended to assert phantom-column drop + orphan-test drop.

**Files:** `recce/util/per_node_db.py` (+25/-12), `tests/test_per_node_db.py` (+80/-43). All 25 unit tests pass.

cc @even-wei — third commit on the branch, want a final review pass before merge?

even-wei:
# [Round 2] Code Review: PR #1348

**Files reviewed:** 2 (`recce/util/per_node_db.py`, `tests/test_per_node_db.py`)
**HEAD reviewed:** `05df3532`
**New commits since prior review:** `53759bd2` (normalize node_tests casing), `05df3532` (drop phantom schema.yml columns)
**Passes run:** A, C, D, E, F, G

## Status of prior findings

| Prior finding | Status |
|---|---|
| **BLOCKER 1** — `_pick_primary_key` regresses to `None` on every Snowflake model | **RESOLVED** |
| **BLOCKER 2** — Cloud reader's per-column test markers silently drop on Snowflake | **RESOLVED** |
| **ISSUE 1** — New test omits primary_key / test-marker recovery | **RESOLVED** by `test_catalog_is_authoritative_when_present` |
| **ISSUE 2** — Contract test on real Snowflake fixtures would fail | **RESOLVED** (verified — see below) |

**Empirical re-verification on the same Snowflake `jaffle-shop-expand` artifacts:**

```
Total nodes: 1949
Nodes with unique tests: 275
  -> primary_key set:    275   (was 0/275 in round 1)
  -> primary_key None:   0
Orphan tests (column_name not in columns for same node): 0
```

Spot-check: `model.jaffle_shop.stg_products` now has `pk='PRODUCT_ID'`, columns `['PRODUCT_ID', 'PRODUCT_NAME', 'PRODUCT_TYPE']`, and tests `[('PRODUCT_ID', 'not_null'), ('PRODUCT_ID', 'unique')]` — all catalog-cased and join-clean.

`tests/test_cli_per_node_db.py::TestContractWithGetModel` (both `test_extracted_rows_match_get_model` and `test_primary_key_and_tests_derived_from_manifest_only`) **now passes** when pointed at real Snowflake artifacts via `RECCE_SCALE_FIXTURE_DIR`.

## Validation Results (fresh read on the new code)

The two-step approach is correct:

1. `if catalog_cols:` short-circuits the manifest fall-through entirely when catalog covers the node. Comment block at L352-L371 documents the intent (catalog = warehouse truth; phantom schema.yml columns dropped to prevent false schema-change alarms and SQL errors on profile).
2. `canonical_to_stored = {col.lower(): col for col in ordered_cols}` then re-keys `node_tests.column_name` to whatever casing was stored. Tests whose column isn't in `ordered_cols` (phantom) are dropped via `if stored is None: continue`.
3. `unique_cols = {t.column_name for t in normalized_tests if ...}` (L423) — note the rename from `tests` to `normalized_tests`, ensuring `_pick_primary_key` walks against the catalog-cased set. The case-sensitive `if col in unique_column_names` check at `_pick_primary_key:321` now works because both sides are catalog-cased.
4. The `if stored == t.column_name: normalized_tests.append(t) else: rebuild` branch (L409-L420) is a small allocation-saving optimization for the common non-Snowflake path. Functionally equivalent because `NodeTestRow` is `@dataclass(frozen=True)`.

Edge cases traced:
- **Catalog with internal case-duplicates** (e.g., `{"ID": ..., "id": ...}`): first wins via `seen_canonical`. Practically never happens — warehouses don't return case-duplicate columns from `INFORMATION_SCHEMA`.
- **Catalog node entry with `columns: {}` (empty)**: `if catalog_cols:` is falsy → manifest fallback. Slightly surprising but practically irrelevant (no real warehouse table has 0 columns).
- **Phantom test (`not_null_<node>_<col>` where `<col>` is not in catalog OR manifest)**: `_derive_tests` is over-eager and matches anything starting with the prefix; pre-existing behavior. The new normalization now silently drops these — a side-benefit, not a regression.
- **Manifest-fallback path** (catalog has no entry for node): `ordered_cols` from manifest, `canonical_to_stored` built from those, tests normalized to manifest casing. Test markers / primary_key still derive correctly. New unit test `test_manifest_fallback_when_catalog_has_no_entry_for_node` covers this.

All three column-identity-bearing artifacts now agree on case:

1. `columns.column_name` — catalog casing (when catalog covers node) or manifest casing (fallback).
2. `node_tests.column_name` — normalized to (1) via `canonical_to_stored`.
3. `nodes.primary_key` — derived from (1) ∩ (2), so always matches a row in `columns`.

Verified empirically: 0 orphan test rows on the 1949-node Snowflake fixture.

Cloud reader (`recce-cloud-infra/api_server/core/recce_context/per_node_db_reader.py`) is now compatible without modification — its case-sensitive `tests_by_column.get(col_name, set())` joins cleanly because both keys are catalog-cased.

### Pass D: Error Handling — PASS

`col_name.lower()` is safe (JSON keys are strings). No new exception paths.

### Pass E: Test Coverage — PASS

- `test_catalog_is_authoritative_when_present` covers: phantom column drop, test row drop for phantom column, primary_key recovery on catalog-cased columns, `not_null` test normalization, multiple test types on the same node. Asserts the full set of post-conditions.
- `test_manifest_fallback_when_catalog_has_no_entry_for_node` covers the catalog-empty path.
- Real-fixture contract tests pass.
- `unique_cols` derivation (L423) is now correctly sourced from `normalized_tests` rather than `raw_tests` — easy to overlook on review; verified.

### Pass F: Diff-Specific Checks — PASS (with NOTE)

The behavior change is materially stronger than the original PR's "dedup case-insensitively." The current PR now also **drops manifest-only columns entirely** when catalog covers the node. The PR description's "Fix: dedup by `lower(col_name)` while preserving original casing" still describes the round-1 approach; the round-3 follow-up comment captures the new "phantom-column drop" rationale, but the top-of-PR description is stale. See NOTE 1 below.

Per node: O(n_columns) dict construction + O(n_tests) iteration. Negligible at scale (verified scale tests still pass).

### Verification Results

- `uv run ruff check recce/util/per_node_db.py tests/test_per_node_db.py` → clean.
- `uv run pytest tests/test_per_node_db.py` → 14/14 pass (was 13/13 in round 1; new test added).
- `uv run pytest tests/test_cli_per_node_db.py` → 11 pass, 5 skipped (scale tests still gated on `RECCE_SCALE_FIXTURE_DIR`).
- `RECCE_SCALE_FIXTURE_DIR=...jaffle-shop-expand-main pytest TestContractWithGetModel` → both contract tests **pass** (round 1: would have failed if fixtures were checked in).

## Verdict: GO

No BLOCKERs, no ISSUEs.

### Notes

1. **PR description is slightly stale.** The "What this PR does" section still describes the round-1 dedup-only approach ("manifest-only columns still come through with their declared casing"). After `05df3532`, manifest-only columns are dropped entirely when catalog covers the node. Recommend folding the round-3 rationale (phantom column drop, false schema-change alarms, SQL errors on profile) into the PR description so future archaeologists don't have to read the comment thread.

2. **`_pick_primary_key` is still case-sensitive at the implementation level** (`recce/util/per_node_db.py:320-321`). It works now only because the call site at L423 hands it catalog-cased data on both sides. A future change to `_derive_tests` or to how `unique_cols` is built that re-introduces lowercase test names would silently regress primary_key derivation again. Defensive option: make `_pick_primary_key` itself case-insensitive (one-line change). Not blocking — the contract is upheld at the call site today.

3. **Stale-catalog edge case.** If a user adds a column to schema.yml + SQL, runs `dbt run`, but forgets `dbt docs generate`, the catalog is missing the new column. After this PR, the new column will silently disappear from the side panel. Practically uncommon — the canonical `recce init --cloud` workflow is `dbt build && dbt docs generate && recce init --cloud`. Worth being aware of for support cases. Not actionable in this PR.

## What I could not verify

- Whether the cloud reader has been deployed at the version that matches what's on `main` of `recce-cloud-infra`. Even with this fix, sessions uploaded before this ships will still serve duplicate columns / missing markers — that's stated in the PR description and accepted.
- Frontend rendering of `primary_key`. Not scoped to this PR.

- Phantom column or phantom test rows leaking into `columns` / `node_tests`.
- Case-mismatched joins between `columns` and `node_tests` — none on real Snowflake artifacts.
- Performance regression — none observable at the 1918-node scale.
- Behavior changes in the manifest-fallback path that could break non-Snowflake adapters (DuckDB, Postgres) — none; tests on lowercase-everywhere fixtures still pass.

wcchang1115:
## Code Review — PR #1348

### Summary
Dedupes `per_node.db` columns case-insensitively with catalog casing winning. **The change is scoped to the cloud-only `per_node.db` artifact and does not affect the local `/api/info`, `/api/cll`, or `/api/models/{model_id}` endpoints.** Tests pass (`25 passed, 5 skipped`); flake8 clean.

### Answer to the reviewer's specific question

> "is it a safe change to the /info /models /cll these api endpoints? because its case is also important for lineage graph"

**Yes — these local endpoints are untouched by this PR.** Tracing the call graphs:

| Endpoint | Code path | Touches `extract_rows_from_artifacts`? |
|---|---|---|
| `GET /api/info` | `recce/server.py:596` → `context.get_lineage_diff()` → `build_merged_lineage()` | No |
| `POST /api/cll` | `recce/server.py:651` → `dbt_adapter.get_cll()` (sqlglot SQL analysis) | No |
| `GET /api/models/{id}` | `recce/server.py:702` → `context.get_model()` → `dbt_adapter.get_model()` → `get_columns()` (warehouse query) | No |

`extract_rows_from_artifacts` (the only function changed) is consumed in exactly one place: `recce/cli.py:768` inside the `recce init --cloud` upload block, writing the SQLite file that Cloud's `/sessions/{id}/models/{model_id}` reader (the side panel) consumes. The OSS server contains no `per_node.db` reader. The lineage **graph** uses `node_id` strings (e.g. `model.pkg.fact_bookings`) which are not column names and so are unaffected by column casing.

CLL graph rendering is also untouched: `cll_map.json` is built via `dbt_adapter.build_full_cll_map()` from compiled-SQL analysis, not from manifest/catalog `columns` blocks.

### Findings

#### [Warning] PR description undersells the behavior change

**File:** `recce/util/per_node_db.py:372-396`

**Issue:** The PR description says:

> "Fix: dedup by lower(col_name) while preserving original casing. Catalog rows are written first (their casing + dtype win); **manifest-only columns still come through with their declared casing.**"

But the actual code switched from a **union** to an **if/else** — when the catalog has any column for a node, the manifest's `columns` block is now skipped entirely:

```python
if catalog_cols:
    for col_name, col_info in catalog_cols.items():
        ...
else:
    for col_name in manifest_cols.keys():
        ...
```

The added test `test_catalog_is_authoritative_when_present` (line 322) confirms this: a `phantom_col` declared only in `schema.yml` (not in catalog) is **dropped**, not merged. The previous behavior would have unioned it in.

The inline comment (line 352-371) explains and justifies this as intentional ("schema.yml drift produces phantom columns → false schema-change alarms + SQL errors on profile"), but the PR description still reads like a pure case-dedup. This is a meaningful semantic change that reviewers should approve consciously, not gloss over.

**Suggestion:** Update the PR description to call out the catalog-authoritative-when-present behavior explicitly (e.g. "Phantom schema.yml columns absent from the catalog are dropped, mirroring local-mode `DbtAdapter.get_model()` which only sees warehouse columns"). The comment on line 352-371 is a great seed for the description.

#### [Info] Edge case: partial catalog coverage

**File:** `recce/util/per_node_db.py:374`

**Issue:** The `if catalog_cols:` branch treats the catalog as authoritative for the **entire** node as long as it has at least one column. In the rare case where `dbt docs generate` produced a partial column list for a model (column added by post-hook, late-binding view, permission filtering on the catalog query, etc.), real warehouse columns documented only in `schema.yml` would now be dropped from the side panel.

This is unlikely in practice — `information_schema.columns` is queried atomically per relation — but worth noting. Local-mode `get_columns()` queries the warehouse live and so doesn't have this risk. Acceptable trade-off given that the alternative is the duplicate-rows bug, but worth a comment.

**Suggestion:** Either add a one-line note in the existing comment block ("This means partial-catalog nodes lose schema.yml columns; acceptable since dbt's catalog query is atomic per relation") or leave as-is — judgment call.

### Verdict

Approved. The change is scoped, well-tested (catalog-authoritative semantics, manifest fallback, primary-key remap, test-column normalization, phantom-test drop all covered), and does not touch the local OSS API endpoints the reviewer asked about. The only gap is between PR description and actual semantics — a docs nit, not a correctness issue.

wcchang1115 (approved):

PR #1350, opened 2026-05-01
fix(cll): cross-env catalog fallback in per_node.db (drop manifest)

even-wei:
**PR checklist**

- [x] Tests added/updated (16 unit tests, +2 new for cross-env fallback)
- [x] DCO signed

**What type of PR is this?**

`fix` — follow-up to DRC-3349 addressing cross-env column casing mismatch.

When one env's catalog is partial (typical of `dbt docs generate --select state:modified+`), the per_node.db emitter previously fell back to schema.yml columns for nodes outside the selection. That introduced **lowercase columns** that disagreed with the **UPPERCASE catalog data** from the other env, producing false schema-change alarms in the side panel.

This PR aligns the emitter with OSS Recce semantics:

1. **Drop the manifest fallback.** OSS bulk lineage (`DbtAdapter.get_lineage_cached:767-781`) is catalog-only — when the catalog has no entry, the node's `columns` field is left unset, never filled from `manifest.columns`. This PR matches that.
2. **Add cross-env catalog fallback.** When this env's catalog has no entry for a node, try the *other* env's catalog before giving up. Closest available substitute for OSS's live `get_columns()` warehouse query (which Cloud can't perform in the precompute path) — for unchanged models the column list and types are identical between envs, and casing is always warehouse-truth.

Empirical re-emission against MR 5666 manifests (4,851 (node,env) groups, base catalog complete + current catalog partial):

| Metric | Before round-4 | After round-4 |
|---|---:|---:|
| Total columns | 99,716 | 134,388 |
| Lowercase columns | 35,386 | 8,510 (catalog quirks) |
| **Cross-env casing mismatches** | **26,000** | **0** |
| Nodes with primary_key | 1,126 | 1,126 |
| Tests joinable | 4,959/4,959 | 4,978/4,978 |

Column count rose because cross-env fallback fills in column data for unchanged models that previously had only lowercase manifest data — those nodes now get UPPERCASE catalog data from the other env.

Resolves DRC-3356
Refs DRC-3349 (per_node.db column list alignment, parent)

- 204 LOC total (28 prod + 176 test).
- Removes one test (`test_primary_key_falls_back_to_manifest_column_order_without_catalog`) — that behavior is gone by design.
- Renames `test_extract_rows_with_null_catalog` → `test_no_columns_emitted_without_catalog` and `test_manifest_fallback_when_catalog_has_no_entry_for_node` → `test_no_columns_when_neither_catalog_has_node`. Both tests' expectations change from "manifest fallback emits columns" to "no columns emitted" (matches OSS).
- Two new tests: `test_cross_env_catalog_fills_in_when_same_env_catalog_partial`, `test_same_env_catalog_wins_over_cross_env_catalog`.

**Trade-off (worth calling out)**:

Customers who never run `dbt docs generate` will see "no columns" for most nodes in the side panel. That matches OSS's behavior — Recce already requires catalog for column visibility, so this isn't a new constraint. It is, however, a stricter response than the previous Cloud-only manifest-fallback behavior, so existing customers who weren't running docs generate may notice.

\`\`\`release-note
Fixes false schema-change alarms in the Recce Cloud lineage side panel when one environment's catalog is partial. Column data now consistently uses warehouse-truth casing (catalog) across base and current envs. As a side effect, models without any catalog data ('dbt docs generate' was never run) will show "no columns" in the side panel — matching OSS Recce's behavior.
\`\`\`


even-wei:
# Code Review: PR #1350 (self-review)

**Reviewed at:** db489eff (HEAD of `fix/per-node-db-cross-env-catalog-fallback`)
**Files reviewed:** 3 (`recce/util/per_node_db.py`, `recce/cli.py`, `tests/test_per_node_db.py`)
**Categories:** Business logic, CLI wiring, Tests
**Passes run:** A (correctness), C (cross-references), D (error handling), E (test coverage), F (diff-specific), G (performance), H (async)

## Validation Results

### Pass A — Correctness & Logic — PASS (with notes)

- Cross-env fallback condition `if not column_source` (`recce/util/per_node_db.py:382`) correctly fires when same-env catalog has no entry. Tests cover the partial-catalog scenario (`test_cross_env_catalog_fills_in_when_same_env_catalog_partial`) and same-env precedence (`test_same_env_catalog_wins_over_cross_env_catalog`).
- Behavior matches OSS `DbtAdapter.get_lineage_cached` (catalog-only, no schema.yml fallback) — verified directly against `recce/adapter/dbt_adapter/__init__.py:765-789`.
- CLI tuple extension (`recce/cli.py:752-770`) correctly pairs `(env → other-env catalog)`: current ↔ base_catalog, base ↔ curr_catalog.
- The `has_target / has_base` comment at `recce/cli.py:744-751` is correct: when only one env is real, `dbt_adapter.{base,curr}_catalog` is path-swapped to the present env, so passing both to `extract_rows_from_artifacts` is redundant but a no-op (lookups against a same-keyed dict produce no fallback hits).

### Pass C — Cross-Reference Consistency — PASS

- Adding keyword-only `cross_env_catalog: Optional[dict] = None` is backwards compatible: all 6 existing call sites in `tests/test_cli_per_node_db.py` continue to compile and pass without modification.
- Tuple unpacking `for env_name, manifest, catalog, cross_catalog in envs_to_emit:` matches the 4-element tuples produced above.

### Pass D — Error Handling — PASS

- `_to_dict` (`recce/cli.py:772`) handles None and missing `to_dict` gracefully.
- `extract_rows_from_artifacts` handles `cross_env_catalog=None` correctly via `((cross_env_catalog or {}).get("nodes") or {})`.
- Existing top-level `try / except` in `init` catches per_node.db emitter failures without aborting CLL upload.

### Pass E — Test Coverage & Quality — PASS

- 16/16 tests pass in `test_per_node_db.py`; 11/11 in `test_cli_per_node_db.py` (5 fixture-skipped).
- New tests are specific (assert exact column lists with casing, with messages on failure).
- Renamed tests' new expectations match the new behavior (assertions inverted from "manifest fallback emits columns" to "no columns emitted").

### Pass F — Diff-Specific Checks — PASS

- Removed test (`test_primary_key_falls_back_to_manifest_column_order_without_catalog`) is correct — that behavior is gone by design.
- Renames preserve test intent and update assertions.
- All callers of `extract_rows_from_artifacts` continue to work: existing positional 3-arg calls remain valid since `cross_env_catalog` is keyword-only with default.
- No public-API rename or signature break.

### Pass G — Performance — PASS

- One additional dict lookup per node (`cross_env_catalog_nodes.get(node_id)`) only when same-env returns no columns. Existing 1918-node `test_extraction_under_2s` budget is unaffected.

### Pass H — Async/Concurrency — N/A

Pure data processing.

### Verification Results

- `uv run pytest tests/test_per_node_db.py` → 16 passed
- `uv run pytest tests/test_cli_per_node_db.py` → 11 passed, 5 skipped (fixture-dependent)
- `uv run black --check recce/util/per_node_db.py recce/cli.py tests/test_per_node_db.py` → clean
- `uv run flake8 recce/util/per_node_db.py recce/cli.py tests/test_per_node_db.py` → clean

## Verdict: GO

No BLOCKERs or ISSUEs. Five NOTEs follow — all are observations, none block merge.

### Notes

1. **Sources column data lost (pre-existing, amplified)** — `recce/util/per_node_db.py:361,362` only consults `catalog["nodes"]`, never `catalog["sources"]`. Source entries iterated via `_iter_manifest_nodes` (`_MANIFEST_NODE_SECTIONS = ("nodes", "sources", ...)`) will never find catalog data. Before this PR, the manifest fallback emitted lowercase source columns; after, sources show "no columns". OSS `get_lineage_cached:790-795` consults `catalog.sources` for sources — emitter still diverges from OSS for sources specifically. Not a regression introduced by this PR (catalog.sources was never consulted), but the manifest-fallback removal makes the side-panel UX worse for sources. Worth a follow-up to mirror OSS by also borrowing from `catalog["sources"]`.

2. **Empty same-env `columns: {}` triggers fallback** — `recce/util/per_node_db.py:382`: the condition `if not column_source` fires both when the catalog has no entry AND when the entry's `columns` dict is empty. The docstring (lines 340-347) describes only the former case. dbt's `catalog.json` derives `columns` from `information_schema`, which always has at least one row per materialized table, so empty-columns entries don't appear in practice. Theoretical only.

3. **`_to_dict` defined inside `init`** — `recce/cli.py:772-777`: closure redefined every CLI invocation. Functionally identical to a module-level helper; pure style.

4. **Redundant None guards** — `recce/util/per_node_db.py:361,362`: `((catalog or {}).get("nodes") or {}) if catalog else {}` — the trailing `if catalog else {}` is dead code because `(None or {}).get("nodes") or {}` already yields `{}`. Same on the cross-env line. Cosmetic.

5. **Cross-env path under-tested for PK / test normalization** — Tests cover (a) cross-env column emission and (b) same-env primary-key + test-name normalization. There is no test that exercises the combination: cross-env catalog supplying columns AND a unique test on those columns. Empirical PR data (4978/4978 tests joinable) suggests this works in practice; a unit test would lock the contract.

## What I could not verify

- End-to-end Cloud emit → side-panel render of the new behavior on a real Snowflake dataset. The empirical numbers in the PR description (4,851 (node,env) groups, 0 cross-env mismatches after) are taken on trust.
- Behavior under Postgres/BigQuery: identifier case folding rules differ. The PR's premise that "casing is warehouse-truth and env-independent" holds for Snowflake (UPPERCASE folding) but may need revisiting if Postgres folding lands lowercase across envs vs. upper-quoted ad-hoc.

- Type mismatches between `extract_rows_from_artifacts` callers and the new keyword param (none — all positional calls still valid).
- Stale references to the removed manifest-fallback path (none — checked CLI and both test files).
- Race conditions or concurrent access (N/A — single-process emitter).
- Performance regressions (none — added work is one dict.get per node).
- Security concerns: SQL injection / PII leak / secret exposure (N/A — no user input, no logging of catalog contents).

Generated by /recce-dev:claude-code-review (self-review).

gcko:
# Code Review: PR #1350

**Files reviewed:** 3 (`recce/util/per_node_db.py`, `recce/cli.py`, `tests/test_per_node_db.py`)
**Categories:** Business logic, tests
**Reviewed at SHA:** `db489eff`

## Validation Results

Traced the column-source decision tree against the OSS reference (`DbtAdapter.get_lineage_cached` at `recce/adapter/dbt_adapter/__init__.py:709-781`):

- **Same-env catalog precedence:** `column_source = catalog_cols` is unconditional; cross-env is only consulted when `not column_source`. Confirmed by `test_same_env_catalog_wins_over_cross_env_catalog`.
- **Cross-env fallback:** Only triggered when same-env catalog has no entry (or an empty `columns` dict). The casing argument holds — Snowflake/BigQuery identifier folding is per-warehouse, not per-env, so the other env's catalog is a faithful proxy.
- **Empty third tier:** With no manifest fallback, schema.yml lowercase columns can no longer leak in. The renamed tests' `columns == []` assertions prove the path is wired correctly.
- **`_pick_primary_key`:** Walks `ordered_cols` in catalog order; with no catalog there are no ordered cols → `primary_key = None`. `test_primary_key_is_none_without_catalog` confirms.
- **Test normalization:** `canonical_to_stored.get(t.column_name.lower())` — orphan tests (column names that don't exist in the chosen `column_source`) are dropped, not silently emitted with manifest casing.

- **Function signature:** `cross_env_catalog: Optional[dict] = None` is keyword-only (after `*`). Pre-existing positional callers in `tests/test_cli_per_node_db.py` (lines 818, 884, 1058, 1069) still work — verified by running them and they pass.
- **CLI tuple shape:** `envs_to_emit` tuples are now 4-tuples `(env_name, manifest, catalog, cross_catalog)`. Both append sites and the unpack at `cli.py:779` agree. `_to_dict` helper applied symmetrically to `catalog` and `cross_catalog`.
- **OSS parity claim:** `DbtAdapter.get_lineage_cached:767` is `if catalog is not None and unique_id in catalog.nodes` — catalog-only, no schema.yml fallback. PR's "drop manifest fallback" matches.

- `cross_env_catalog=None` handled at line 362 with the same `or {}` chain as `catalog`.
- `cross_node = cross_env_catalog_nodes.get(node_id)` returns `None` if the other env doesn't have the node → `(cross_node or {}).get("columns") or {}` yields `{}` → no rows emitted. No KeyError or AttributeError surface.
- **CLI path-swap interaction (`cli.py:527-532`):** When only one env has artifacts, `context_kwargs` aliases the missing target path to the present one, so `dbt_adapter.curr_catalog` and `dbt_adapter.base_catalog` end up pointing at the same content. Cross-env fallback then has no extra data to contribute (correctly degrades to no-op) — traced the case `has_target=True, has_base=False`.

Coverage on `recce/util/per_node_db.py`: **97%** (155 statements, 4 missed). Missed lines are defensive guards (`__exit__` short-circuit, RuntimeError when not opened, `not node_name` early return) and one dedup `continue` branch.

Test changes align with the behavior change:
- 2 new tests added (`test_cross_env_catalog_fills_in_when_same_env_catalog_partial`, `test_same_env_catalog_wins_over_cross_env_catalog`) — both exercise the new code path with realistic UPPERCASE catalog vs lowercase manifest data.
- 3 tests renamed + assertions inverted to match the new "no schema.yml fallback" contract. Names are now accurate.
- 1 test removed (`test_primary_key_falls_back_to_manifest_column_order_without_catalog`) — appropriate because the behavior is gone by design.

- **Behavior change documented:** the release note explicitly calls out the "no columns when no catalog" change for customers who don't run `dbt docs generate`.
- **No partial renames:** old test name strings don't appear elsewhere in the repo.
- **Public API backward compatible:** `extract_rows_from_artifacts` adds a kwarg with default `None`. `tests/test_cli_per_node_db.py` (untouched) continues to call it positionally.

Cross-env fallback adds at most one `dict.get()` and one `.get("columns")` per node when same-env catalog misses. Negligible vs the existing `_iter_manifest_nodes` cost. No new I/O, no nested loops.

The PR's empirical numbers (4,851 node-env groups → 134,388 columns at 0 mismatches, vs 99,716 at 26,000 mismatches) demonstrate the correctness improvement without a performance regression. The scale test (`test_extraction_under_2s`) still passes.

No async code or shared mutable state introduced.

### Verification Results

```
flake8: clean
make test: 1142 passed, 5 skipped (skips are scale fixtures unavailable locally)
pytest tests/test_per_node_db.py tests/test_cli_per_node_db.py: 27 passed, 5 skipped
coverage(recce/util/per_node_db.py): 97% (155 stmts, 4 missed)
```

## Verdict: GO

All validation passes clean. No blockers. No issues.

### Notes

1. **Minor stylistic inconsistency, `recce/cli.py:782-784`.** A new `_to_dict()` helper was introduced for `catalog` and `cross_catalog`, but the existing `manifest_dict = manifest.to_dict() if hasattr(manifest, "to_dict") else manifest` line was not refactored to use it. Cosmetic; leaves an obvious follow-up. Not blocking.

2. **Subtle divergence from OSS in an edge case, `recce/util/per_node_db.py:381-384`.** OSS `get_lineage_cached:767` enters the columns block on `unique_id in catalog.nodes` (membership), so a node that's in the catalog with an empty `columns` dict gets `nodes[unique_id]["columns"] = {}` (explicitly empty). The PR's gate is `if not column_source` (truthiness of the columns dict), so the same case falls through to cross-env. In practice catalog entries with empty `columns` are rare, and the PR's behavior is strictly more generous (more data, not less), so this is benign — but the docstring claim "mirroring OSS recce semantics" is slightly broader than the actual contract. Worth a one-line clarification if you revisit the comment.

3. **Untested branch: case-insensitive dedup at `per_node_db.py:394-395`.** The `seen_canonical` dedup logic guards against the same column appearing twice with different casings in the chosen `column_source`. No test in this PR (or the existing suite) exercises that branch — coverage shows line 395 unhit. The defensive value is real (Snowflake catalog quirks the comment mentions), but a regression introducing duplicate rows for the same canonical column would slip through. Not blocking; worth a one-case test when convenient.

4. **Pre-existing: sources never get catalog columns.** `_iter_manifest_nodes` yields source nodes, but `catalog_nodes = catalog["nodes"]` only — `catalog["sources"]` is never consulted (compare to OSS `get_lineage_cached:795-799` which does handle `catalog.sources`). This PR doesn't introduce or worsen that gap, and cross-env fallback also doesn't help sources. Out of scope here, but flagging in case it becomes the next ticket.

## What I could not verify

- The 4,851-group / 26,000→0 mismatch table from the PR description (would need MR 5666 manifests).
- Behavior on real BigQuery / Redshift catalogs (only Snowflake casing semantics were considered in detail).

- Mismatched tuple unpacking between `envs_to_emit` appends and the unpack site (3 vs 4 elements) — both sides agree.
- Mock drift in `tests/test_cli_per_node_db.py` after the keyword-only kwarg addition — positional-arg call sites still work.
- Performance regression on the scale test — under budget.
- Path-swap induced double-counting when only one env has artifacts — fallback is a no-op in that case.
- Lint or type errors — clean.

gcko (approved):

Issue #1354, opened 2026-05-05
[Bug] Column lineage error 500


anentropic:
### Current Behavior

I ran `dbt docs generate` and `dbt docs generate --target-path ./target-base` so that both `catalog.json` are updated

I ran `recce server --single-env` clicked on a table, then "Columns" and clicked on a column:

<img width="312" height="96" alt="Image" src="https://github.com/user-attachments/assets/76f84692-14c5-48b1-8b23-bfa02a098184" />

### Expected Behavior

No error, column lineage displayed

### Steps To Reproduce

I ran `recce server --single-env` clicked on a table, then "Columns" and clicked on a column

### Relevant log output

```shell
File "myproject/dbt/.venv/lib/python3.11/site-packages/fastapi/routing.py", line 355, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "myproject/dbt/.venv/lib/python3.11/site-packages/fastapi/routing.py", line 243, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "myproject/dbt/.venv/lib/python3.11/site-packages/recce/server.py", line 651, in column_level_lineage_by_node
    cll = dbt_adapter.get_cll(
          ^^^^^^^^^^^^^^^^^^^^
  File "myproject/dbt/.venv/lib/python3.11/site-packages/recce/adapter/dbt_adapter/__init__.py", line 1317, in get_cll
    cll_data_one = deepcopy(self.get_cll_cached(cll_node_id, base=False))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "myproject/dbt/.venv/lib/python3.11/site-packages/recce/adapter/dbt_adapter/__init__.py", line 1706, in get_cll_cached
    parent_key = f"{table_id_map[d.node.lower()]}_{d.column}"
                    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'connect_appointment_totals'
```

### Additional Context

In this project we have a dbt `generate_alias_name` macro which prepends `vw_` prefix to models in `intermediate/` path, which are configured to materialise as view, in order to match legacy naming conventions in our warehouse.

I added some debugging as re-ran:

```python
1706 	            try:
1707 	                parent_key = f"{table_id_map[d.node.lower()]}_{d.column}"
[...]

PR #1356, opened 2026-05-05
perf(cll): cache negative results in compiled-table resolver

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

perf + test

Follow-up to #1355 picking up two review notes that landed against the second commit there but didn't make it in before merge:

1. **Copilot — negative-result memoization** ([#1355 review on `__init__.py:1589`](https://github.com/DataRecce/recce/pull/1355#discussion_r3189458562)). The first pass memoized successful resolutions into `table_id_map` but left misses (unresolved or ambiguous names) re-scanning `manifest.nodes` / `manifest.sources` on every call. For nodes whose compiled SQL repeats the same unresolved reference across many column dependencies, that's O(N*M) on the cold path.

   This PR caches misses too. The helper signature becomes `Dict[str, Optional[str]]` and stores `None` for both unknown and ambiguous names. Repeats short-circuit on the existing `if key in table_id_map: return ...` fast path. Both paths now have the same one-shot cost.

2. **even-wei — caplog assertion on the warning** ([#1355 review note 3](https://github.com/DataRecce/recce/pull/1355#issuecomment-4380456593-pending)). The deduped warning was added in #1355 but no test asserted it fires. Adds a `caplog` block to `test_cll_with_compiled_code_unknown_parent_does_not_500` verifying:

   - exactly one warning per node (dedup invariant)
   - the warning names the affected node and the unresolved parent (debuggability)

Unit tests for `_resolve_compiled_table` are extended to verify the new negative-cache invariant on both the unknown-name and ambiguous-alias paths.

Follow-up to #1354 / #1355. No new issue.

- Smallest possible diff: 23 source lines, 34 test lines. Changes only the helper added in #1355 — no new public surface, no behavior change other than the negative-cache short-circuit.
- The `Dict[str, Optional[str]]` type widening is local. The Jinja path's writes to `table_id_map` (`__init__.py:1626`, `:1668`, `:1683` post-#1355) only ever store non-None unique_ids, so they're forward-compatible.
- Confirmed against `tests/adapter` (56 passed) and the new `test_cll_with_compiled_code_unknown_parent_does_not_500` caplog block reproduces the dedup invariant: a model with two unresolved references emits exactly one warning containing both names sorted.

```release-note
NONE
```

---
_Generated by [Claude Code](https://claude.ai/code/session_01SFLmgJ5YXeGTL2ULVM1vQT)_


kentwelcome (approved):
LGTM

PR #1358, opened 2026-05-07
fix(mcp): accept session_id env var for cloud MCP validation

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`/kind bug`

DRC-3345 added a new `--session` flag for cloud MCP mode and strict validation that errors when `--cloud` is set without `--session`. The legacy AI-summary entrypoint at `recce_instance_launcher/src/recce_ai_summary_entrypoint` invokes `recce mcp-server --sse` with `RECCE_SESSION_ID` set in the environment. Click binds that env var to `--session-id`, and `patch_derived_args()` then promotes `session_id` to `cloud=True`. With 1.48.0.dev0, this implicit-cloud path no longer satisfied the new check (`cloud_session` only reads from the explicit `--session` flag), so the CLI exited before MCP could start.

Effect in production: every AI-summary container started, `recce mcp-server` exited with `[Error] --session is required when using --cloud with recce mcp-server.`, the entrypoint timed out after 30 health-check attempts, and the GitLab/GitHub PR comment stuck on "Generating summary... Please wait." across all sessions for ~17 hours starting `2026-05-06 09:33 UTC` (right after 1.48.0.dev0 published and the launcher image rebuilt to pull it).

The fix: fall back to `session_id` when `cloud_session` is unset, so the new validation honors the same source of truth that `patch_derived_args()` already uses. Genuine misuse (`--cloud` with no session source at all) still errors out — the existing `test_cmd_mcp_server_cloud_requires_session` confirms.

```diff
- cloud_session = kwargs.pop("cloud_session", None)
+ cloud_session = kwargs.pop("cloud_session", None) or kwargs.get("session_id")
```

Resolves DRC-3383

- Added `test_cmd_mcp_server_session_id_env_var_satisfies_cloud_session` as a regression test that reproduces the production failure (RED before fix, GREEN after).
- All 176 CLI/MCP tests pass; full suite 1185/1190 (5 pre-existing unrelated static-route failures need frontend build).
- After merge and 1.48.0.dev1 publishes to PyPI, the `recce-instance-launcher` image will auto-rebuild on `repository_dispatch` and pick up the fix. No recce-cloud-infra change required.

\`\`\`release-note
Fix `recce mcp-server` exiting with "--session is required" when only the `RECCE_SESSION_ID` env var is provided. The validation now accepts either the explicit `--session` flag or the `RECCE_SESSION_ID`/`RECCE_SNAPSHOT_ID` env var (matching the existing `--session-id` behavior).
\`\`\`

### DIFF
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -2561,7 +2561,9 @@ def mcp_server(state_file, sse, host, port, **kwargs):
     handle_debug_flag(**kwargs)
     patch_derived_args(kwargs)
     is_cloud_mcp = kwargs.get("cloud", False)
-    cloud_session = kwargs.pop("cloud_session", None)
+    # DRC-3383: also accept the session_id env var (RECCE_SESSION_ID) — patch_derived_args()
+    # already promotes session_id to cloud=True, so the validation must honor the same source.
+    cloud_session = kwargs.pop("cloud_session", None) or kwargs.get("session_id")
 
     if is_cloud_mcp and not cloud_session:
         console.print("[[red]Error[/red]] --session is required when using --cloud with recce mcp-server.")
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -345,6 +345,32 @@ class TestCommandMCPServer(TestCase):
         assert "single_env" not in call_kwargs
         assert "Base artifacts not found" not in result.output
 
+    @patch("asyncio.run")
+    @patch("recce.mcp_server.run_mcp_server", new_callable=MagicMock)
+    @patch("recce.config.RecceConfig")
+    @patch("recce.util.api_token.prepare_api_token", return_value="token-abc")
+    def test_cmd_mcp_server_session_id_env_var_satisfies_cloud_session(
+        self, mock_prepare_api_token, mock_recce_config, mock_run_mcp_server, mock_asyncio_run
+    ):
+        # DRC-3383 regression: the AI-summary entrypoint invokes `recce mcp-server --sse`
+        # with RECCE_SESSION_ID set in the environment. patch_derived_args() promotes
+        # session_id to cloud=True, so the new --session validation must accept the
+        # session_id env var as the session source — otherwise the CLI exits before
+        # the MCP server starts.
+        with patch.object(Path, "is_dir", return_value=False):
+            result = self.runner.invoke(
+                cli_command_mcp_server,
+                ["--sse"],
+                env={"RECCE_SESSION_ID": "sess-from-env"},
+            )
+
+        assert result.exit_code == 0, result.output
+        assert "--session is required" not in result.output
+        mock_run_mcp_server.assert_called_once()
+        call_kwargs = mock_run_mcp_server.call_args.kwargs
+        assert call_kwargs["cloud"] is True
+        assert call_kwargs["session"] == "sess-from-env"
+
 
 def test_cli_shows_update_available_warning():
     """CLI should show update available warning when a newer version exists."""


even-wei:
# Code Review: PR #1358

**Commit reviewed:** `f882905b`
**Branch:** `feature/drc-3383-recce-gitlab-summary-generation-gets-stuck-waiting-for-mcp`
**Files reviewed:** 2 (`recce/cli.py`, `tests/test_cli.py`)
**Categories:** Business logic / CLI, Tests

## Validation Results

Traced every scenario through `mcp_server` validation after the fix:

| Invocation | `session_id` (envvar) | `cloud_session` (`--session`) | `is_cloud_mcp` | Effective `cloud_session` | Result |
|---|---|---|---|---|---|
| `--cloud --session foo` | – | `"foo"` | True | `"foo"` | starts |
| `--cloud` (no session) | – | None | True | None | errors `--session is required` |
| `--session foo` (no `--cloud`) | – | `"foo"` | False | `"foo"` | errors `--cloud is required` |
| `RECCE_SESSION_ID=x` only | `"x"` | None | True (via `patch_derived_args`) | `"x"` | starts (was broken pre-fix) |
| `--session foo` + `RECCE_SESSION_ID=x` | `"x"` | `"foo"` | True | `"foo"` (explicit wins) | starts |
| no flags, no env | None | None | False | None | starts in local mode |

`patch_derived_args` (cli.py:38–44) sets `cloud=True` whenever `session_id` or `share_url` is truthy, which is exactly what the new fallback aligns with. The `or kwargs.get("session_id")` correctly defers to `--session` when both sources exist (short-circuit), and `kwargs.get` (vs `pop`) preserves the existing pattern of `session_id` flowing through to `run_mcp_server` via `**kwargs` — no new aliasing or shadowing introduced.

- `--session-id` option lives in `recce_hidden_options` (cli.py:233–239) with `envvar=["RECCE_SESSION_ID", "RECCE_SNAPSHOT_ID"]`. `mcp_server` registers `recce_hidden_options` (cli.py:2490), so `kwargs["session_id"]` is guaranteed to exist (None if unset).
- `--session` option (cli.py:2483) maps to `cloud_session` — `kwargs.pop("cloud_session", None)` matches.
- `run_mcp_server` signature (mcp_server.py:2478) accepts `session: Optional[str] = None`; the call at cli.py:2622 passes `session=cloud_session`. Type/keyword aligned.
- `cloud=True` flows into `kwargs` from `patch_derived_args`, then into `run_mcp_server(**kwargs)`, binding to `cloud: bool = False`. The downstream defense at mcp_server.py:2535–2537 (`if cloud: ... if not session: raise ValueError`) stays consistent.

- Both error branches still fire correctly (verified in the Pass A table). The `--cloud is required when using --session` branch (cli.py:2571) is unreachable for the env-var path because `patch_derived_args` already promotes to `cloud=True` — that is intended behavior, not a regression.
- Empty-string edge: `RECCE_SESSION_ID=""` → `session_id=""` (falsy) → no `cloud=True` promotion → falls through to local mode. Acceptable; matches pre-fix behavior.
- No new exceptions, no resource lifecycle changes.

- **Verified red-before-green**: temporarily reverted cli.py to the pre-fix one-liner and re-ran the new test. It failed with `assert 1 == 0` and the exact production error message:
  > `[Error] --session is required when using --cloud with recce mcp-server.`

  After restoring the fix, all 5 `TestCommandMCPServer` tests pass (4.89s).
- Mock targets correct: `recce.mcp_server.run_mcp_server`, `recce.config.RecceConfig`, `recce.util.api_token.prepare_api_token` are all imported inside `mcp_server()`, so module-level patches resolve at call time. Same pattern as the four sibling tests.
- Assertions are specific (`cloud is True`, `session == "sess-from-env"`, exit_code 0, absence of the error string). Not vacuous.
- Test exercises `--sse`, which matches the production failure mode (the AI-summary entrypoint uses `--sse`).

- Single-line behavioral change with WHY-comment citing DRC-3383 and pointing to `patch_derived_args` as the source-of-truth justification.
- No public API change, no migration, no flag default flip, no breaking behavior for existing invocations (verified all 6 input shapes in Pass A).
- No stale references — `cloud_session`/`session_id` naming unchanged.

### Pass G: Performance — N/A

No hot-path or query changes.

No async surface change.

### Verification Results

| Check | Result |
|---|---|
| `flake8 recce/cli.py tests/test_cli.py` | exit 0 |
| `black --check recce/cli.py tests/test_cli.py` | "All done! 2 files would be left unchanged" |
| `isort --check-only recce/cli.py tests/test_cli.py` | exit 0 |
| `pytest tests/test_cli.py::TestCommandMCPServer` | 5 passed in 4.89s |
| `pytest tests/test_cli.py` (full module) | 19 passed; 1 unrelated failure in `test_cmd_version` (env-only — `dbt-core` not installed in my local sandbox; not affected by this PR) |

## Verdict: GO

The fix is minimal, targeted, and correct. The new regression test reproduces the production failure (verified red-before-green), the existing `test_cmd_mcp_server_cloud_requires_session` still passes (genuine misuse without any session source still errors), and no other validation branch changes behavior. PR description, code change, and test coverage are all consistent with the actual production trace.

### Notes

1. `kwargs.get("session_id")` (vs `pop`) leaves `session_id` in `kwargs` when it gets splatted to `run_mcp_server`, where it lands in `**kwargs`. In cloud mode this is harmless (`run_mcp_server` only consumes `session`), and in local mode `session_id` was already being passed through pre-fix — so this is consistent with prior behavior, not a finding.

## What I could not verify

- Behavior in the actual `recce_instance_launcher` container against the published 1.48.0.dev1 build — that is the post-merge validation step the PR description already calls out.
- Whether the GitHub Actions CI matrix installs the `mcp` extra. Locally the test class is gated by `@pytest.mark.skipif(... "mcp package not installed")`, and I had to `pip install 'recce[mcp]'` to unblock it. If CI does not install the extra, the new regression test will silently SKIP and the bug could re-regress unnoticed.

- Logic inversion or short-circuit bug in the `or` fallback — none; explicit `--session` correctly wins over env var.
- Type/signature drift between `cli.py` call and `run_mcp_server` definition — none.
- Mock target mismatch with the actual import paths — none.
- Regression in the second validation branch (`--cloud is required when using --session`) — not reachable for the env-var path, but that is correct since `patch_derived_args` promotes to `cloud=True` first.
- Hidden side effects from `kwargs.pop` vs `kwargs.get` change in flow — verified `session_id` was already in kwargs pre-fix, so no new aliasing.
- PII/secret leak in error path — none; the error message contains no user data.

kentwelcome (approved):
LGTM

PR #1375, opened 2026-05-14
fix(breaking): surface unknown when CTE-internal change can't be resolved

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` — UI consistency bug.

The Columns tab in the model panel and the column-level lineage (CLL) view used two independent analyzers with asymmetric failure modes. When `parse_change_category` ran without a parent schema (e.g., catalog shipped with missing source columns from a Paradime CI race), sqlglot's `qualify` was skipped, so column refs through CTEs lost their table qualifiers and `source_column_change_status` couldn't trace inner-scope changes back to outer projections. The analyzer silently returned `change.columns = {}` and `category = "non_breaking"`, while the CLL view loudly returned `transformation_type = "unknown"` for every column. Result: the Columns tab showed no marker, but clicking the column opened a "unknown" CLL view — the two views disagreed.

This PR detects that unresolvable state inside `_diff_select_scope`: when a selected source has changes but the change didn't propagate to any outer projection, every outer column is marked `"unknown"` and the node category becomes `"unknown"`. Frontend renders a new `?` badge in the Columns tab so it matches the CLL view's loud signal.

Reproduction lives at `workspace/jaffle-shop-expand-prs/pr-19-cte-case-when-degraded{,-base}/`. Before the fix:

| Surface | category | columns |
|---|---|---|
| Path A (Columns tab) | `non_breaking` | `{}` |
| Path B (CLL) | n/a | every column `transformation_type=unknown` |

After the fix:

| Surface | category | columns |
|---|---|---|
| Path A | `unknown` | every outer column `"unknown"` |
| Path B | n/a | every column `transformation_type=unknown` |

Both surfaces now agree.

Resolves DRC-3409.

- Backend change is localized to `_diff_select_scope` (one new helper + one branch before return). The guard fires only when an inner CTE source has a `modified` column that never gets traced back to an outer projection, OR when a breaking source has no column attribution. `change_category not in ("breaking",)` preserves breaking as the higher signal.
- `_unresolvable_cte_change` walks `selected_sources` (not `sources`) — unused CTEs in the WITH clause are inert. Verified against the existing `test_no_schema` regression case (uses an unused `renamed` CTE).
- Frontend: new `changeUnknown` row flag mirrors `definitionChanged`; new `?` badge with `--schema-badge-unknown-bg/fg` tokens (muted neutral); existing impacted-badge precedence respected.
- Out of scope: `_diff_union_scope` could also silently drop a child "unknown" — left for a follow-up since unions over CTEs with unresolvable inner changes are not the reported customer pattern.

**Test plan**:
- [x] 52 backend tests in `tests/util/test_breaking.py` pass
- [x] 3719 frontend tests pass (`pnpm vitest run`)
- [x] End-to-end repro against the degraded-catalog fixture: pre-fix server returns `non_breaking` + empty `columns` (the bug); post-fix server returns `unknown` + all 9 columns `unknown` (the fix)

Yes.

```release-note
The Columns tab on the model panel now shows a "?" badge for columns whose change status couldn't be resolved (e.g., when catalog.json is missing source column info), matching the column-level lineage view's "unknown" state. Previously these columns appeared unchanged.
```

Generated with assistance from Claude Code.

---

### Reproduction

Captured against a degraded-catalog fixture variant (`pr-19-cte-case-when-degraded` + `pr-19-cte-case-when-degraded-base`) that reproduces the Paradime CI-race condition observed in the field. The original `pr-19-cte-case-when` fixture ships a full catalog and doesn't trigger the fallback path. Degradation: `target/catalog.json` has `source.jaffle_shop.ecom.raw_orders` removed from `catalog.sources`, leaving empty column metadata for that parent. With no source-column schema, sqlglot's `qualify()` cannot attach table qualifiers to inner CTE column refs, and `parse_change_category` falls into the new fallback.

- **Pre-fix worktree**: `recce-pr1375-pre-fix` at main SHA `2fc9ef2d` (version `1.50.0.dev0`).
- **Post-fix worktree**: this PR branch at HEAD `21319f24` (version `1.49.0.dev0`).
- Both servers launched with `--target duckdb --new-cll-experience` against the degraded fixtures on ports 28083 (pre-fix) and 28084 (post-fix). Captured with headless Chromium via standalone Playwright (1440×1800 @ 2x DPR).

### Direct API evidence (`POST /api/cll` for `stg_orders_with_year`, `change_analysis=true`)

| Field | Pre-fix (port 28083) | Post-fix (port 28084) |
|---|---|---|
| node `change_status` | `modified` | `modified` |
| node `change_category` | `non_breaking` (the bug) | `unknown` (fixed) |
| `columns` size | 0 (silently empty) | 9 |
| `columns.location_year.change_status` | (not present) | `unknown` |

### Screenshots

| # | Surface | What it shows | File |
|---|---|---|---|
| SS-1 | Pre-fix Columns tab | Empty column list, canvas labels model **Non Breaking** — the bug | [SS-1-before-columns-tab.png](docs/screenshots/drc-3409/SS-1-before-columns-tab.png) |
| SS-2 | Pre-fix CLL view (sidebar Lineage) | Sidebar marks model **Modified** while canvas says **Non Breaking** — surfaces disagree | [SS-2-before-cll-view.png](docs/screenshots/drc-3409/SS-2-before-cll-view.png) |
| SS-3 | Post-fix Columns tab | Every column row shows the new `?` badge (muted neutral); canvas labels model **Unknown** | [SS-3-after-columns-tab.png](docs/screenshots/drc-3409/SS-3-after-columns-tab.png) |
| SS-4 | Post-fix CLL view (sidebar Lineage) | Canvas labels model **Unknown**; sidebar marks Modified; surfaces now agree (Modified-but-Unknown is the intended end state) | [SS-4-after-cll-view.png](docs/screenshots/drc-3409/SS-4-after-cll-view.png) |
| SS-5 | Post-fix paired agreement | Columns tab `?` badges + canvas `Unknown` label visible together — no contradiction | [SS-5-after-paired-agreement.png](docs/screenshots/drc-3409/SS-5-after-paired-agreement.png) |

_Screenshot files removed from the branch (`docs/screenshots/drc-3409/`) post-approval, per repo-hygiene policy. The table above documents what each capture verified; the actual PNGs were Playwright-captured against the degraded-catalog fixture and are archived locally if anyone wants to re-see a specific frame._


even-wei:
# Code Review: PR #1375

**SHA** `772b6bd9` · **Verdict** NO-GO

## Issues
1. `recce/util/breaking.py:256` — Guard requires `not changed_columns`, so the loud-fail never fires when the outer SELECT picks up *any* column change while an inner CTE simultaneously has an unresolvable inner edit. Result: a mixed-edit PR (one direct outer-level rewrite plus one CTE-internal edit on a different column) still silently reports only the outer-detected column, swallowing the inner-unresolvable one. This is the exact false-quiet shape the PR aims to close — narrowed to "all-or-nothing" cases.
   Evidence: `if change_category not in ("breaking",) and not changed_columns and _unresolvable_cte_change():`. Drop `not changed_columns` and union `{col: "unknown"}` over outer columns whose status is not already set; or document the narrow scope explicitly and add a regression test for the mixed case.
   Pass A / F.

2. `js/packages/ui/src/components/summary/utils.ts:135-139` — `calculateChangeSummary` enumerates only `added | removed | modified`; the new `"unknown"` column status is silently skipped, so the dashboard column-change counter understates impact for any model that now shows a `?` badge. The Columns tab matches CLL, but the summary counters do not — restoring an asymmetry on a different surface.
   Evidence: `for (const status of Object.values(columnChanges)) { if (status === "added") col_added++; else if (status === "removed") col_removed++; else if (status === "modified") col_changed++; }` — no branch for `"unknown"`. Either add a new `col_unknown` bucket or fold unknown into `col_changed` with a comment.
   Pass C.

3. `js/packages/storybook/stories/schema/fixtures.tsx:35-41` — Storybook fixture's `isRowChanged` (which drives badge selection in stories) only recognizes `definitionChanged`, not `changeUnknown`. The storybook surface — used for visual regression and design review of this exact UI — will not show the new `?` badge styling against the same row-changed background that the production grid uses. Add `changeUnknown === true` to the OR.
   Evidence: `return row.baseType !== row.currentType || row.reordered === true || row.definitionChanged === true;` — `changeUnknown` flag is absent.
   Pass C.

## Notes
- `recce/util/breaking.py:120` — `_unresolvable_cte_change` iterates `selected_sources` (not `sources`), confirmed correct against sqlglot semantics: unused CTEs are filtered out.
- `recce/util/breaking.py:126` — `src_change.category in ("breaking", "partial_breaking", "unknown")` includes `"unknown"`, but the dispatcher at `parse_change_category:350` early-returns on inner-scope `unknown`, so the `"unknown"` arm here is unreachable in practice. Not a bug; the dead arm reads as if cascading were supported.
- `recce/models/types.py:139` — `ChangeStatus` Literal widened to include `"unknown"`, but `NodeDiff.change_status` (also typed `ChangeStatus`) is never set to `"unknown"` by any code path. The comment at `NodeDiff` ("Only available if change_status is 'modified'") is now slightly stale relative to the broader union.
- Backend tests (50/50) and frontend tests (88/88 for ColumnNameCell + toSchemaDataGrid) pass on HEAD; `make flake8` clean; `pnpm type:check` clean.

even-wei:
# Code Review: PR #1375 (Re-Review)

**SHA** `d892c063` · **Verdict** GO · **Incremental** (vs `772b6bd9`)

## Prior findings — disposition

| Prior | Status | Notes |
|-------|--------|-------|
| ISSUE 1: `breaking.py:256` mixed-edit false-quiet | **Closed** | Guard rewritten to per-column stamping; `traced_source_columns` correctly distinguishes "saw this source column" from "modeled the change". Regression test `test_mixed_outer_change_and_unresolvable_cte_change` locks the bug shape; `test_outer_modified_column_keeps_partial_with_unresolvable_cte` confirms `partial_breaking` is preserved alongside per-column `unknown`. |
| ISSUE 2: `summary/utils.ts:135` unknown counter | **Closed** | `status === "modified" \|\| status === "unknown"` lands in `col_changed` as the prior review recommended; 3 new vitest cases pin the rollup including the mixed case. |
| ISSUE 3: `fixtures.tsx:35` storybook drift | **Closed** | `changeUnknown === true` added to `isRowChanged`; matches `SchemaView.tsx:359` production rendering exactly. |
| NOTE: `selected_sources` vs `sources` | Closed | Still correct — unused CTEs filtered. |
| NOTE: dead `"unknown"` arm in cascade check | Closed | Removed; replaced with sharper `category == "breaking" and not src_change.columns` signal. |
| NOTE: `NodeDiff.change_status` Literal staleness | **Still flag** (NOTE) | Out of scope for this PR; tracker-worthy. |

## Adversarial probes (8 scenarios run; all behaved correctly)

- Nested CTE (3 layers) with inner unresolvable modification → `unknown` cascades up correctly via the per-scope dispatcher.
- Multiple unresolvable columns in one CTE → all outer columns stamped `unknown`.
- CTE column *removed* (not modified) → no false-loud; `partial_breaking` + `w=removed` (status filter `== "modified"` on line 145 is the right gate; added/removed surface via outer projection list directly, as the comment claims).
- Recursive CTE with internal modification → correctly surfaces `unknown`.
- WHERE clause referencing unresolvable CTE column → stays `unknown` (does not falsely escalate to `breaking`).
- Pure `non_breaking` (added-only) outer + CTE that only adds (not modifies) a column → stays `non_breaking`. No spurious escalation. The `== "modified"` gate prevents added/removed CTE cols from tripping the loud-fail unnecessarily.

## Notes
- `breaking.py:281` — escalation `non_breaking → unknown` only fires when the outer projection list is genuinely unmodified; any `partial_breaking` or `breaking` outer signal is preserved. The North Star ("false-loud over false-quiet") is honored without manufacturing false-loud for routine non-modifying CTE edits.
- `breaking.py:140` — new `category == "breaking" and not src_change.columns` arm subsumes the prior dead `"unknown"` cascade; comment is honest about it being a category-level signal with no traceable attribution.
- `recce/models/types.py:139` — `NodeDiff.change_status` Literal includes `"unknown"` but no code path sets it; not blocking, but worth a follow-up to either narrow the type or wire a producer.

## Limits
- No runtime check against a real dbt project; verification is unit-test + adversarial SQL probe only. Backend (52 breaking) + frontend (3 new summary utils) all green on `d892c063`.

wcchang1115:
# Code Review: PR #1375 (Incremental)

**SHA** `c7b031ed` · **Verdict** NO-GO · **Incremental** (vs `37b2b434`)

## Prior findings — disposition

| Prior | Status | Notes |
|-------|--------|-------|
| Issue 1: `info.ts:84` `MergedNodeData.change.columns` missing `"unknown"` | **Closed** | Widened in `ac544cd1` |
| Issue 2: `info.ts:10` `NodeColumnData.change_status` missing `"unknown"` | **Closed** | Widened in `ac544cd1` |
| Issue 3: `cll.ts:37` `CllNodeData.change_status` missing `"unknown"` | **Closed** | Widened in `ac544cd1` |
| Sibling: `info.ts:81` `MergedNodeData.change_status` | **Closed** | Widened in `ac544cd1` |
| Note: `breaking.py:140` branch (C) reachability | **Deferred** | Explicitly tracked in ac544cd1's commit body; pin test from 37b2b434 stays |
| Limit: prior reviewer could not run frontend tests locally | **Resolved** | 3756 frontend tests, 53 backend `test_breaking` tests, lint, type-check all green at HEAD |

## Issues

1. `js/packages/ui/src/components/lineage/columns/LineageColumnNode.tsx:171` — `symbols: Record<ColumnChangeStatus, string>` and `cllChangeStatusColors`/`cllChangeStatusBackgrounds*` have no `"unknown"` key, so the new `"unknown"` runtime value (now flowing through the widened wire types) renders a column row with no tinted background, no left-accent border, and an empty 14×14 indicator circle — visually identical to "no change", i.e. the same symptom DRC-3409 was filed about, just on the CLL column-lineage surface instead of the schema column tab.
   Evidence: `NodeColumnData.change_status` is widened at `info.ts:10` ✓, but the next layer is still narrow — `ColumnAnnotation.changeStatus` at `computeColumnLineage.ts:7` (`"added" | "removed" | "modified"`) and `LineageGraphColumnNode.data.changeStatus` at `contexts/lineage/types.ts:62`. The `as ColumnAnnotation["changeStatus"]` casts at `computeColumnLineage.ts:44,68` and the `as LineageGraphColumnNode` cast at `lineage.ts:297` silently carry `"unknown"` through at runtime; the renderer then evaluates `palette["unknown"]` (line 170 / lines 315-322) and `symbols["unknown"]` (line 193) → `undefined` everywhere. ac544cd1's wire-widening leaves the same anti-pattern the prior review flagged — silent narrowing via `as` casts — one layer deeper.
   Pass F.

## Notes

- `ac544cd1`'s commit message says the cast at `computeColumnLineage.ts:44` is "papered over" — that paper-over was acknowledged in the prior review (Finding 2's evidence). The wire-format widening alone doesn't unpaper it; the symbol/color tables in `LineageColumnNode.tsx:171` and `styles.tsx:687-696` need entries for `"unknown"` (or upstream needs to coerce `"unknown"` → `"modified"` before reaching the CLL renderer) for the indicator on a CTE-affected column to render in CLL view.
- `d7d238ae`/`c7b031ed` net effect on `tests/util/test_breaking.py:1508`: the docstring still names "Paradime" in a public repo, framed as "a Paradime CI race condition." Partner-relations call, not a code concern — flagging only because the round-trip explicitly considered and re-introduced the name.
- Type-check, lint, and the full frontend + breaking-test suites pass at `c7b031ed`. The wire-format widening for `MergedNodeData.change_status` and `CllNodeData.change_status` (`info.ts:81`, `cll.ts:37`) is defensive — current backend writers only set node-level `change_status` to added/removed/modified — but matches the backend `ChangeStatus` Literal, so the alignment is correct.

## Limits

- The renderer bug under Issue 1 is reasoned from a code trace, not from a runtime repro on the CLL view with a CTE-affected column. The chain is mechanical (cast → narrow store → narrow map lookup → `undefined`), but a screenshot of CLL view on the `205DataLab` fixture would be the cleanest verification.

wcchang1115 (changes requested):
Hi Even,
Backend looks solid; please add a test covering the `category == "breaking" and not src_change.columns` branch in `_unresolvable_cte_change` (currently unreachable by any existing test).

Frontend LGTM.

Thanks!

even-wei:
Thanks @wcchang1115. Pushed `37b2b434` with a pin test for the DISTINCT-in-CTE scenario, but I wanted to flag something I hit while writing it: **branch (C) is currently unreachable** under any input I could construct.

### Why

In `_diff_select_scope` the upstream check at `recce/util/breaking.py:75-79`:

```python
for source_name, source in new_scope.sources.items():
    if scope_changes_map.get(source) is not None:
        change = scope_changes_map[source]
        if change.category == "breaking":
            change_category = "breaking"
```

…always sets the outer `change_category = "breaking"` whenever a source CTE is itself "breaking". The loud-fail guard at line 277 (`change_category != "breaking" and _unresolvable_cte_change()`) then short-circuits before branch (C) ever runs.

I empirically confirmed three natural shapes for `(category="breaking", columns={})` at the CTE level — DISTINCT added, WHERE introduced, sub-CTE turning breaking — all of them route through the upstream check first and never reach branch (C):

```
DISTINCT-in-CTE:       category='breaking' columns={}
WHERE-added-in-CTE:    category='breaking' columns={}
DISTINCT-in-sub-CTE:   category='breaking' columns={}
```

### What the new test pins

`test_breaking_cte_without_column_attribution` asserts exactly that current behavior — `category="breaking", columns={}` — with a docstring pointing at the gap. So the test is a **behavior pin**, not a coverage of branch (C). A future fix that lets branch (C) fire will visibly flip the assertion (caught at PR review).

### Proposal

I'd like to handle the reachability fix as a follow-up rather than expand this PR's scope:

- **Option A (recommended, smallest):** Relax the guard at line 277 from `change_category != "breaking"` to always tag unresolved columns (the `if column_name not in changed_columns` filter already protects already-classified columns). This makes branch (C) fire even when the outer is already "breaking", surfacing per-column "unknown" instead of empty columns. Strictly a UX improvement — current breaking-with-empty-columns is the same shape as the original DRC-3409 bug (Columns tab silent on a breaking model).
- **Option B:** Delete branch (C) as dead code.

I lean A — happy to file a follow-up issue if you agree, and we can land that fix + the test asserting the desired "unknown" surface in one shot. WDYT?

The two lower-priority gaps you flagged (unused-CTE inertness, added/removed-only sources) — happy to add those as separate pin tests in this PR or fold them into the follow-up, whichever you prefer.

wcchang1115 (changes requested):
# Code Review: PR #1375

**SHA** `37b2b434` · **Verdict** NO-GO

> Updated after a second-opinion pass: withdrew the original Finding 3 (`patchLineageDiffFromCll.ts` locals) as a consequence of Finding 2 rather than a separate issue, and added a missed sibling type-drift at `cll.ts:37`.

## Issues
1. `js/packages/ui/src/api/info.ts:84` — `MergedNodeData.change.columns?: Record<string, "added" | "removed" | "modified">` doesn't include `"unknown"`, but the server now emits `"unknown"` values here via the new `_diff_select_scope` fallback.
   Evidence: backend `recce/models/types.py:135-140` widens `ChangeStatus` to include `"unknown"`; `recce/util/breaking.py:277-280` writes `changed_columns[column_name] = "unknown"`. `MergedNode` uses `NodeChange` directly with no pre-serialization filter, so the value reaches the wire.
   Pass C.

2. `js/packages/ui/src/api/info.ts:10` — `NodeColumnData.change_status?: "added" | "removed" | "modified"` doesn't include `"unknown"`, but `/api/cll` now serializes `CllColumn.change_status = "unknown"` whenever the upstream `change.columns` value is `"unknown"`.
   Evidence: `recce/adapter/dbt_adapter/__init__.py:1205-1208` sets `col_obj.change_status = column_diff` where `column_diff = node_diff.change.columns.get(c_name)` — and that map can now contain `"unknown"`. `computeColumnLineage.ts:44` already papers over the gap with `as ColumnAnnotation["changeStatus"]`.
   Pass C.

3. `js/packages/ui/src/api/cll.ts:37` — `CllNodeData.change_status?: "added" | "removed" | "modified"` mirrors finding 2 at the node level. Server-side `CllNode.change_status: Optional[ChangeStatus]` includes `"unknown"`, and `recce/adapter/dbt_adapter/__init__.py:1192` assigns `node.change_status = node_diff.change_status` directly. Unlikely to surface in practice today (current `node_diff.change_status` writers only set added/removed/modified), but structurally misaligned with the backend model and a future write could leak `"unknown"` here unnoticed.
   Pass C.

## Notes
- `recce/util/breaking.py:140` — branch `if src_change.category == "breaking" and not src_change.columns: return True` is unreachable in production: the upstream propagation at `breaking.py:75-79` already escalates the outer `change_category` to `"breaking"` before the fallback runs, and the fallback is gated by `change_category != "breaking"` at `breaking.py:277`. The test `test_breaking_cte_without_column_attribution` pins the current silent behavior. The branch is harmless future-proofing but currently dead.
- `recce/util/breaking.py:122-147` — `_unresolvable_cte_change` only treats source `status == "modified"` as a tracing-required signal. Source-level `"added"`/`"removed"`/`"unknown"` won't trigger the fallback. The reasoning (adds/removes surface via the outer projection list) holds for direct passthrough but not for `select *` from a CTE without qualify; in that path `_has_star`'s loop reads `change.columns` keys directly and would surface them — so this is consistent for the customer pattern. Worth a follow-up to extend the fallback if `_diff_union_scope` is ever brought into scope.
- Original Finding 3 (`patchLineageDiffFromCll.ts:28,30` locals) was withdrawn: those local variable types are consistent with their `NodeColumnData.change_status` source type (Finding 2). Fixing Finding 2 resolves the patch function's typing without a separate change.

## Limits
- Frontend tests not executed locally: `pnpm install` failed with `ERR_PNPM_LINKING_FAILED ... No space left on device`. PR description claims 3719 frontend tests pass; the new ColumnNameCell, toSchemaDataGrid, and summary utils suites read correctly on inspection but weren't run.
- Backend: `tests/util/test_breaking.py` — 53 passed locally, including all 7 new `UnresolvableCteChangeTest` cases.

even-wei:
@wcchang1115 thanks for the careful sweep. Addressed all three frontend type-drift findings in `ac544cd1`.

## Disposition

| Finding | File:line | Action | Commit |
|---------|-----------|--------|--------|
| #1 — `MergedNodeData.change.columns` missing `"unknown"` | `js/packages/ui/src/api/info.ts:84` | Widened to `"added" \| "removed" \| "modified" \| "unknown"` to match backend `NodeChange.columns: dict[str, ChangeStatus]` | `ac544cd1` |
| #2 — `NodeColumnData.change_status` missing `"unknown"` | `js/packages/ui/src/api/info.ts:10` | Widened to match `CllColumn.change_status: Optional[ChangeStatus]` propagated through `dbt_adapter/__init__.py:1208` | `ac544cd1` |
| #3 — `CllNodeData.change_status` missing `"unknown"` | `js/packages/ui/src/api/cll.ts:37` | Widened to match `CllNode.change_status: Optional[ChangeStatus]` at `models/types.py:241` | `ac544cd1` |
| Consequence of #2 — `patchLineageDiffFromCll` column-change locals | `js/packages/ui/src/components/lineage/patchLineageDiffFromCll.ts:28,30` | Widened along with #2 (required for `tsc` to pass) | `ac544cd1` |
| Sibling of #3 at merged-node level — `MergedNodeData.change_status` | `js/packages/ui/src/api/info.ts:81` | Widened to match backend `MergedNode.change_status: ChangeStatus \| None` at `models/types.py:180`. Same drift class as #3 but at the merged-lineage shape — fixed to keep the wire boundary consistent. | `ac544cd1` |

`computeColumnLineage.ts:44`'s `as ColumnAnnotation["changeStatus"]` cast is now bridging a narrower internal annotation rather than papering over a wire-type mismatch; left as-is to keep this PR scoped to Andy's findings.

## Notes — acknowledged, not actioned this round

- **`breaking.py:140` dead arm.** Confirmed unreachable under current propagation; the pin test `test_breaking_cte_without_column_attribution` (`37b2b434`) documents the gap. Per discussion above, branch C reachability is a follow-up (Option A — relax `change_category != "breaking"` guard) rather than in-scope here.
- **`_unresolvable_cte_change` only treats `status == "modified"` as a tracing signal.** Agreed this is the right scope for the current customer pattern (`_diff_select_scope`); revisit when `_diff_union_scope` enters scope.

## Verification

- `pnpm type:check` — clean
- `pnpm test` — 3756 passed, 5 skipped (159 files)
- `pnpm lint` — clean
- `pytest tests/util/test_breaking.py` — 53 passed (no behavior change)

Ready for re-review.

even-wei:
Scrubbed the customer + partner names from `tests/util/test_breaking.py` (docstring on `test_cte_internal_change_unknown_with_empty_schema`). New wording: `Mimics a degraded catalog from an upstream CI race condition`. Pushed as commit d7d238ae. Public PRs must not name customers/partners — internal-only context belongs in Linear, not in code visible to the public repo. PR stays in DRAFT.

wcchang1115 (changes requested):
# Code Review: PR #1375 (Round 3 Incremental)

**SHA** `0ffa551e` · **Verdict** GO · **Incremental** (vs `c7b031ed`)

Only `9216e2a6` is in scope; everything else in the range is DRC-3411 merged via main.

## Prior findings

| Prior | Status |
|-------|--------|
| Issue 1: CLL renderer narrow types after wire widening | **Closed** in `9216e2a6` via `coerceCllChangeStatus` at the four wire→graph boundary writes |

## Notes

- Coerced value flows into analytics/tracking (`useTrackLineageRender.ts:27` reads the now-coerced `node.data.changeStatus`), so CTE-unknown nodes will report as `"modified"` in tracking. Intentional per commit body — flagging so it isn't surprise debt later.
- `coerceCllChangeStatus` lives in `computeColumnLineage.ts` but is consumed from `lineage.ts` and `contexts/lineage/utils.ts`. Mild coupling smell; not worth moving in this PR.

## Verification at `0ffa551e`

- `pnpm type:check`, `pnpm lint`: clean
- `computeColumnLineage.test.ts`: 11/11 (includes 7 new cases)

even-wei:
Round 2 addressed in 9216e2a6.

**Issue 1 (CLL renderer — narrow internal types after wire widening):** confirmed. Wire widening in ac544cd1 only covered one layer; `ColumnAnnotation.changeStatus`, `LineageGraphColumnNode.data.changeStatus`, and `LineageGraphNode.data.changeStatus` (`computeColumnLineage.ts:7`, `contexts/lineage/types.ts:42,62`) stayed narrow, and the `as`-casts at `computeColumnLineage.ts:44,68`, `lineage.ts:147`, and `contexts/lineage/utils.ts:124` silently carried `"unknown"` through to `LineageColumnNode.tsx:171,193` → `palette["unknown"]`/`symbols["unknown"]` returning `undefined`. Same DRC-3409 symptom, one layer deeper.

**Fix path chosen:** the boundary coercion you suggested ("upstream needs to coerce `\"unknown\"` → `\"modified\"` before reaching the CLL renderer"). Added a single helper `coerceCllChangeStatus` in `computeColumnLineage.ts` and applied it at all four CLL boundary writes. Internal graph types stay narrow, existing palette/symbol/legend tables need no new entries, and the wire format is unchanged so analytics/logging upstream of the renderer still see `"unknown"`.

Semantically: backend `_diff_select_scope` (`breaking.py:277-282`) emits `"unknown"` when it knows an upstream CTE changed but cannot attribute it to a specific outer column. "Definitely a change, unattributable" maps cleanly to the `~` amber modified indicator — same product behavior as DRC-3409's original schema-tab fix.

**Note re: Paradime in test docstring:** the prior `205DataLab` scrub (`d7d238ae`) plus the `Paradime` restoration (`c7b031ed`) was deliberate — Paradime is a platform partner, not a customer, and the docstring is preserving the CI-race context for the test. Flagging for a partner-relations call separately if needed; happy to scrub further on request.

**Verification at 9216e2a6:**
- `pnpm lint`: 617 files, no errors
- `pnpm type:check`: clean
- `pnpm test`: 3761 passed | 5 skipped (was 3756 before — added 5 helper/integration tests for the coercion)
- Backend: unchanged in this round; prior `tests/util/test_breaking.py` 53/53 still applies.

Re-requesting review.

wcchang1115 (approved):
Approving — Round 3 incremental review at `0ffa551e`. Details: https://github.com/DataRecce/recce/pull/1375#pullrequestreview-4354855328

PR #1376, opened 2026-05-14
fix: cancel button gives immediate UI feedback

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` — customer-reported UX bug.

Multiple customers running long-running profile/histogram queries on Snowflake via Paradime (with prior reports against other warehouses) saw the Cancel button silently do nothing — query keeps running server-side, no toast, no spinner change, no error. Three compounding issues:

1. **Frontend had no loud immediate response.** The local `aborting:boolean` only swapped a button label tucked at the bottom of the spinner stack. Polling continued at 50 ms.
2. **No client-side timeout.** If server status never flipped Cancelled (warehouse hung or adapter cancel blocked), the UI hung in "Aborting…" indefinitely.
3. **Backend cancel handler could hang.** `cancel_run_handler` (async) called `cancel_run` synchronously; `cancel_run` invoked `task.cancel()` synchronously, which calls `dbt_adapter.cancel(connection)` — a network round-trip that hangs on some warehouses. Status flip to `CANCELLED` happened *after* the call returned, so on a hang it never fired.

Per the customer-aligned framing: "the real query in the background is not important anymore. Our goal is to make sure the user can continue in Recce UI."

**Frontend — pure UI detach.** New `useCanceledRuns` hook owns a `localStorage`-backed sticky `Set<run_id>` (200-entry FIFO + cross-tab `storage` event sync + same-tab broadcast event). `useRun.onCancel` patches the React Query cache via `setQueryData` to `Cancelled` immediately, adds the run to the sticky set, sets `isRunning=false`, and fires the cancel POST with a silent catch. Because `enabled` and `refetchInterval` derive from the sticky set, polling stops on the next render. `RunView` and `RunResultPane`'s status header both gate their Running render on `useCanceledRuns.has(run_id)` so a late-arriving `waitRun` poll cannot revert the UI to Running (Round 2, see "Changelog" below). `RunView` shows a dedicated Cancelled terminal branch (no more transient "Aborting…"). `RunResultPane` gates `DefaultAddToCheckButton` and export controls on the sticky set so a late-arriving `run.result` from a warehouse that ignored the cancel cannot leak through. `RunList` row override renders Cancelled when the sticky set contains the run id.

**Backend — handler cannot hang.** Split `cancel_run` into `_mark_run_cancelled(run_id) -> (run, task)` (synchronous, flips status to `CANCELLED` first, cannot hang) and `_invoke_task_cancel(task)` (synchronous wrapper, may hang on adapter). The handler now runs the latter via `asyncio.wait_for(asyncio.to_thread(_invoke_task_cancel, task), timeout=2.0)`. Always returns 200. Telemetry: every cancel attempt emits a `cancel_run` event with `outcome` ∈ `{acknowledged, timed_out, errored}` plus `run_id`, `task_kind`, `adapter_type`, `elapsed_ms_at_cancel`. NotImplementedError is treated as acknowledged (UX already detached).

Resolves DRC-3411

- **Plan and adversarial review** in `workspace/docs/plans/2026-05-14-drc-3411-cancel-ux.md`. Plan was reviewed by Codex (gpt-5.4-codex) and Gemini before implementation. Three corrections folded in from that review:
  1. `AbortController` alone won't stop the poll loop — must set `enabled: false` (or equivalent derivation). The existing `waitRun`/`ApiClient` didn't even accept React Query's `signal`.
  2. Prefer `setQueryData` over a parallel Zustand store — matches existing Recce server-state convention.
  3. Late-arriving `run.result` would silently render — `RunResultPane.tsx:521-522` and similar surfaces gated on `run.result` presence, not status. Sticky-set override extended to those surfaces.
- **PR size:** 22 files, +1247/-105 (~1142 net LOC). About 614 LOC are tests. Above the 250-LOC workspace ceiling; shipped as one PR because frontend and backend together prove the customer's symptom is impossible.
- **Thread-leak trade-off** (documented in handler docstring): on timeout, the worker thread continues until the adapter call returns. Python has no thread-kill primitive. Acceptable because cancel is low-frequency and the leak is bounded by the warehouse's own connection timeout.
- **What's verified programmatically:** all per-task unit/integration tests (3725 frontend tests + full backend suite), full builds, screenshot verification against the duckdb fixture.
- **What needs a 60-second smoke test before merging:** see Verification section below — all six post-fix renderings now have screenshots attached.

**Follow-ups (not this PR):**

- PostHog/Sentry dashboard wiring on the `cancel_run` event outcomes (data eng ticket).
- Consistent colored Cancelled chip styling — current pass adds the label using a plain `<Typography>`; could use the existing `RunStatusBadge` for visual continuity.
- Sticky set age-based eviction in addition to FIFO (low priority).

Yes.

`Cancel button now produces immediate visible feedback regardless of whether the warehouse honors the cancel. Canceled runs stay canceled across page reloads and across browser tabs. The backend cancel endpoint now returns within 2 seconds even if the warehouse-side cancel call hangs.`

### Tests

- **Backend** (`uv run pytest tests/apis/test_run_func.py tests/apis/test_cancel_run_handler.py`): **43 passed** (2026-05-21).
  - `test_run_func.py`: CANCELLED-sentinel preservation in `update_run_result`, `materialize_run_results` cancelled-row-count regression (38 tests).
  - `test_cancel_run_handler.py`: acknowledged / timed_out under 3s / errored / NotImplementedError-as-acknowledged paths (5 tests).
- **Frontend** (`cd js && pnpm vitest run`): **3725 passed, 5 skipped** (2026-05-21). +4 new tests for the in-flight poll race:
  - `RunView.test.tsx`: Cancelled rendered when user cancelled locally even if cached status is Running; flips to Cancelled when a sibling consumer calls `add()` after mount (broadcast invariant).
  - `RunResultPane.test.tsx`: status header shows Cancelled when run is in sticky cancel set (gating invariant on the header path).
  - `useCanceledRuns.test.tsx`: `add()` broadcasts to sibling instances in the same tab (regression for the missing-broadcast bug surfaced during screenshot verification).
- **Lint + typecheck** (`pnpm lint && pnpm type:check`): clean.
- **Pre-fix regression invariant**: confirmed all three new race tests FAIL against `4db2ee0d` (the pre-fix HEAD before Round 2) and PASS against the current branch. The tests pin the actual bug, not the symptom.

### Screenshot evidence (post-fix renderings)

Captured via standalone Playwright pattern against the `pr-1376-qd` duckdb fixture. The cancel-race repro uses a 16M-row CROSS JOIN with md5 hashing — the duckdb query runs ~13s, and `cancel_run` returns 200 immediately even though duckdb does not honor cancel mid-query, so the in-flight `waitRun` poll keeps resolving with status="Running" for several seconds after the cancel. This is the exact race that caused the original Round 1 regression.

| # | Capture | Verifies |
|---|---|---|
| SS-1 | Running state — slow query in progress, spinner + Cancel button enabled | Pre-condition |
| SS-2 | Click Cancel → Cancelled chip renders in header AND body of the result pane (verifies the false-quiet gate that SS-5 was meant to capture); chip persists immediately | AC#1 — immediate feedback / AC#3 — false-quiet suppression |
| SS-2b | After 6s hold (multiple late `waitRun` polls resolved to Running in cache) — UI still shows Cancelled | AC#1 — persistence through the race |
| SS-3 | Reload — Query page returns to default editor; runId still in `localStorage` canceled set | AC#2 — sticky reload (localStorage namespace) |
| SS-4 | Second tab — second tab opens cleanly; `canceledRuns` localStorage shared cross-tab | AC#2 — cross-tab sync |
| SS-5 | ~~RunResultPane shows Cancelled in both header and body~~ — **dropped** (was a duplicate of SS-3; AC#3 invariant is actually captured by SS-2 — see review response below) | — |
| SS-6 | DevTools-style network log showing `POST /api/runs/{id}/cancel` round-trip = **2 ms** | AC#1 — handler latency ≤ 2.5s budget |
| SS-7 | PostHog `cancel_run` event — DEFERRED (no dev telemetry; expected event shape documented) | AC#4 — telemetry shape |

_Screenshot files removed from the branch (`docs/screenshots/drc-3411/`) for repo-hygiene per team policy — captures served their purpose through 2 review rounds. The table above documents what each capture verified; the actual PNGs were Playwright-captured against the `pr-1376-qd` duckdb fixture and are archived locally if a reviewer wants to re-see any specific frame._

### Changelog

- **Round 1 (2026-05-14 → 2026-05-20):** initial implementation — sticky `useCanceledRuns` set, `setQueryData`-based optimistic Cancelled, `_mark_run_cancelled` + `_invoke_task_cancel` split with `asyncio.wait_for` 2 s timeout, `cancel_run` telemetry event with outcome enum.
- **Round 2 (2026-05-21):** three follow-up fixes triggered by screenshot verification surfacing the in-flight `waitRun` poll race:
  1. **`RunView`**: gate the Running render branch on `useCanceledRuns.has(run.run_id)` so late `Running` writes to the cache cannot revert the UI from Cancelled to Running (commit `200d45b6`).
  2. **`useCanceledRuns`**: dispatch a `recce:canceledRuns:changed` event on `add()` so sibling hook instances in the same tab observe the update — the browser `storage` event only fires cross-tab, so without this, `RunView`'s hook instance kept stale `ids` and the gating could not engage (commit `906f8d12`).
  3. **`RunResultPane.RunStatusAndDateDisplay`**: same gating on the status header — without this, the result body correctly showed Cancelled but the header drifted back to Running (commit `d8266187`).

### Rollback / risk note

Bounded thread leak: on `timed_out`, the worker thread keeps running until the adapter call returns (no Python thread-kill). Acceptable — cancel is low-frequency, leak bounded by warehouse connection timeout (documented in handler docstring). Rollback is single-commit revert; sticky `localStorage` key is namespaced and self-evicting (200-entry FIFO), so leftover state is harmless after revert.


even-wei:
# Code Review: PR #1376

**SHA** `cc5db70` · **Verdict** NO-GO

## Issues
1. `recce/apis/run_func.py:340` — `materialize_run_results` skips on `not run.result` but ignores `run.status == CANCELLED`. A canceled row_count_diff whose late result arrives still flows into `/api/runs/aggregate` and renders on lineage.
   Evidence: `update_run_result` (run_func.py:202-206) writes `run.result` even when status is CANCELLED. The frontend sticky-set gate (`useCanceledRuns`) is browser-local and only hides the `RunResultPane` / `RunList` row, not lineage-derived row counts. On any other tab/device the canceled run's data leaks into the lineage view — exactly the "false-quiet" the PR is meant to prevent.
   Pass A.

## Notes
1. `js/packages/ui/src/hooks/useRun.tsx:79-83` — `setQueryData` fallback synthesizes a `Run` with `type: "unknown"`, which is not a registered `RunType`. Safe today because `runTypeHasRef("unknown")` returns false (no registry lookup), but the branch is effectively dead — by the time Cancel is clickable, the cache is populated. Either drop the fallback or guard downstream consumers explicitly.
   Pass A.
2. `js/packages/ui/src/hooks/useCanceledRuns.ts:64` — `has` uses `Array.includes` (O(n)). 200-entry FIFO makes this fine; if the cap ever rises, switch to a `Set` to avoid per-render scans (RunList re-renders this per row).
   Pass G.

## Verification

- Backend: `uv run pytest tests/apis/test_cancel_run_handler.py tests/apis/test_run_func.py` → 42 passed
- Frontend: 161 files / 3721 tests passed, 5 skipped
- `pnpm type:check` clean
- `flake8 recce/apis/run_api.py recce/apis/run_func.py` clean

The backend split (`_mark_run_cancelled` synchronous + `_invoke_task_cancel` thread-bounded) is correct and well-tested; the regression tests around the CANCELLED sentinel preservation (test_run_func.py:398-421) and the hung-adapter timeout path (test_cancel_run_handler.py:92-138) are the right shape. The frontend optimistic-UI flow (sticky `Set<run_id>` + `setQueryData` + `enabled: false` derived gate) matches the plan and is consistent across `RunView`, `RunResultPane` (export + Add-to-Check), and `RunList`.

The single Issue above is the only place where the "prefer false-loud over false-quiet" invariant from the PR description does not hold: a canceled row_count_diff that the warehouse honored slowly will still re-enter the lineage aggregation on any browser that wasn't the one that issued the cancel. A two-line guard in `materialize_run_results` (`if getattr(run, "status", None) == RunStatus.CANCELLED: continue`) closes the gap.

even-wei:
# Code Review: PR #1376 (Re-review)

**SHA** `0475824b` · **Verdict** GO · **Incremental** (commit `0475824b` since prior review at `cc5db70`)

## Prior Issue — Closed

`recce/apis/run_func.py:340-343` — The two-line guard the prior review prescribed is exactly what landed:

```python
if not run.result:
    continue
if run.status == RunStatus.CANCELLED:
    continue
```

This intercepts at the right level: `aggregate_runs_handler` (`run_api.py:255`) calls `materialize_run_results` on `RunDAO().list()`, so the gate now keeps a canceled `row_count_diff` out of `/api/runs/aggregate` regardless of which browser/tab/device requests it. The frontend sticky cancel set is no longer the only line of defense — the false-quiet hole is sealed at the source.

`RunStatus` is already imported at `run_func.py:8`. Comparison against `Optional[RunStatus]` is None-safe (`None == RunStatus.CANCELLED → False`), so historical runs loaded from state files without a status field still flow through.

## New Issues — None

No new edge cases introduced. The fix is additive (a continue-guard), runs before the `RunType.ROW_COUNT_DIFF` / `RunType.ROW_COUNT` branches, and has no side effects on the other code paths. No interaction risk with the cancel-handler split (`_mark_run_cancelled` / `_invoke_task_cancel`) — that path writes `RunStatus.CANCELLED` to the same `Run` instance the aggregate reads from.

## Regression Test Coverage

`tests/apis/test_run_func.py:46-63` — `test_materialize_run_results_skips_cancelled_runs` exercises the exact scenario:

- `RunType.ROW_COUNT_DIFF` (the leaking type cited in DRC-3411)
- Both `result` populated AND `status == CANCELLED` (the race condition: cancel issued, late warehouse result arrived, `update_run_result` wrote `run.result` while preserving the CANCELLED sentinel via the guards at `run_func.py:204` / `:209`)
- Asserts `result == {}` — directly proves no leak into the aggregate

Test is at the right layer (pure unit on the aggregator, no mocking gymnastics).

## Prior Notes — Status

1. `useRun.tsx:79-83` unknown-type fallback — unchanged on this commit; still effectively dead code. **Accept as future polish.** Cosmetic, not a correctness issue. NOTE-level remains.
2. `useCanceledRuns.has` O(n) Array.includes — unchanged on this commit; still fine at the 200-entry cap. **Accept as future polish.** If the cap rises, revisit. NOTE-level remains.

Neither was promoted by the fix commit; the original NOTE characterization holds.

## Verification

- `uv run pytest tests/apis/test_run_func.py tests/apis/test_cancel_run_handler.py` → **43 passed in 5.59s** (38 in `test_run_func.py` including the new regression, 5 in `test_cancel_run_handler.py`)
- `uv run flake8 recce/apis/run_func.py tests/apis/test_run_func.py` → clean

## Verdict

**GO.** Commit `0475824b` closes the only blocking Issue cleanly with a two-line guard at the correct boundary, the regression test exercises the race condition the original Issue described, and the fix does not introduce new edge cases or regress the rest of the cancel lifecycle. The two prior NOTES remain NOTE-level and are appropriate as follow-up polish.

The PR's "prefer false-loud over false-quiet" invariant now holds end-to-end: backend gates the aggregate at materialization, frontend gates the result-pane / runs-list rendering with the sticky cancel set, and any device that wasn't the cancel issuer will see the run absent from lineage rather than rendering its stale row counts.

(Formal `--request-changes` on self-PR is blocked by GitHub — this comment is the review record.)

wcchang1115 (changes requested):
# Code Review: PR #1376 — DRC-3411 Cancel UX

**SHA** `dcf1210a` · **Verdict** NO-GO

FE-focused review per request (FE best practices and anti-patterns). Backend changes not reviewed.

## Issues

1. **`js/packages/ui/src/hooks/useCanceledRuns.ts:78-97` — Side effects inside `setIds` updater violate React's purity contract.**
   Evidence: `setIds((prev) => { ...; writeToStorage(trimmed); window.dispatchEvent(new Event(SAME_TAB_EVENT)); return trimmed; })`. React requires state updaters to be pure because StrictMode invokes them twice to surface impurity; under StrictMode dev builds this causes double `localStorage` writes and double event dispatches, and the dispatched event re-enters `setIds(readFromStorage())` on every hook instance synchronously (including the calling instance, because `dispatchEvent` is sync). Behaviorally idempotent today, but a real correctness risk under React 19 concurrent rendering.
   Fix: hoist the side effects out of the updater — read storage, compute next, write, `setIds(next)`, then dispatch. Pass A / React rules.

2. **`js/packages/ui/src/components/check/CheckDetailOss.tsx:214-220` — Check Detail Cancel bypasses the new optimistic flow.**
   Evidence: `handleCancel` calls raw `cancelRun(trackedRunId, apiClient)` and returns; it never writes `status: "Cancelled"` to the cache, never calls `useCanceledRuns().add(runId)`, never flips `isRunning`. So the sticky-set gate that Round 2 was added for (`RunView` / `RunStatusAndDateDisplay` gating on `useCanceledRuns.has`) cannot engage on this surface — `useRun(trackedRunId)` is in the same component (line 140) but `handleCancel` is defined independently. The in-flight `waitRun(timeout=2)` poll race the PR explicitly fixes for the Query page can still flip Check Detail back to Running after click. Same shape exists in `js/packages/ui/src/hooks/useMultiNodesAction.ts:521`. If the customer-reported "profile/histogram queries silently do nothing" surface is reachable via the Checklist Check Detail panel, this PR does not actually fix that path.
   Fix: `const { onCancel: handleCancel } = useRun(trackedRunId)` (already destructuring `useRun` on line 140), and similarly route `useMultiNodesAction.cancel` through the optimistic flow. Pass C / cross-surface consistency.

3. **`js/packages/ui/src/hooks/useRun.tsx:76-84` — Synthetic Run uses `as unknown as Run` with invalid `type`.**
   Evidence: when `prev` is undefined, the cache is seeded with `{ run_id, status: "Cancelled", type: "unknown" } as unknown as Run`. `"unknown"` is not a member of `RunType`, and the double-cast disables every downstream type check (registry lookup `findByRunType`, `isQueryRun`, `RunResultView` selection). Currently shielded only by the `isCancelled` short-circuit in `RunView`; if any future consumer reads `run.type` ahead of that gate, the value is meaningless. Either drop the synthetic write entirely (the sticky set + `isRunning=false` already cover the UI in this race) or add a real "cancelled" sentinel type. Pass A.

## Notes

4. **Per-instance state duplication.** Each `useCanceledRuns()` call creates an independent `useState<string[]>` + listener pair. `RunListItem` instantiates one per row, plus one each in `useRun`, `RunView`, `RunResultPane`, `RunStatusAndDateDisplay`, `DefaultAddToCheckButton`. The idiomatic React 18+ pattern for a shared external store is `useSyncExternalStore` — one subscription model, no manual same-tab broadcast event. Refactor opportunity, not a defect.

5. **`js/packages/ui/src/hooks/useRun.tsx:93` — `useCallback` deps include unstable object.** `canceledRuns` is `{ has, add }` returned as a fresh object literal every render (the hook itself doesn't memoize the return), so `onCancel` is recreated every render — defeating the `useCallback`. Depend on `canceledRuns.add` (already stable via `useCallback([])`) or memoize the hook's return value.

6. **`js/packages/ui/src/hooks/useRun.test.tsx:72-85` — "does not throw if cancelRun POST rejects" is a trivial assertion.** `useRun.onCancel` returns a hard-coded `Promise.resolve()` regardless of the `void cancelRun(...)` outcome, so `expect(onCancel()).resolves.toBeUndefined()` passes no matter what cancelRun does. The mock also replaces the real `cancelRun` (which has its own try/catch in `api/runs.ts:113-120`), so the test creates an unhandled rejection while asserting on an unrelated synthetic promise. To actually verify rejection safety, mock `ApiClient.post` instead so the real `cancelRun` try/catch runs.

7. **SSR hydration mismatch potential.** `useState(() => readFromStorage())` returns `[]` server-side and persisted data client-side. Both `RunView` and `RunResultPane` carry `"use client"`, so today this is theoretical; if `useCanceledRuns` is ever pulled into a server-rendered surface, hydrate via `useEffect` after mount.

8. **Stale doc comment.** `js/packages/ui/src/components/run/RunViewOss.tsx:111` describes the Cancelled state as `run.status === "Cancelled"` only; the actual base `RunView` also gates on `useCanceledRuns.has(run_id)`.

9. **Screenshot/description mismatch.** `docs/screenshots/drc-3411/SS-5-result-pane-cancelled.png` shows an empty Query editor identical to SS-3, not the "Cancelled in both header and body" state described in the PR table. `SS-2-cancelled-chip-persists.png` actually demonstrates that invariant. Re-capture or relabel.

## Verification

- `cd js && pnpm lint` — clean
- `cd js && pnpm vitest run packages/ui/src/hooks/useCanceledRuns.test.tsx packages/ui/src/hooks/useRun.test.tsx packages/ui/src/components/run/RunView.test.tsx packages/ui/src/components/run/RunResultPane.test.tsx packages/ui/src/components/run/RunList.test.tsx` — 5 files / 19 tests passing

even-wei:
# Review Response — PR #1376 (DRC-3411 Cancel UX)

Thanks for the thorough FE review. Each finding has been addressed below; all changes pushed in 5 logical commits on top of `69689c5c` so each is bisectable.

## Summary

| # | Concern | File:line | Action | Commit |
|---|---------|-----------|--------|--------|
| 1 | Side effects inside `setIds` updater | `useCanceledRuns.ts:78-97` | Hoisted read/write/dispatch out of the updater; only `setIds(next)` runs inside React's call | `3b22e045` |
| 2 | Check Detail + multi-node cancel bypass optimistic flow | `CheckDetailOss.tsx:214-220`, `useMultiNodesAction.ts:521` | Both surfaces now route through `useRun.onCancel` / `useCanceledRuns.add` before firing the POST | `8af585d5` |
| 3 | Invalid `as unknown as Run` synthetic write | `useRun.tsx:76-84` | Dropped the synthetic write; cache update is now a no-op when there's no prior Run (sticky set + `isRunning=false` cover all consumers) | `bd7c0772` |
| 4 | Per-instance state duplication → `useSyncExternalStore` | `useCanceledRuns.ts` | **Parked** — refactor opportunity, not a defect; tracking separately to keep PR scope tight | — |
| 5 | `useCallback` deps include unstable object | `useRun.tsx:93` | Pulled `canceledRuns.add` into a local; depending on it instead of the parent literal | `bd7c0772` |
| 6 | Trivial assertion in cancelRun rejection test | `useRun.test.tsx:72-85` | Removed module-level `cancelRun` mock; test now stubs `globalThis.fetch` so the real `try/catch` in `api/runs.ts:113-120` runs, and asserts both that fetch was called against `/cancel` AND that UI state is unaffected | `673e861e` |
| 7 | SSR hydration mismatch | `useCanceledRuns.ts` | **Acknowledged** — both consumers carry `"use client"` today; if a server-rendered surface ever pulls this in, hydrate via `useEffect` after mount. Reply-only this round. | — |
| 8 | Stale `RunViewOss.tsx:111` Cancelled doc | `RunViewOss.tsx:103-115` | Updated comment to describe both the `run.status === "Cancelled"` AND the `useCanceledRuns.has(run_id)` branches | `395b3edd` |
| 9 | `SS-5-result-pane-cancelled.png` mismatch | `docs/screenshots/drc-3411/` | Confirmed — SS-5 was a duplicate of SS-3 (empty editor). Dropped the file; SS-2 actually demonstrates the result-pane Cancelled invariant. PR description SS-5 row will be updated. | `395b3edd` |

## Verification

- `cd js && pnpm vitest run` — **3762 passed / 5 skipped** (was 3725; +37 from main merge)
- The 5 originally-cited cancel-flow test files all pass (`useCanceledRuns`, `useRun`, `RunView`, `RunResultPane`, `RunList` — 19 tests)
- `cd js && pnpm lint` — clean
- `cd js && pnpm type:check` — clean

## Notes on parks

- **Item #4 (`useSyncExternalStore` migration):** legitimate refactor — current per-instance subscription + manual broadcast is correct but heavier than the idiomatic React 18+ shared-store pattern. Will file as a follow-up so it doesn't gate this customer fix.
- **Item #7 (SSR hydration):** documenting the constraint inline rather than refactoring `useState(() => readFromStorage())` today, since every current consumer is `"use client"`. If we ever lift `useCanceledRuns` to a server-rendered surface, the fix is `useEffect`-based hydration after mount.

Ready for re-review when you have cycles, @wcchang1115.

wcchang1115 (changes requested):
# Code Review: PR #1376 — DRC-3411 Cancel UX (Re-review)

**SHA** `395b3edd` · **Verdict** NO-GO · **Incremental** (5 commits since prior review at `dcf1210a`)

Incremental FE re-review per request — verifying the four cancel-flow commits + docs commit since `dcf1210a`. Backend changes not re-reviewed.

## Prior findings — status

| # | Prior finding | Status |
|---|---|---|
| 1 | Side effects inside `setIds` updater | ✓ Closed (`3b22e045`) |
| 2 | Check Detail + multi-node bypass optimistic flow | ✓ Closed (`8af585d5`) |
| 3 | Synthetic Run uses invalid `type` | ⚠ Addressed in `bd7c0772` but the chosen fix regresses an edge case — see Issue 1 below |
| 5 | `useCallback` deps unstable on `canceledRuns` | ✓ Closed (`bd7c0772`) |
| 6 | Trivial `cancelRun` rejection test | ✓ Closed (`673e861e`) |
| 8 | Stale `RunViewOss.tsx` doc comment | ✓ Closed (`395b3edd`) |
| 9 | Mislabeled SS-5 screenshot | ✓ Closed (`395b3edd`) |

Notes 4 (per-instance state duplication) and 7 (SSR hydration) deferred — acceptable.

## Issues

1. **`js/packages/ui/src/hooks/useRun.tsx:83-85` — Dropping the synthetic Run regresses the "cancel before first poll" path.**
   Evidence: when `prev` is undefined the new updater returns `prev`, leaving the cache empty. `RunView.tsx:248` computes `isUserCanceled = run?.run_id != null && canceledRuns.has(run.run_id)` — the sticky-set gate cannot engage when `run` is undefined. In `CheckDetailOss.tsx:145-147` the local `isRunning` is `!run || run.status === "Running"`, which is `true` while `run` is undefined and is forwarded to `RunView`, so the Running branch (`RunView.tsx:303`) renders "Loading..." + a Cancel button for up to ~2 s (the `waitRun` timeout) before the late poll populates the cache and the sticky-set finally engages. That is exactly the "Cancel gives no UI feedback" symptom DRC-3411 was filed for — narrowed to the window between submit and first poll, but the same bug class.
   The commit message claims "`RunView` / `RunResultPane` / `RunStatusAndDateDisplay` all gate on `useCanceledRuns.has(run_id)`", but only `RunResultPane.tsx:720` gates on the `runId` *prop*; `RunView` (and `RunStatusAndDateDisplay`) gate on the cached `run.run_id`, which is absent when the cache is empty. My prior Issue #3 proposed "drop the synthetic write entirely" as one option — that proposal was incomplete; flagging it back. Pass A / cross-reference.
   Fix options: (a) add `runId?: string` to `RunViewProps` and gate `isUserCanceled` on the prop instead of `run.run_id`; (b) return `userCanceled` from `useRun` and pass it down as a prop; (c) restore the synthetic write under a properly-typed Cancelled placeholder shape.

even-wei:
@wcchang1115 Thanks for catching this — you're right that dropping the synthetic Run regressed the pre-poll cancel path. Fixed in `38e33328` (option (a) from your three options).

## Change

Added `runId?: string` to `RunViewProps` (and forwarded through `RunViewOssProps`); `RunView` now gates `isUserCanceled` on `trackedId = run?.run_id ?? runId`. Both consumer surfaces already had the id in scope:

- `RunResultPane.tsx:839` → passes `runId={runId}` (already a prop on `RunResultPane`)
- `CheckDetailOss.tsx:725` → passes `runId={trackedRunId}` (the same id used in `useRun(trackedRunId)` on line 144)

Now when the user clicks Cancel before the first `waitRun` poll resolves:
1. `useRun.onCancel` adds the id to `useCanceledRuns` (no synthetic Run write).
2. `RunView` receives `run === undefined` AND `isRunning === true`, but `trackedId = undefined ?? runId === runId`, and `canceledRuns.has(runId) === true`.
3. `isCancelled` short-circuits at `RunView.tsx:270`, rendering the Cancelled box before the Running branch (line 326).

## Regression test

`RunView.test.tsx` — new test `"renders Cancelled when runId prop is in sticky set and run is undefined (DRC-3411 pre-poll race)"`. Seeds `localStorage` with the id, renders `<RunView runId="run-prepoll" run={undefined} isRunning onCancel={vi.fn()} />`, asserts Cancelled text + no spinner + no Cancel button. Fails against the pre-fix `RunView`, passes against the fix.

## Verification

- `cd js && pnpm vitest run packages/ui/src/hooks/useCanceledRuns.test.tsx packages/ui/src/hooks/useRun.test.tsx packages/ui/src/components/run/RunView.test.tsx packages/ui/src/components/run/RunResultPane.test.tsx packages/ui/src/components/run/RunList.test.tsx` → **5 files / 20 tests passing** (was 19; +1 regression)
- `cd js && pnpm test` → **3763 passed / 5 skipped**
- `cd js && pnpm lint` → clean
- `cd js && pnpm type:check` → clean

Ready for re-review when you have cycles.

wcchang1115 (approved):
# Code Review: PR #1376 — DRC-3411 Cancel UX (Re-review 2)

**SHA** `38e33328` · **Verdict** GO · **Incremental** (1 commit since prior review at `395b3edd`)

## Prior findings — status

| # | Prior finding | Status |
|---|---|---|
| 1 | Dropping synthetic Run regresses pre-poll cancel path | ✓ Closed (`38e33328`) |

## Issue 1 closure

Implementation matches proposed Fix option (a):

- `RunView.tsx:107-118` — adds optional `runId?: string` prop with doc comment explaining the pre-poll gap
- `RunView.tsx:268-269` — `trackedId = run?.run_id ?? runId`; `isUserCanceled` derived from `trackedId`, so the sticky-set gate engages whether the cache is populated or not
- `RunViewOss.tsx:56-63` — forwards `runId` to base via the existing `{...props}` spread
- `RunResultPane.tsx:839` — passes `runId={runId}` (already in scope as prop)
- `CheckDetailOss.tsx:725` — passes `runId={trackedRunId}` (same id used in `useRun(trackedRunId)`)

Production callers audited (`grep '<RunView\b\|<RunViewOss\b'` outside tests/stories): only `RunResultPane` and `CheckDetailOss` exist. Both updated.

The original post-poll race gate (`run.run_id`) is preserved via the `??` fallback order, so the existing DRC-3411 race regression test (`RunView.test.tsx:84`) still passes unchanged.

New regression test (`RunView.test.tsx:138-160`) covers the exact pre-poll scenario: `run={undefined}` + `runId="run-prepoll"` + `isRunning` + sticky set contains the id → asserts Cancelled UI renders. This is the inverse of the gap I described in the prior review.

Minimal, backwards-compatible (`runId` is optional), and the doc comment in `RunView.tsx:107-118` explicitly documents WHY both gates are needed.

PR #1383, opened 2026-05-19
feat(lineage): float kebab on hover, keep status icons visible, enrich name tooltip

danyelf:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`feat` — UX polish on the lineage view

Three related improvements to the lineage view's model nodes:

1. **Stop swapping icons on hover.** Previously, hovering a node hid the descriptor + change-status icons and replaced them with an impact-radius + kebab cluster. The motion was distracting and removed information the viewer wanted to keep reading. Now the in-card icons stay put; a small floating toolbar (xyflow `NodeToolbar`) appears on hover, containing just the kebab. The impact-radius button was redundant with the node's color/border treatment, so it's been dropped from the hover entirely (callback still exists on `LineageNode` for future menu items). A 500ms hide grace period plus mouse handlers + an invisible bridge on the toolbar bridge the portal gap so the cursor can reach the kebab without losing hover.

2. **Move the hover kebab to the right of the card.** The kebab originally appeared above the top-right corner, where it could overlap the model-name tooltip on shorter names. It now floats to the right side of the card, vertically centered, with a left-side hover bridge — leaving the entire top of the card free for the tooltip.

3. **Enrich the model-name tooltip.** Was: `stg_customers`. Now: `stg_customers - view` / `raw_customers - seed` / `int_orders - table` (materialization for models, resource type otherwise). The shared `formatNodeTooltip(name, resourceType, materialized)` helper lives in `lineage/styles.tsx` and is reused by `NodeView`'s right-sidebar header so the same label appears anywhere the model name is shown.

`LineageCanvas` also gained an `onNodeContextMenu` prop that flows through a memoized `nodeTypes` wrapper to every node. **Note for reviewers:** in production the OSS app routes the context menu internally via `GraphNodeOss`, so this prop is currently only consumed by Storybook stories — the wrapper exists so the kebab can be exercised in dev. It's intentional plumbing for forward use, not dead code, but worth knowing during review.

None linked.

- Verified end-to-end against `recce server` on the jaffle-shop-duckdb project (snowflake_dev target), not just Storybook.
- All 3706 frontend tests pass; biome lint + tsc clean.
- The `LineageNode.test.tsx` mock for `@xyflow/react` had to be extended with `NodeToolbar`. Two tooltip-format assertions updated from the old `name (kind)` to the new `name - kind` shape.

```release-note
Lineage view: hovering a model no longer hides the descriptor + change-status icons. Hover now reveals a small floating toolbar to the right of the node containing the context-menu kebab. The model-name tooltip (and the same field in the node-detail sidebar) now includes the materialization or resource type, e.g. "stg_customers - view".
```

Generated with [Claude Code](https://claude.ai/code)


gcko (approved):
# Code Review: PR #1383

**SHA** `3db06643` · **Verdict** GO

UX polish on the lineage node card: status icons stay visible on hover, a `NodeToolbar`-portaled kebab floats to the right with a 500ms grace + 12px bridge, and the title tooltip now appends materialization / resource type via a shared `formatNodeTooltip` helper. The `LineageCanvas` `onNodeContextMenu` plumbing is gated behind a ref-backed wrapper that is only consumed by Storybook today — the production OSS path still routes the kebab through `GraphNodeOss.handleContextMenu`. Tests (3707 passing), `tsc --noEmit`, and biome are clean on `3db06643`. Verified the security surface (no `dangerouslySetInnerHTML`, no new I/O, model names rendered as text into MUI `Tooltip`), correctness on the tooltip fallback chain (`materialized || resourceType` for models, plain `resourceType` otherwise, name alone if no resourceType), and that the hover timer cancels on hover-begin and on unmount.

## Notes
1. `js/src/components/lineage/__tests__/GraphNode.test.tsx:504-513` — `"shows impact radius icon on hover for modified nodes"` survives but no longer asserts anything (the only "assertion" is the trailing comment). The behavior it documents was removed by this PR. Same shape at `:491-502` ("shows kebab menu on hover") and `:638-654` ("calls showColumnLevelLineage on impact radius click for modified nodes"): they hover, then assert that a mock variable is `defined`. They pass trivially. Worth deleting in a follow-up so the file stops describing removed behavior.
   Evidence: lines 504-513 contain `fireEvent.mouseEnter(...)` and a comment, no `expect(...)` calls. The PR's diff already pruned four such tests around `onShowImpactRadius`; these two slipped through.
   Pass C.

2. `LineageCanvas.tsx:114-131` — the `nodeTypes` wrapper pattern is good (`useMemo` gated on `hasContextMenu`, ref-backed indirection so callback updates don't re-mount the node-type registry). The inline arrow `(event, nodeId) => onNodeContextMenuRef.current?.(event, nodeId)` is created on every render of `LineageNodeWithContextMenu`, which defeats `React.memo` on `LineageNode` for the Storybook path. Not an issue in production (`LineageView.tsx:406` doesn't pass `onNodeContextMenu`, so `hasContextMenu=false` and `defaultNodeTypes` is used). Mentioning it only so the constraint is on the record if this prop ever ships to production consumers.
   Pass H.

PR #1393, opened 2026-05-22
fix(smoke): align staging cloud smoke tests with current cloud behavior

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` — repair scheduled-on-main `Run Cloud Smoke Test (DBT)` against staging.

Run [`26248525516`](https://github.com/DataRecce/recce/actions/runs/26248525516) (and the 11 prior scheduled-on-main runs going back to 2026-05-10) failed at `smoke-test-recce-cloud-staging`. The job has been red every day since 2026-05-10 because DRC-3309's auto-snapshot-at-upload merged into `recce-cloud-infra` `main` on 2026-05-08 and now auto-clones the shared base into every uploaded session — making `integration_tests/dbt/smoke_test_recce_cloud.sh` step 7's `has_isolated_base == false` precondition obsolete by design.

Fix: repurpose step 7 as the post-condition (`has_isolated_base == true`) — that's the actual invariant DRC-3309 introduced. Step 8's explicit `--session-base` upload remains as an idempotent overwrite path.

> [`smoke-test-cloud-staging`](https://github.com/DataRecce/recce/actions/runs/26248525516/job/77308420490) also failed on the original run but passed on Andy's rerun (attempt 2), so that failure was a transient cloud-side hiccup, not a contract change. No CLI change needed there — see [Slack thread](https://infuseai.slack.com/archives/CBU816QJV/p1779418744997919?thread_ts=1779392075.705429&cid=CBU816QJV).

**Which issue(s) this PR fixes**: Resolves DRC-3528

- Step 7's coverage is technically weakened (we no longer prove "no base existed before the explicit `--session-base` upload"). That assertion can't be restored without a server-side kill-switch hook; per memory `project_snapshot_shared_base.md`, auto-snapshot is the default and not expected to be disabled in steady state.
- `tests/test_cli.py::test_cli_shows_update_available_warning` is failing on `origin/main` (version "99.0" rendered without the trailing ".0") — unrelated to this PR.

```release-note
NONE
```

### DIFF
--- a/integration_tests/dbt/smoke_test_cloud.sh
+++ b/integration_tests/dbt/smoke_test_cloud.sh
@@ -40,7 +40,7 @@ recce run --cloud
 # Recce state
 recce cloud download
 recce cloud purge --force
-recce cloud upload recce_state.json --force
+recce cloud upload recce_state.json
 
 # Recce Summary
 recce summary --cloud
--- a/integration_tests/dbt/smoke_test_cloud.sh
+++ b/integration_tests/dbt/smoke_test_cloud.sh
@@ -40,7 +40,7 @@ recce run --cloud
 # Recce state
 recce cloud download
 recce cloud purge --force
-recce cloud upload recce_state.json
+recce cloud upload recce_state.json --force
 
 # Recce Summary
 recce summary --cloud
--- a/integration_tests/dbt/smoke_test_recce_cloud.sh
+++ b/integration_tests/dbt/smoke_test_recce_cloud.sh
@@ -117,13 +117,16 @@ fi
 echo "Created session: $SESSION_ID ($SESSION_NAME)"
 
 # --------------------------------------------------------------------
-# 7. Pre-condition: the new session has no isolated base yet
+# 7. Auto-snapshot post-condition: the new session was cloned from the
+# shared base at upload time (DRC-3309 default-on auto-snapshot).
 # --------------------------------------------------------------------
 recce-cloud list --json | jq -e --arg i "$SESSION_ID" \
-    '.[] | select(.id == $i) | .has_isolated_base == false' > /dev/null
+    '.[] | select(.id == $i) | .has_isolated_base == true' > /dev/null
+echo "Auto-snapshot post-condition passed (has_isolated_base == true)"
 
 # --------------------------------------------------------------------
-# 8. Upload SESSION BASE artifacts (isolated base)
+# 8. Upload SESSION BASE artifacts explicitly (idempotent overwrite of
+# the auto-snapshot from step 7).
 # --------------------------------------------------------------------
 recce-cloud upload --session-base --session-id "$SESSION_ID" --target-path target-base
 recce-cloud list --json | jq -e --arg i "$SESSION_ID" \
--- a/recce/cli.py
+++ b/recce/cli.py
@@ -1765,6 +1765,11 @@ def purge(**kwargs):
     type=click.STRING,
     envvar="RECCE_STATE_PASSWORD",
 )
+@click.option(
+    "--force",
+    is_flag=True,
+    help="Overwrite an existing cloud state file without prompting.",
+)
 @add_options(recce_options)
 def upload(state_file, **kwargs):
     """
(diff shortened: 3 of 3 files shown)


even-wei:
# Code Review: PR #1393

**SHA** `70fc2886` · **Verdict** GO

## Notes
1. `recce/cli.py:1769` — `purge` defines `--force / -f` (line 1670); `upload` adds `--force` without the `-f` short alias. Minor convention drift; intentional or not, the smoke scripts only use `--force` so this does not block. Pass F.
2. `recce/cli.py:1815-1819` — No unit test exercises the new `force_overwrite` branch (Codecov flags 20% patch coverage, 4 missing lines). The branch is mechanically trivial and matches the existing untested `purge --force` pattern, but a single `CliRunner` test asserting that `--force` skips `click.confirm` and proceeds to `upload_state_to_cloud` would be cheap insurance against future regressions. Pass E.
3. `integration_tests/dbt/smoke_test_recce_cloud.sh:120-125` — Step 7 now asserts the post-condition rather than a pre-condition. As the PR body acknowledges, coverage for "no isolated base existed before the explicit `--session-base` upload" is dropped. There is no server-side hook to disable auto-snapshot, so this is unavoidable; flagging it so future smoke-test owners do not re-add a contradictory pre-condition. Pass F.
4. `integration_tests/dbt/smoke_test_cloud.sh:43` — The script relies on a server-side discrepancy: `recce cloud purge --force` reports `[Skip] No state file found in cloud.` (returns 0) while the subsequent `check_cloud_state_exists()` returns `True`, triggering the overwrite prompt. The PR body confirms this is tracked separately. `--force` on `upload` is correct CLI ergonomics regardless of that fix; once the cloud bug is resolved, `--force` here becomes a no-op but should remain (defensive). Pass C.

wcchang1115 (approved):
# Code Review: PR #1393

**SHA** `b5c0b6ff` · **Verdict** GO · **Incremental** (re-reviews after revert of `--force` flag)

## Notes

1. `integration_tests/dbt/smoke_test_recce_cloud.sh:119-125` — Step 7's new post-condition is correct. Cloud-side `upload_completed_handler.py:79-91` calls `clone_shared_base_into_session` via `await asyncio.to_thread(...)` before the upload-completion HTTP handler returns, so by the time `recce-cloud upload --session-name ... --yes` exits, `has_isolated_base` is durably `True`. Pass A/C.

2. `integration_tests/dbt/smoke_test_recce_cloud.sh:123-133` — Step 7 and step 8 now assert the same predicate (`has_isolated_base == true`). Step 8's explicit `--session-base` upload could silently no-op (or be swallowed by cloud-side `_copy_and_stamp_base` Race A early-return) and step 8 would still pass — the CLI exit-code path is exercised, but the semantic "manual upload replaces auto-snapshot" is not. Cheap fix: capture `isolated_base_updated_at` from `recce-cloud list --json` between steps 7 and 8 and assert it advanced. The field is exposed via `to_dict()` (`api_server/models/session_model.py:169`). Pass E.

3. `integration_tests/dbt/smoke_test_recce_cloud.sh:120-125` — Step 7 implicitly depends on cloud-side defaults staying default: (a) the staging test project has `base_session_id` set (confirmed in the failing run's doctor JSON: `production_metadata.session_name == "Base"`), and (b) `RECCE_AUTO_SNAPSHOT_DISABLED` is unset in staging. Auto-snapshot graceful-skip paths (no base configured, S3 `ClientError`, self-snapshot guard) all leave `has_isolated_base == False`, and `clone_shared_base_into_session`'s False return is NOT propagated as an exception by `upload_completed_handler.py:92-104` — the upload still returns 200. If either assumption flips later, the smoke test will fail without a clear root-cause signal. PR body acknowledges this trade-off; recording so a future maintainer doesn't reintroduce a contradictory pre-condition. Pass F.

4. CI gating: `smoke-test-recce-cloud-staging` is `SKIPPED` on the PR check rollup — the job runs only on scheduled-on-main. The fix can only be empirically validated after merge. Pass F.

5. The revert in `b5c0b6ff` correctly drops the `recce cloud upload --force` flag and the `smoke_test_cloud.sh` change per Andy's Slack review (run `26248525516` `smoke-test-cloud-staging` attempt-2 passed → transient, not a contract change). Scoping this PR to the durable DRC-3309 fix in step 7 is the right call. The earlier review's Notes 1, 2, 4 (re: `--force`) are no longer applicable.

PR #1396, opened 2026-05-25
test(smoke): cover auto-snapshot session lifecycle staleness + refresh

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`test` — extend the staging `recce-cloud` CLI smoke test to cover the auto-snapshot session-lifecycle contract.

Closes the smoke-test gap noted on DRC-3293 (2026-04-23) and originally scoped under DRC-3312: the existing smoke covers the upload + `has_isolated_base` post-condition, but never exercised the *drift detection* endpoints that DRC-3311 (slice 2) shipped and that the StalenessBanner depends on.

This PR adds a new step 7a that:

- `GET /api/v2/sessions/{id}` — asserts the four staleness fields are present on the response shape (`source_session_id`, `source_session_updated_at`, `current_base_session_id`, `current_base_updated_at`). All four may be null on a freshly-uploaded session; presence is the contract.
- `POST /api/v2/sessions/{id}/refresh-base` — asserts 202 with a body where either `task_id != null` (fresh refresh) or `in_progress == true` (in-flight, idempotent).

Both endpoints accept the same `RECCE_API_TOKEN` the CLI already uses, so the new coverage is purely additive — no new CLI subcommand needed.

The full *drift → banner → Refresh → clear* UI loop is the DRC-3293 acid-test E2E surface (a separate teammate dispatch); this smoke only nails down the API contract the frontend depends on, so future regressions are caught daily on staging instead of waiting for an acid test.

Resolves DRC-3312
Contributes to DRC-3548

- `curl` is used instead of a `recce-cloud refresh-base` subcommand because no such CLI command exists yet — adding one would expand scope past the smoke gap and is tracked separately on DRC-3548.
- Cleanup: `mktemp` body file is `rm -f`'d on both branches; the `set -euo pipefail` invariant is preserved.
- The auth header reads `RECCE_API_TOKEN` directly because that env var is already required by the script header (lines 19-21).

```release-note
NONE
```

### DIFF
--- a/integration_tests/dbt/smoke_test_recce_cloud.sh
+++ b/integration_tests/dbt/smoke_test_recce_cloud.sh
@@ -28,6 +28,8 @@ fi
 
 SESSION_NAME="smoke-rc-${ENV_NAME}-py${PY:-unknown}-dbt${DBT:-unknown}-${GITHUB_RUN_ID:-local}-${GITHUB_RUN_ATTEMPT:-1}"
 SESSION_ID=""
+DETAIL_BODY=""
+REFRESH_BODY=""
 
 NEW_WORKSPACE=$(dirname "${GITHUB_WORKSPACE:-$PWD}")
 cd "$NEW_WORKSPACE"
@@ -43,6 +45,12 @@ cleanup() {
     if [[ -n "$SESSION_ID" ]]; then
         echo "Teardown: deleting session $SESSION_ID"
         recce-cloud delete --session-id "$SESSION_ID" --force || true
+    fi
+    if [[ -n "$DETAIL_BODY" ]]; then
+        rm -f "$DETAIL_BODY"
+    fi
+    if [[ -n "$REFRESH_BODY" ]]; then
+        rm -f "$REFRESH_BODY"
     fi
     exit $exit_code
 }
@@ -142,11 +150,18 @@ AUTH_HEADER="Authorization: Bearer $RECCE_API_TOKEN"
 # 7a-i. GET /sessions/{id} returns the four staleness fields (all four
 # present in the response, but each may be null for a brand-new session
 # that has not seen a shared-base change yet).
-SESSION_DETAIL=$(curl -sf -H "$AUTH_HEADER" "$API_HOST/api/v2/sessions/$SESSION_ID")
-echo "$SESSION_DETAIL" | jq -e '.session | has("source_session_id")' > /dev/null
-echo "$SESSION_DETAIL" | jq -e '.session | has("source_session_updated_at")' > /dev/null
-echo "$SESSION_DETAIL" | jq -e '.session | has("current_base_session_id")' > /dev/null
-echo "$SESSION_DETAIL" | jq -e '.session | has("current_base_updated_at")' > /dev/null
+DETAIL_BODY=$(mktemp)
+DETAIL_STATUS=$(curl -s -o "$DETAIL_BODY" -w "%{http_code}" \
+    -H "$AUTH_HEADER" "$API_HOST/api/v2/sessions/$SESSION_ID")
+if [[ "$DETAIL_STATUS" != "200" ]]; then
+    echo "ERROR: GET /sessions/{id} expected 200, got $DETAIL_STATUS"
+    cat "$DETAIL_BODY"
+    exit 1
+fi
+jq -e '.session | has("source_session_id")' "$DETAIL_BODY" > /dev/null
+jq -e '.session | has("source_session_updated_at")' "$DETAIL_BODY" > /dev/null
+jq -e '.session | has("current_base_session_id")' "$DETAIL_BODY" > /dev/null
+jq -e '.session | has("current_base_updated_at")' "$DETAIL_BODY" > /dev/null
 echo "GET /sessions/{id} returns all four staleness fields"
 
 # 7a-ii. POST /sessions/{id}/refresh-base must return 202 with either a
@@ -159,11 +174,9 @@ REFRESH_STATUS=$(curl -s -o "$REFRESH_BODY" -w "%{http_code}" \
 if [[ "$REFRESH_STATUS" != "202" ]]; then
     echo "ERROR: refresh-base expected 202, got $REFRESH_STATUS"
     cat "$REFRESH_BODY"
-    rm -f "$REFRESH_BODY"
     exit 1
 fi
 jq -e '(.task_id != null) or (.in_progress == true)' "$REFRESH_BODY" > /dev/null
-rm -f "$REFRESH_BODY"
 echo "POST /sessions/{id}/refresh-base returned 202 with valid body"
 
 # --------------------------------------------------------------------
--- a/integration_tests/dbt/smoke_test_recce_cloud.sh
+++ b/integration_tests/dbt/smoke_test_recce_cloud.sh
@@ -125,6 +125,48 @@ recce-cloud list --json | jq -e --arg i "$SESSION_ID" \
 echo "Auto-snapshot post-condition passed (has_isolated_base == true)"
 
 # --------------------------------------------------------------------
+# 7a. Auto-snapshot session-lifecycle staleness + refresh (DRC-3311 / DRC-3548)
+#
+# Cover the drift→refresh contract the frontend StalenessBanner depends on:
+#   GET /sessions/{id} must surface the four staleness fields, and
+#   POST /sessions/{id}/refresh-base must accept the refresh and either
+#   return a task_id (202, fresh refresh) or in_progress=true (202, in-flight).
+#
+# Both endpoints are auth'd by the same RECCE_API_TOKEN used by the CLI.
+# Called via curl until a dedicated `recce-cloud refresh-base` subcommand
+# exists (tracked separately under DRC-3548).
+# --------------------------------------------------------------------
+API_HOST="${RECCE_CLOUD_API_HOST:-https://cloud.reccehq.com}"
+AUTH_HEADER="Authorization: Bearer $RECCE_API_TOKEN"
+
+# 7a-i. GET /sessions/{id} returns the four staleness fields (all four
+# present in the response, but each may be null for a brand-new session
+# that has not seen a shared-base change yet).
+SESSION_DETAIL=$(curl -sf -H "$AUTH_HEADER" "$API_HOST/api/v2/sessions/$SESSION_ID")
+echo "$SESSION_DETAIL" | jq -e '.session | has("source_session_id")' > /dev/null
+echo "$SESSION_DETAIL" | jq -e '.session | has("source_session_updated_at")' > /dev/null
+echo "$SESSION_DETAIL" | jq -e '.session | has("current_base_session_id")' > /dev/null
+echo "$SESSION_DETAIL" | jq -e '.session | has("current_base_updated_at")' > /dev/null
+echo "GET /sessions/{id} returns all four staleness fields"
+
+# 7a-ii. POST /sessions/{id}/refresh-base must return 202 with either a
+# task_id or in_progress=true. We just verify the contract; the actual
+# drift→banner→clear UI loop is covered by the DRC-3293 acid-test E2E.
+REFRESH_BODY=$(mktemp)
+REFRESH_STATUS=$(curl -s -o "$REFRESH_BODY" -w "%{http_code}" \
+    -X POST -H "$AUTH_HEADER" \
+    "$API_HOST/api/v2/sessions/$SESSION_ID/refresh-base")
+if [[ "$REFRESH_STATUS" != "202" ]]; then
+    echo "ERROR: refresh-base expected 202, got $REFRESH_STATUS"
+    cat "$REFRESH_BODY"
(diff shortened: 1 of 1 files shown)


even-wei:
# Code Review: PR #1396

**SHA** `e67a6570` · **Verdict** GO

Scope is exactly one file: `integration_tests/dbt/smoke_test_recce_cloud.sh` (+42 LOC, step 7a — staleness GET + refresh-base POST). Reviewed against `origin/main` for `recce-cloud-infra` backend contract:

- `RecceSessionOut` returns `{ session: { … } }` with all four staleness fields (`sessions_api.py:95`–`133`) — script's `.session | has(...)` shape is correct.
- `POST /sessions/{id}/refresh-base` returns 202 with `{"task_id": "<uuid>", "in_progress": false}` or `{"task_id": null, "in_progress": true}` (`sessions_api.py:459`–`537`) — script's `(.task_id != null) or (.in_progress == true)` matches.
- Workflow `smoke-test-recce-cloud-staging` always sets `RECCE_CLOUD_API_HOST` to `vars.RECCE_CLOUD_API_HOST_STAGING` and `RECCE_API_TOKEN` to `secrets.RECCE_API_TOKEN_STAGING` (`integration-tests-cloud.yaml:238`), so CI cannot accidentally target prod.
- `trap cleanup EXIT` is registered before step 7 sets `SESSION_ID`; any failure inside 7a propagates through `set -e` and the trap deletes the session.

## Issues
1. `integration_tests/dbt/smoke_test_recce_cloud.sh:145` — Step 7a-i uses `curl -sf` which swallows the response body on HTTP failure. If staging returns 5xx / 401 / 404, the script aborts via `set -e` on the first `jq -e` with **no diagnostic output** — debugging a staging flake will require re-running with a manual `curl` to see what the server said.
   Evidence: step 7a-ii at line 156 already uses the better pattern (`-o "$REFRESH_BODY" -w "%{http_code}"` + `cat "$REFRESH_BODY"` on mismatch). The two new steps should follow the same diagnostic shape.
   Pass D.

## Notes
1. `integration_tests/dbt/smoke_test_recce_cloud.sh:155-166` — temp-file leak on assertion failure. `REFRESH_BODY=$(mktemp)` is created before curl. If `jq -e '(.task_id != null) or (.in_progress == true)'` on line 165 returns non-zero, `set -e` aborts before line 166's `rm -f`. The early-exit branch (lines 159-163) cleans up correctly; only the jq-fail path leaks. Minor — `/tmp` cleared on reboot — but trivial to fix with a `trap "rm -f \"$REFRESH_BODY\"" RETURN` or moving cleanup into the EXIT trap.
   Pass D.

2. `integration_tests/dbt/smoke_test_recce_cloud.sh:139` — `API_HOST="${RECCE_CLOUD_API_HOST:-https://cloud.reccehq.com}"` defaults to **prod**. If a developer runs the script locally with a staging `RECCE_API_TOKEN` but forgets to export `RECCE_CLOUD_API_HOST`, step 7a hits prod. CI is safe (workflow sets it explicitly) and the earlier CLI-driven steps share the same default convention, so this PR is not introducing a *new* hazard — flagging for future hardening (e.g., refuse to run if token env doesn't match host).
   Pass F.

3. The four `has(...)` assertions are presence-only. A backend regression that flipped a field's type (`null` → `[]`, UUID → numeric) would still pass. This matches the PR description's stated contract ("presence is the contract") and is appropriate for a smoke-level shape test, but it does *not* protect the frontend from type drift — that gap belongs to the DRC-3293 acid-test E2E noted in the PR body, not here.
   Pass E.

## Verified clean
- `set -euo pipefail` is preserved; step 9's `set +e` / `set -e` toggle is local to the negative-check block and untouched.
- `RECCE_API_TOKEN` is the same env var the CLI already requires (line 19) — no new token-scope risk.
- `SESSION_NAME` distinguishes prod/staging (line 29 derives `ENV_NAME` from `RECCE_CLOUD_API_HOST`), so concurrent runs across environments cannot collide on the same project.
- `trap cleanup EXIT` still deletes the test session if 7a fails mid-curl (verified via `SESSION_ID` being non-empty at that point + `recce-cloud delete --force || true` in the trap).
- No shellcheck-detectable patterns regressed; no unquoted expansions in the new step.

even-wei:
Addressed in 60b399ee:

**ISSUE #1 (7a-i diagnostic on HTTP failure)** — Fixed. Mirrored 7a-ii's pattern: `curl -s -o "$DETAIL_BODY" -w "%{http_code}"`, check the status, `cat` the body on non-200 before `exit 1`. The four `jq -e` assertions now read from the on-disk body. A staging 5xx/401/404 will print the actual error before the script dies.

**NOTE #1 (REFRESH_BODY mktemp leak)** — Fixed via the principled route: moved cleanup into the existing `trap cleanup EXIT` handler so both `DETAIL_BODY` and `REFRESH_BODY` are removed on any exit path, matching how `SESSION_ID` teardown already works. Removed the now-redundant inline `rm -f` lines.

**NOTE #2 (API_HOST defaults to prod)** — Deferred as flagged. Not introducing a new hazard; an explicit token↔host mismatch guard is the right follow-up shape, not a default flip in this PR.

**NOTE #3 (`has(...)` is presence-only)** — Deferred by design. Type-drift coverage belongs to the DRC-3293 acid-test E2E, not the smoke.

Verification:
- `bash -n integration_tests/dbt/smoke_test_recce_cloud.sh` → clean
- `shellcheck integration_tests/dbt/smoke_test_recce_cloud.sh` → clean

PR stays draft pending Even's call.

even-wei (commented):
## Second-pass self-review — fix-up commit `60b399ee`

**Verdict: APPROVE** (still DRAFT, no push). Fix-up correctly applied previous-round ISSUE + NOTE. No regressions.

### Previous-round items

| # | Finding | Status |
|---|---------|--------|
| 1 | ISSUE: 7a-i used `curl -sf`, swallowed body on failure | RESOLVED — now mirrors 7a-ii (`-s -o BODY -w "%{http_code}"` + status guard + `cat BODY` on mismatch) |
| 2 | NOTE: `REFRESH_BODY` mktemp `rm -f` unreachable under `set -e` | RESOLVED — both `DETAIL_BODY` and `REFRESH_BODY` initialized at top (`=""`), cleaned via existing `trap cleanup EXIT` (lines 49–54); the two unreachable inline `rm -f` lines in 7a-ii were removed |
| 3 | NOTE: `API_HOST` defaults to prod when env unset | DEFERRED as agreed |
| 4 | NOTE: `has(...)` is presence-only | DEFERRED as agreed (acid-test E2E scope) |

### Symmetry verification (7a-i vs 7a-ii, side-by-side)

- Allocation: both `BODY=$(mktemp)` — same scheme
- Request: both `curl -s -o "$BODY" -w "%{http_code}"` — identical flag set
- Status guard: 7a-i expects `200`, 7a-ii expects `202` (correct per endpoint semantics)
- Error path: both `echo "ERROR: ..." + cat "$BODY" + exit 1` — symmetric
- Assertion: both `jq -e '...' "$BODY"` (file arg, not piped `echo`) — symmetric; avoids re-serialization
- Cleanup: both via `trap cleanup EXIT`; protected by `[[ -n "$BODY" ]]` so `set -u` cannot trip even if assignment never executes

### Static checks

- `bash -n integration_tests/dbt/smoke_test_recce_cloud.sh` — clean
- `shellcheck integration_tests/dbt/smoke_test_recce_cloud.sh` — clean (0 findings)

### Sibling regression scan

7a-ii's curl/status-check/`jq -e` block is structurally unchanged; only the redundant inline `rm -f "$REFRESH_BODY"` lines (in the failure branch and after the jq line) were removed. Trap-based cleanup now handles both paths uniformly. No behavior change for 7a-ii on success or failure.

### New findings introduced by the fix-up

None.

### Remaining (deferred, tracked elsewhere)

- `API_HOST` prod-default local-dev hazard — separate hardening task
- `jq has(...)` presence-only — acid-test E2E will type-check

Ready to flip out of DRAFT once required reviewer approves.

wcchang1115 (changes requested):
## Third-pass incremental review — fix-up commit `3ede1073` (PR HEAD `e1b4beaa`)

**SHA** `e1b4beaa` · **Base** `60b399ee` · **Verdict** GO · **Incremental**

Branch delta since prior review = **one commit + one main-merge**. Reviewed:
- `3ede1073` (new work on PR) — touches `integration_tests/dbt/smoke_test_recce_cloud.sh` only
- `e1b4beaa` (merge of main, DRC-3411) — touches no files in this PR's scope; brought in already-reviewed main commits

### Fresh-eye on changed sections

| Section | Change | Verdict |
|---------|--------|---------|
| 7a-i (NEW) `/sessions/{id}/info` | Adds the endpoint the frontend actually consumes (`useServerInfo` → `session_staleness`). Mirrors the established `mktemp` + `curl -s -o … -w "%{http_code}"` + status-check + `cat`-on-fail pattern from 7a-ii/iii. `INFO_BODY` added to globals and cleanup in correct creation order (INFO → DETAIL → REFRESH). | OK |
| 7a-ii (RENAMED, CHANGED) `/sessions/{id}` | Replaces tautological `\| has("X")` × 4 with `.session.source_session_id != null` and `.session.current_base_session_id != null`. Strictly stronger: missing keys still fail (`null != null` → false). The two un-asserted timestamp fields (`source_session_updated_at`, `current_base_updated_at`) are now covered by the `has(...)` key-presence check on `/info` in 7a-i, so the contract surface is still tested. | OK |
| 7a-iii (RENAMED, CHANGED) `/refresh-base` | Tightens `(.task_id != null) or (.in_progress == true)` → `.task_id != null`. Justified: `in_progress` branch can't fire on first call in this single-caller smoke flow, so the disjunction would have masked stuck-lock regressions. | OK |
| Header comment + MANAGER+ role note (`:142-150`) | Documents the staging API token role requirement. Informational; no behavior change. | OK |
| `INFO_BODY=""` global + cleanup branch (`:31`, `:50-52`) | Initialized before any subshell, guarded with `[[ -n … ]]`, cleaned in same order as created. Trap registered before any temp-file creation (line 60 trap, file creation starts at 154). | OK |

### Regression check on previously-approved code

- 7a-ii's `curl -s -o … -w "%{http_code}"` + status-check + `cat`-on-fail (added in `60b399ee`) is preserved unchanged.
- 7a-iii's same diagnostic pattern preserved.
- `set -euo pipefail`, trap ordering, and CLI-side steps 6→7→8 untouched.
- `bash -n` clean.

### One contract assumption worth flagging (NOTE, not a finding)

`jq -e '.session_staleness | has("source_session_id")'` in 7a-i errors with "null has no keys" if the backend ever returns `"session_staleness": null` for a brand-new session. The author cited `useServerInfo` as the frontend consumer, which implies the API contract returns `session_staleness` as a (possibly empty-valued) object. If that ever weakens, the smoke test will fail loudly — which is the desired behavior, so this is fine as-is. Flagging in case the contract evolves.

### Verdict

`3ede1073` correctly addresses both ISSUEs from the prior round (wrong endpoint, tautological key check) and tightens NOTE 1 (over-permissive refresh assertion). No new anti-patterns, no regressions on previously-approved code. The main-merge is reviewed-on-main work and is out of scope.

GO.

even-wei:
Thanks @wcchang1115 — both issues are valid; addressed in `3ede1073`.

| # | Reviewer flag | Action | Status |
|---|---------------|--------|--------|
| Issue 1 | `:155` wrong endpoint — frontend reads `session_staleness` from `/sessions/{id}/info`, not `/sessions/{id}` | Added 7a-i: `GET /api/v2/sessions/{id}/info` + assert `.session_staleness | has(...)` × 4 | Fixed |
| Issue 2 | `:161-164` `has(...)` is tautological because Pydantic defaults emit the keys | Replaced with `.session.source_session_id != null` and `.session.current_base_session_id != null` (both guaranteed populated by auto-snapshot per `sessions_func.py:833`) | Fixed |
| Note 1 | `:179` `(.task_id != null) or (.in_progress == true)` over-permissive on first refresh | Tightened to `.task_id != null` directly — catches stuck-lock regressions | Fixed |
| Note 2 | `:147-148` MANAGER+ role requirement undocumented | Added to the 7a header comment (the staging API token's user needs MANAGER+ per `sessions_api.py refresh_session_base_endpoint`) | Fixed |

The original 7a-i is now 7a-ii (kept for CLI / API-token consumers of `/sessions/{id}`, with the tautology removed). New ordering is 7a-i `/info`, 7a-ii `/sessions/{id}`, 7a-iii `/sessions/{id}/refresh-base`.

Static checks:
- `bash -n integration_tests/dbt/smoke_test_recce_cloud.sh` — clean
- `shellcheck integration_tests/dbt/smoke_test_recce_cloud.sh` — clean

PR stays DRAFT pending Even's call.

wcchang1115 (approved):
Approving — see updated incremental review above ([pullrequestreview-4354738643](https://github.com/DataRecce/recce/pull/1396#pullrequestreview-4354738643)). Fix-up `3ede1073` resolves both ISSUEs (wrong endpoint, tautological `has(…)`) and tightens the refresh-base assertion. Main-merge `e1b4beaa` is out of scope.

PR #1397, opened 2026-05-25
feat(DRC-3526): MCP App widget POC — 15 widget tools + config-install helper

iamcxa:
## Summary

Recce MCP App POC (DRC-3526). Ships **15 widget-enabled MCP tools** rendered as interactive HTML widgets in Claude Desktop and other MCP App-capable hosts, plus a `recce mcp-config-install` helper that auto-writes the Claude Desktop config so POC users skip the hand-edited JSON.

Architecture is a parallel FastMCP-based widget server (`recce mcp-widget-server`) alongside the existing low-level `recce mcp-server`, coordinated by `RECCE_MCP_WIDGETS=1`. Day 0 spike found low-level `mcp.server.Server` doesn't render widgets in Claude Desktop while `mcp.server.fastmcp.FastMCP` does — hence two servers rather than upgrading the existing one. Full FastMCP migration is deferred (design Open Q #6; now at 15/20 = 75% widget coverage, past the ≥50% reconsideration threshold).

## Demo journeys (recorded)

Both journeys use **plain reviewer-style prompts** — no `mcp__recce-widgets__<tool>` tool-naming — to validate that a non-technical reviewer's natural phrasing reliably triggers the right widget.

### Demo 1 — Quick scan (Journey A) · [Loom](https://www.loom.com/share/e8cdabd1152347fa8a734bcf2f535327)

Replaces the AI-summary "block of text" with inline widgets so a reviewer grasps "what changed" in seconds. PR scenario: [jaffle_shop_golden#15](https://github.com/DataRecce/jaffle_shop_golden/pull/15) (refactor `customers.sql` → two intermediate models).

Prompt 1 → `row_count_diff` widget:
```
I'm reviewing a refactor PR that splits customers.sql into two new intermediate
models. Before I approve, I want to verify that row counts haven't changed
across any of the affected models. Can you compare row counts for the models
modified in this PR?
```

Prompt 2 → `schema_diff` widget:
```
Now show me what columns changed — anything added, removed, or with a
different type — across the same modified models.
```

### Demo 2 — Investigate impact (Journey 2) · [Loom](https://www.loom.com/share/6f0a2691609946dbb5a7fac2da4a1328)

Blast-radius summary → evidence drilldown, with the agent following the `next_action` advisory hint. PR scenario: [jaffle_shop_golden#16](https://github.com/DataRecce/jaffle_shop_golden/pull/16) (`amount` DOUBLE → DECIMAL on `stg_payments`).

Prompt 1 → `impact_analysis` widget:
```
I'm reviewing a PR that changes the amount column type in stg_payments from
DOUBLE to DECIMAL(10,2). Can you show me the blast radius — which downstream
models are affected, and how serious the impact is?
```

Prompt 2 → `profile_diff` widget:
```
Following the next_action hint from the impact analysis, let me see the actual
statistical profile of the amount column in stg_payments — compare the mean,
distribution, null rate, and other stats between base and current to confirm
this is just a rounding effect within expected range.
```

> **Fixed in this PR (commit `f692f086`):** `impact_analysis`'s downstream `value_diff` step derived its per-column diff list from the CURRENT relation only (`get_model` → `get_columns(base=False)`), then applied `b."col" IS DISTINCT FROM c."col"` to **both** the base (`b`) and current (`c`) relations. When a column had drifted — present in current but not base — Snowflake raised a Binder Error (`invalid identifier 'B."full_name"'` / `Table "b" does not have a column named "full_name"`). The builder now diffs only the **intersection** of base ∩ current columns via `get_model(node_id, base=True)` (which manages its own warehouse connection), mirroring the canonical `ValueDiffTask`; drifted columns are reported via `schema_changes`, not fed into the diff expression, and value_diff is skipped when the PK itself has drifted. This was a pre-existing value_diff correctness bug surfaced by the demo, **not** introduced by the widget work. Verified by a new regression unit test (asserts drifted columns never reach the generated SQL) plus a read-only A/B against live Snowflake (old SQL shape → `000904 invalid identifier`; intersected shape → executes clean).

## What's in this PR

**New surface**:
- `recce mcp-widget-server` CLI subcommand (local mode only — iter 1 scope)
- `recce mcp-config-install` CLI helper — writes both server entries + `RECCE_MCP_WIDGETS=1` into `claude_desktop_config.json` (macOS, iter 1 scope), with `--dry-run` and `--yes`, and a backup of the existing config
- `recce/widget_server.py` — FastMCP server with 15 `@mcp.tool` delegates + matching `@mcp.resource` handlers, idiomatic shape (Pydantic input/output models, `CallToolResult` with short `content` + `structuredContent`, tool annotations, ext-apps SDK theme helpers)
- `recce/data/mcp/*.html` — 15 self-contained HTML widgets, each with exhaustive `@media (prefers-color-scheme: dark)` overrides and Claude design-token theming
- `WIDGET_TOOLS` filter in `recce/mcp_server.py` — when `RECCE_MCP_WIDGETS=1`, removes the 15 widget-eligible tools from the main server's `list_tools` (defers to widget server)

**15 widgets by tier**:

| Phase | Tier | Tools |
|-------|------|-------|
| A | 1–2 | `row_count_diff`, `schema_diff`, `get_server_info`, `list_checks`, `get_model` |
| B | 3 | `query`, `query_diff`, `value_diff`, `value_diff_detail`, `top_k_diff` |
| C | 4 (charts) | `histogram_diff`, `profile_diff` |
| D | 5 (mini-graph) | `get_cll`, `impact_analysis` |
| E | 6 (DAG, v1) | `lineage_diff` — hand-roll SVG, hard 10-node cap, graceful empty-state over cap |

**Bug fixes** (pre-existing, surfaced by the POC):
- `recce/track.py` `print(traceback.format_exc())` → `file=sys.stderr` — stdout traceback is fatal for any stdio MCP server when RecceConfig fails
- `recce mcp-server` / `mcp-widget-server` now `os.chdir(--project-dir)` before `RecceConfig`, so Claude Desktop's `cwd=/` spawn finds the dbt project + recce.yml (removes the bash-wrapper workaround)
- `SchemaChange` class-name collision — Phase D's `impact_analysis` defined a second module-level `SchemaChange` that shadowed `schema_diff`'s model, breaking `schema_diff` serialization with a Pydantic "column / change_status Field required" error. Renamed to `ColumnSchemaChange`; regression test pins both field surfaces.

**Quality / docs**:
- `tests/test_widget_server.py` — 35 tests (registration via public FastMCP API, env-var coordination 18-vs-20 tools, file-missing graceful degradation, `CallToolResult` shape, Pydantic structuredContent round-trip, annotations, collision guard, `lineage_diff` 10-node-cap branch)
- `tests/test_mcp_config_install.py` — 5 tests (project-dir validation, dry-run no-write, backup creation)
- `docs/mcp-widgets.md` — template doc for Scott's hand-off: file layout, config example, "Add widget N+1" walkthrough, structuredContent contract, gotchas, Python-vs-TypeScript SDK note

## Test plan

- [x] `pytest tests/test_widget_server.py` → 35 passed
- [x] `pytest tests/test_mcp_config_install.py` → 5 passed
- [x] `pytest tests/ -k "mcp or widget or schema_diff or impact or lineage"` → 363 passed, no widget-related regressions
- [x] Cwd=/ smoke (`recce mcp-widget-server --project-dir <dbt>`) — exit 0, clean stderr, no stdout pollution
- [x] Manual end-to-end in Claude Desktop — Demo 1 + Demo 2 recorded (Looms above); widgets render with real data, dark mode correct, plain prompts trigger correct tools
- [ ] Manual end-to-end in Cowork (parallel OAuth POC, DRC-3527 — not a blocker)

Note: `tests/test_server.py::test_spa_route_lineage` fails locally because the frontend SPA build artifact (`recce/data/lineage/index.html`) isn't present in this worktree — pre-existing, unrelated to this PR (fails identically with the diff stashed; resolved by `cd js && pnpm run build`).

## User-facing changes

**For Claude Desktop POC users**:
- Install config the easy way: `recce mcp-config-install --project-dir /path/to/dbt` (writes both entries + env var, backs up existing config). Or hand-edit per `docs/mcp-widgets.md`.
- `RECCE_MCP_WIDGETS=0`/unset → zero change, all 20 tools text-only from `recce mcp-server`, backward compatible.
- `RECCE_MCP_WIDGETS=1` → `recce mcp-server` serves 5 non-widget tools, `recce mcp-widget-server` serves the 15 widget tools. No name collision.
- Other MCP hosts (no MCP Apps support) → graceful degradation: host ignores `_meta.ui.resourceUri`, uses the tool's one-sentence `content` text.

## What's NOT in this PR

- Cloud session / snapshot support for `recce mcp-widget-server` (iter 1 = local mode only). **This is the strategic next step** — MCP ↔ Recce Cloud is what lets non-technical users skip git/dbt entirely and is where PR-selection + artifact-packaging are properly solved (not as extra local tools).
- Full FastMCP migration collapsing the two-server architecture (design Open Q #6; now warranted at 75% coverage)
- `lineage_diff` beyond v1 — progressive degradation for >10 nodes (shrink / zoom / aggregate) and the ReactFlow `@datarecce/ui/lineage` borrow contract
- Lifespan refactor (`_recce_server` still module-global)
- OAuth POC for Cowork distribution (Kent Huang DRC-3527, parallel)

## Reviewer notes

- `docs/mcp-widgets.md` is the canonical Scott hand-off doc — if the "Add widget N+1" walkthrough leaves a cold reader unsure, that's PR-blocking feedback.
- Commit history is deliberately preserved (not squashed) — each commit is one widget / fix / refactor step with meaningful diff boundaries; review in order.
- CSS follows project conventions: space-separated `rgb()`, no `@datarecce/ui` internal-path imports (widget HTML is plain JS).

---

## Review follow-up — deferred items (⚠️ this PR is NOT rushing to merge)

A full review pass (kc-pr-review + kc-pr-review-resolve) ran on `d089e737`. **Actionable fixes are committed** (`65ca0739`→`d089e737`): `~` path expansion + atomic config write in `mcp-config-install`, case-insensitive `RECCE_MCP_WIDGETS`, stdio-safe `--debug` traceback, robust `recce/data/**` gitignore (nested build artifacts now ignored; `recce/data/mcp/` source tracked), single-quote HTML escaping across all 15 widgets, `rel="noopener noreferrer"` on the PR link, **widget HTML load anchored on the `recce` package** (no longer depends on namespace-package resolution → reliable under `pip install`), and doc/test-count sync.

even-wei's NO-GO blocker (the `test_widget_server.py` `black`/`isort` format gate) was already resolved in `230af0f0` and is re-verified clean on the current head (`black`/`isort`/`flake8` exit 0).

The items below are **intentionally deferred** — this is an iter-1 POC whose goal is breadth (a widget for every tool), not hardened design. Each is marked with a `TODO(DRC-3526 follow-up)` comment in code; a DRC follow-up ticket will track them:

- **Widget tools bypass the central `call_tool` error classifier/telemetry (DRC-2754).** Errors still surface (FastMCP wraps them); what's deferred is expected-error classification (`table_not_found`/`permission_denied` → warning) + Sentry metrics for the 15 widget tools. Resolve when the two-server architecture is consolidated. — `recce/widget_server.py`
- **`impact_analysis` value_diff drift fix lacks a real-DuckDB e2e.** Covered by a mock test (drifted column absent from generated SQL) + a manual Snowflake A/B; a real-adapter e2e with a PK + a column present only in `current` is the gap. — `recce/mcp_server.py`
- **`schema_diff` has no single-env warning.** The empty-state ("No models found") can read as a clean diff when there is simply no base to compare against. — `recce/widget_server.py`
- **Type contracts use free `str` for fixed-set fields** (`status` / `change_status` / `priority`) — batch a `Literal`/submodel tightening pass. — `recce/widget_server.py`
- **`impact_analysis` doubles base-side warehouse introspection** (the cost of the binder-error correctness fix) — consider caching/batching `get_model(base=True)`. — even-wei N3
- **15 copy-pasted `escapeHtml`/`svgEl` helpers** — extract a shared asset to keep the escaping contract single-sourced. — even-wei N5
- **`check_base_freshness` SHA-freshness is opt-in with no production caller** (the `stale_sha`/`expected_base_sha` path is a tested opt-in API, not pure dead code). Decide in a focused PR: leave as-is, wire a caller, or remove SHA-freshness — it touches the unrelated `recce check-base` command + tests, so it is out of scope here. — `recce/cli.py`

**Pre-merge check:** run a packaged-wheel smoke test (`pip install` the built wheel, start `recce mcp-widget-server`, load one widget) to confirm `recce/data/mcp/` ships and resolves under the installed layout.


even-wei:
# Code Review: PR #1397

**SHA** `87f22fd3` · **Verdict** NO-GO

Verified checks: `flake8` clean (exit 0); targeted suites **156 passed** (`test_widget_server.py` + `test_mcp_config_install.py` + `test_mcp_server.py`) under the project-pinned `mcp 1.27.0` / Python 3.13; `black --check` and `isort --check` FAIL on `tests/test_widget_server.py`; bundled security-review found no exploitable Critical/High.

## Blockers
None.

## Issues
1. `tests/test_widget_server.py` — Fails the project's required `make format` gate (`black` + `isort`).
   Evidence: `black --check` → "1 file would be reformatted" (line ~2218 over-length call); `isort --check` → import block at line ~2248 needs multi-line wrap. AGENTS.md lists `make format` as a pre-commit/CI requirement.
   Pass E / project-convention.

## Notes
1. `recce/cli.py:3174` (`mcp_config_install`) — On re-run the backup is overwritten with the already-modified config, so `.recce.bak` no longer holds the pristine pre-recce original; only the recce entries differ, and re-run is otherwise idempotent (deterministic `new_entries`). Consider skip-if-exists or timestamped backups.
   Pass D.
2. `recce/data/mcp/*.html` — Shared `escapeHtml` escapes `& < > "` but not `'` (`&#x27;`). Not exploitable today (all interpolated attributes are double-quoted and `"` is escaped; SVG labels use `textContent`). Defense-in-depth only; add `.replace(/'/g,"&#x27;")` if a single-quoted attribute is ever introduced.
   Pass B (Security, LOW).
3. `recce/mcp_server.py:1885` (`impact_analysis`) — New `get_model(node_id, base=True)` adds a second per-model warehouse introspection round-trip inside the impacted-models loop. Correctness fix (avoids binder failure on drifted columns) but doubles base-side introspection on large impacted sets.
   Pass G.
4. `recce/cli.py:3012` (`mcp_config_install`) — `recce mcp-config-install` and the macOS-only path are correctly OS-guarded, but there is no `uninstall` / removal path; the docstring's "To undo: replace with .recce.bak" is the only recovery, which is lost after a second run (see Note 1). Untested uninstall for a config-mutating installer.
   Pass E.
5. `recce/data/mcp/` — 15 hand-written HTML/JS widgets (~8.4k LOC) carry a copy-pasted `escapeHtml` + `svgEl` helper per file. For a POC this is acceptable, but it's 15 sources of truth for the escaping contract; a shared asset would prevent Note 2 drifting per-file.
   Pass I (architecture, copy-paste).

## Limits
- `uv run pytest` (the default dev path) resolves to a STALE local env (Python 3.10 + an older `mcp` whose `FastMCP.resource()` lacks the `meta` kwarg), yielding 32 spurious `TypeError: ... unexpected keyword argument 'meta'` failures. These are an environment artifact, NOT a code defect — under the pinned `mcp>=1.23.0` (resolved to 1.27.0, where `resource()` DOES accept `meta`) all 156 tests pass. Verified by forcing the 3.13 venv. CI using the locked deps will pass; a developer with a stale global `mcp` will see false failures.
- Did not exercise the widgets live in Claude Desktop (macOS MCP-app rendering) — XSS assessment is static data-flow only. SQL `query`/`query_diff` tool args flow to the pre-existing `_tool_query` path (unchanged by this PR), not re-reviewed for injection here.

even-wei (changes requested):
Claude Code Review: NO-GO. 1 issue (format gate) + 5 notes. See review comment.

.gitignore
```diff
@@ -16,14 +16,36 @@ build
...
-# Ignore build artifacts from frontend
-recce/data
+# Ignore build artifacts from frontend.
+# recce/data/mcp/*.html are SOURCE files (widget HTML for MCP Apps) — not build output.
+# git negation cannot un-ignore files under an ignored directory, so we ignore
+# specific subdirs instead of the whole recce/data tree.
+recce/data/.gitkeep
+recce/data/_next/
+recce/data/chunks/
+recce/data/*.html
```
iamcxa: **MEDIUM** — Enumerated patterns regress the old blanket `recce/data` ignore for *nested* build artifacts. `recce/data/*.html` matches files directly under `recce/data/` only — it does **not** match nested SPA output like `recce/data/lineage/index.html` (the PR's own test-plan note confirms the frontend build emits that file). After `pnpm run build`, those nested artifacts are untracked-but-not-ignored and can be accidentally committed — the exact thing AGENTS.md's "do NOT commit recce/data" rule guards against.

More robust (ignore the dir's contents, re-include the source subdir):
```
recce/data/**
!recce/data/mcp/
!recce/data/mcp/**
```
The inline comment claims git negation can't un-ignore under an ignored dir — that's only true when the *directory itself* is ignored. Ignoring the **contents** (`recce/data/**`) and negating `recce/data/mcp/` works.

recce/cli.py
```diff
@@ -3072,24 +3411,23 @@ def check_base_freshness(
...
         base_sha = manifest_data.get("metadata", {}).get("env", {}).get("DBT_GIT_SHA")
         result["base_sha"] = base_sha
 
-        if base_sha is not None:
-            from recce.git import current_commit_hash
-
-            current_sha = current_commit_hash()
-            result["current_sha"] = current_sha
-            if current_sha and base_sha != current_sha:
+        if base_sha is not None and expected_base_sha:
```
iamcxa, line 3432: **MEDIUM** — SHA freshness is now opt-in (`and expected_base_sha`), but neither shipping caller passes it: `check_base` CLI (L3488) and the MCP startup check (`mcp_server.py:2807`) both omit `expected_base_sha`. So `stale_sha` is now **unreachable in production**, `current_sha` is always `None`, and the `stale_sha` color-map (L3505) + exit-code-2 branch (L3522) are dead.

The rationale (base artifacts come from the base branch → base SHA ≠ feature HEAD → the old auto-check was a false-positive generator) is sound. But this is a behavior change to the existing `recce check-base` command, not mentioned in the PR body and bundled into a widget POC. Please confirm it's intentional — if so, clean up the dead `stale_sha`/`current_sha` path, or wire a caller that actually passes `expected_base_sha`.

recce/widget_server.py
```diff
@@ -0,0 +1,2543 @@
...
+        - Schema (column) changes — use schema_diff instead
+        - SQL output comparison — use query_diff
+        - Single environment without target-base — server warns about
+          single-env mode but returns no useful comparison
+
+    Error Handling:
+        - table_not_found / permission_denied surface in *_meta.status
+        - tool raises only on fundamental dbt/adapter failures
+    """
+    result = await _recce_server._tool_row_count_diff(args.model_dump(exclude_none=True))
```
iamcxa, line 222: **MEDIUM** — Widget tools call `_recce_server._tool_*()` directly, bypassing the central `@server.call_tool` handler in `mcp_server.py:1426`. That handler is where DRC-2754 error classification lives: `_classify_db_error` → downgrade `table_not_found`/`permission_denied` to warnings + `sentry_metrics.count("mcp.expected_error", ...)`. For all 15 widget tools that machinery never runs — expected DB errors become hard errors and there's zero telemetry.

Not a silent failure (FastMCP still wraps the raised exception into an error result), so not blocking for the POC — but it quietly undoes documented investment. Suggest routing widget delegates through a shared classify/telemetry wrapper, or a tracking note that the widget path defers DRC-2754 parity.

recce/mcp_server.py
```diff
@@ -1844,11 +1917,27 @@ async def _tool_impact_analysis(self, arguments: Dict[str, Any]) -> Dict[str, An
...
-                    non_pk_cols = [c for c in columns_info if c != pk]
+
+                    # Restrict the diff to columns present in BOTH the base and current
+                    # relations. The per-column expression `b."col" IS DISTINCT FROM c."col"`
+                    # references the same column on both sides, so a column that has drifted
+                    # (exists in one relation but not the other) fails the warehouse binder.
+                    # Intersect both sides, mirroring ValueDiffTask; drifted columns are
+                    # already reported via schema_changes (Step 2b). get_model(base=True)
+                    # manages its own warehouse connection for the base-side introspection.
+                    base_model_info = self.context.adapter.get_model(node_id, base=True)
```
iamcxa, line 1938: **MEDIUM (test gap)** — This drift fix is correct (intersection + `get_model(base=True)` connection-managed introspection, matching `ValueDiffTask`). But it has **no real-adapter e2e**. `test_mcp_e2e.py::test_schema_changes_detected` builds the drift shape against real DuckDB, yet its `users` fixture declares no `primary_key`, so `if not pk: continue` (L1917) skips value_diff entirely and the `IS DISTINCT FROM` binder path is never exercised on a live warehouse. Coverage is mock-only (drifted column absent from generated SQL) + the documented manual Snowflake A/B.

Per project MEMORY, mocks can't catch the missing-warehouse-connection class of bug. Recommend an e2e fixture with a PK where `current` has a column absent in `base`; assert `impact_analysis` returns value_diff over the common non-PK columns and does not raise.

iamcxa (commented):
@iamcxa — automated review (kc-pr-review, Full tier: code-reviewer + silent-failure-hunter + pr-test-analyzer + type-design-analyzer + HTML-XSS scan + ToB security). Event is **COMMENT**, not approve: (a) tests could not be re-run locally (env SIGKILL/OOM at pytest import — the first background run did exit 0, and the PR body documents 35+5 passing); (b) the `check_base_freshness` behavior change needs author intent; (c) the one production-risk path (Snowflake binder in the value_diff fix) is mock + manual-A/B only, no CI e2e. No CRITICAL/HIGH blocking bugs were found.

## Goal-achievement verdict: ✅ Achieved (POC scope, with follow-up caveats)

**Evidence**
- All 15 tool/resource pairs register correctly; every `_meta.ui.resourceUri` maps to a real `recce/data/mcp/*.html` (verified file-by-file).
- All 15 HTML widgets escape warehouse data consistently (`escapeHtml` / `textContent` / numeric formatters) — **no exploitable stored-XSS sink found**.
- The `impact_analysis` value_diff drift fix is correct — it intersects base ∩ current columns and uses `get_model(base=True)` connection-managed introspection, matching the canonical `ValueDiffTask`.
- Type design is solid for a POC (None-vs-zero distinctions in `RowCountModel`, `ValueDiffColumnRow.matched_p`, `HistogramEnvStats.total` are deliberate and correct).

## Verification Matrix

| # | Concern (from PR body) | Verified? | Evidence |
|---|---|---|---|
| V1 | `RECCE_MCP_WIDGETS=0`/unset → all 20 tools text-only, backward compatible | ✅ | `_widgets_enabled()` reads env at call time; filter only removes WIDGET_TOOLS when enabled (`mcp_server.py:1294`) |
| V2 | `RECCE_MCP_WIDGETS=1` → main 5 + widget 15, no collision | ✅ | `WIDGET_TOOLS` filter + widget server registers exactly the 15-name set; test pins it (`test_widget_server.py:102/173`) |
| V3 | impact_analysis drift fix (Snowflake binder) | ⚠️ | Code correct, but **e2e gap** — see inline on `mcp_server.py:1931` |
| V4 | schema_diff serialization (SchemaChange collision fix) | ⚠️ | Models distinct + field-pinned; **no behavioral output test** — see Advisory F |
| V5 | stdio stdout purity (JSON-RPC channel) | ✅ | No stdout writes in widget_server.py; logging → stderr (`widget_server.py:2521`) |
| V6 | chdir to `--project-dir` before RecceConfig | ✅ | Present in both subcommands; minor: no dir-existence guard (Advisory) |

## Verification Summary (tests)

| Check | Result |
|---|---|
| Local re-execution | ⚠️ Could not run — env SIGKILL/OOM at pytest import (0-byte output; killed during collection). Not a test-logic failure. |
| First background run (widget + config_install + check_base + mcp_server + mcp_cloud_backend) | exit 0 |
| PR-body documented | `test_widget_server.py` 35 passed · `test_mcp_config_install.py` 5 (file actually has 7) · broad selection 363 passed |

## Advisory (not blocking)

- **A — Bundled tooling / scope creep.** `CLAUDE.md` (+37 gbrain boilerplate, includes a per-machine "on this machine" assertion in shared docs) and `.claude/settings.json` (enables `mcp-server-dev` for **all** repo contributors) are unrelated to the widget feature. Suggest splitting into a separate commit/PR, marking POC-only, or moving dev-plugin prefs to gitignored `settings.local.json`.
- **B — External SDK dependency.** All 15 widgets `import` `https://unpkg.com/@modelcontextprotocol/ext-apps@0.4.0` (no SRI). Mitigated by each resource's `csp.resourceDomains: ["https://unpkg.com"]` (host-enforced origin restriction) and this is the standard MCP Apps SDK load pattern — acceptable for POC; consider local vendoring before productionizing.
- **C — Type contracts.** `status` / `change_status` / `priority` etc. are free `str` with legal values only in comments. Converting the widget-branched ones to `Literal[...]` is a zero-cost contract win.
- **D — HTML hardening.** No `<meta>` CSP in the HTML (partially mitigated by `resourceDomains`); `get_server_info.html:317` `pr.url` href has no scheme validation (`javascript:` possible, though source is GitHub metadata not warehouse) and `target="_blank"` lacks `rel="noopener"`.
- **E — Mega-PR.** Feature POC + 4 unrelated pre-existing bugfixes (track stderr / chdir / set_backend cache-key / freshness) + tooling, 37 commits / +14K. Hard to review and to revert. Future: land pre-existing bugfixes as their own PRs.
- **F — Test gaps.** `schema_diff` is the only widget with no behavioral `model_validate` round-trip test (the collision broke exactly that serialization); no absolute tool-count assertion (`assert len==20/5`) so dropping a non-widget tool would pass silently.
- **G — impact_analysis per-step `except Exception`** (pre-existing, `mcp_server.py` ~1866/1901/2042, not in this diff) appends to `errors[]` without logging unexpected exceptions — a code bug is indistinguishable from a warehouse failure. Worth narrowing + `exc_info` logging when that area is next touched.

Nice work on the POC — the escaping discipline, type care, and the documented drift fix are genuinely solid. The items above are hardening + a couple of bundled changes worth confirming.

recce/cli.py
```diff
@@ -2908,6 +2914,336 @@ def mcp_server(state_file, sse, host, port, **kwargs):
...
+    backup_path = config_path.with_suffix(config_path.suffix + ".recce.bak")
+    backup_created = not backup_path.exists()
+    if backup_created:
+        shutil.copy2(str(config_path), str(backup_path))
+
+    # ------------------------------------------------------------------
+    # Merge entries and write
+    # ------------------------------------------------------------------
+    existing_config["mcpServers"].update(new_entries)
+    config_path.write_text(json.dumps(existing_config, indent=2), encoding="utf-8")
```
iamcxa: **MEDIUM** — Non-atomic config write. `write_text` truncates-then-writes in place; a crash mid-write corrupts `claude_desktop_config.json`. The backup is created only on first run (L3218 `backup_created = not backup_path.exists()`), so a mid-write failure on a **re-run** leaves a corrupt config with no fresh backup. There's also no `except` around the write, so an `OSError`/`PermissionError` surfaces as a raw traceback that never points the user at `.recce.bak`.

Recommend write-to-temp + `os.replace()` (atomic rename) plus a guarded `except OSError` that references the backup path.

recce/widget_server.py
```diff
@@ -0,0 +1,2543 @@
...
+    Iter 1 is LOCAL MODE ONLY — cloud/session kwargs are not supported.
+    Register both `recce mcp-server` and `recce mcp-widget-server` entries
+    in Claude Desktop config with RECCE_MCP_WIDGETS=1 set on both.
+
+    mcp.run(transport="stdio") manages its own asyncio event loop internally.
+    Do NOT wrap this function in asyncio.run().
+    """
+    global _recce_server
+
+    if kwargs.get("cloud") or kwargs.get("session"):
```
iamcxa: **LOW** — This guard checks `kwargs.get("session")`, but `--session` binds to dest `cloud_session` (`cli.py:2702`), so the branch never fires. Currently masked because the CLI entry guards `cloud_session` (L2974) and sets `kwargs["cloud"] = False` before calling here — but `run_widget_server` is a reusable entry point and the guard is silently ineffective. Align to `kwargs.get("cloud_session")`.

recce/widget_server.py
```diff
@@ -0,0 +1,2543 @@
...
+        - empty models dict means no schema changes detected in the selected scope
+    """
+    lineage_diff = _recce_server.context.get_lineage_diff().model_dump(mode="json")
+    rich_result = _recce_server._compute_schema_changes(
+        lineage_diff,
+        select=args.select,
+        exclude=args.exclude,
+        packages=args.packages if args.packages is not None else None,
+    )
+    output = SchemaDiffOutput(
```
iamcxa, line 307: **LOW** — `SchemaDiffOutput` has no `warning` field and schema_diff never extracts `_warning`, unlike the other 13 diff tools (e.g. L202 `warning = result.pop("_warning", None)`). In single-env mode it renders "No models found" (`schema_diff.html:264`), which reads as "no schema changes" when the truth is "no base to compare against". (`_tool_schema_diff` also lacks the single-env warning on the main server — pre-existing — but the widget's empty-state makes it more misleading.) Suggest adding a `warning` field + the single-env notice.

recce/widget_server.py
```diff
@@ -0,0 +1,2543 @@
...
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _read_widget_html(name: str) -> str:
+    """Read widget HTML from recce/data/mcp/{name}.html, returning an error stub if missing."""
+    try:
+        ref = importlib.resources.files("recce.data.mcp") / f"{name}.html"
+        return ref.read_text(encoding="utf-8")
+    except (FileNotFoundError, TypeError, ModuleNotFoundError):
```
iamcxa: **LOW** — Missing widget asset degrades to an in-band HTML stub with no `logger.error`/`warning`, so a packaging defect that drops `recce/data/mcp/*.html` ships silently — visible only if a human opens the rendered widget. Add a log line before returning the stub. Minor: the stub says "Run pnpm run build" but these are tracked **source** files (per the .gitignore comment), not build output — the remedy text is misleading. Also consider broadening the `except` (importlib `read_text` can raise `OSError`/`PermissionError`/`UnicodeDecodeError`, which currently propagate instead of degrading).

recce/widget_server.py
```diff
@@ -0,0 +1,315 @@
...
+        return f"<html><body>Widget asset missing: {name}.html. Run pnpm run build.</body></html>"
```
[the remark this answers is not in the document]
iamcxa: Verified — this is not a hard failure. `recce.data.mcp` resolves as a PEP 420 namespace package (Python 3.10+, the project floor); namespace packages do not require `__init__.py`. Evidence: the 35 widget-server tests pass under the pinned `mcp` / Py3.13 env (including resource registration), and the recorded Claude Desktop demos render real widget HTML — both exercise this load path.

As defense-in-depth, `_read_widget_html` now broadens its `except` and logs on any failure (`dbde3247`), so if the path ever does fail (e.g. a wheel that drops `recce/data/mcp/`) it degrades to a stub + a server-log error rather than crashing.

Residual: worth a packaged-wheel smoke test (install the wheel, load one widget) before GA to confirm the data dir ships and resolves under the installed layout.

recce/mcp_server.py
```diff
@@ -1504,34 +1533,59 @@ async def _tool_schema_diff(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
...
+            ]
+            removed = [
+                {"name": col, "type": base_columns[col].get("type", "")}
+                for col in sorted(base_col_names - current_col_names)
+            ]
+            type_changed = []
+            unchanged_count = 0
+            for col in sorted(base_col_names & current_col_names):
+                base_type = base_columns[col].get("type")
+                curr_type = current_columns[col].get("type")
```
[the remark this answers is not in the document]
iamcxa: Acknowledged — minor. `_compute_schema_changes` can emit `base_type` / `curr_type` as `None` (from `.get("type")`), and the widget already tolerates it (`c.type ?? ""`). Tightening the `SchemaChange` annotation to the real shape (`{name,type}` / `{name,base_type,curr_type}` submodels, or `Optional`) is batched with the other type-contract polish (the `Literal`/submodel pass) — deferred for the POC, tracked in the follow-up discussion.

docs/mcp-widgets.md
```diff
@@ -0,0 +1,431 @@
...
+                             # WIDGET_TOOLS set + _widgets_enabled() filter live here.
+  widget_server.py           # FastMCP widget server (iter 1).
+                             # @mcp.tool delegates + @mcp.resource handlers.
+  cli.py                     # mcp-widget-server CLI subcommand added here.
+  data/
+    mcp/                     # Widget HTML asset directory (gitignored via per-extension
+      row_count_diff.html    # allowlist — see .gitignore). Self-contained HTML files.
+      schema_diff.html
+tests/
+  test_widget_server.py      # 5 tests covering WIDGET_TOOLS coordination + widget server.
```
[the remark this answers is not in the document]
iamcxa: Fixed in `c533be93` — widget test count corrected to 35 (the actual `async def test_` count in `test_widget_server.py`).

docs/mcp-widgets.md
```diff
@@ -0,0 +1,431 @@
...
+  `asyncio.run()`.** `mcp.run()` manages its own asyncio event loop internally.
+  Wrapping it raises `ValueError: a coroutine was expected, got None` (fixed in
+  Day 1 cycle 1; see commit `bb6f1261`).
+
+- **`recce/data/` is gitignored as build output.** Widget HTML files use a
+  per-extension allowlist in `.gitignore` to escape the broad `recce/data`
+  ignore rule. The allowlist currently covers `*.html`. If you add new file
+  types (`.css`, `.svg`, `.js`) under `recce/data/mcp/`, check `.gitignore`
+  and add an allowlist entry if needed; otherwise `git add` will silently skip
+  your file.
```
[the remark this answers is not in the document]
iamcxa: Fixed in `c533be93` — the Gotchas section now describes the `recce/data/**` + `!recce/data/mcp/` negation scheme; any file type under `recce/data/mcp/` is tracked automatically (no per-extension allowlist to maintain).

recce/track.py
```diff
@@ -58,7 +58,7 @@ def _show_error_message(self, msg, params):
         if params.get("debug"):
             console.print_exception(show_locals=True)
         else:
-            print(traceback.format_exc())
+            print(traceback.format_exc(), file=sys.stderr)
```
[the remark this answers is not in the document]
iamcxa: Fixed in `fcd7a6c3` — `_show_error_message` now uses `Console(stderr=True)`, so the `--debug` `print_exception` path no longer writes to the stdio JSON-RPC channel. (Thread is marked outdated, but the debug-path stdout write was still present until this commit — only the non-debug branch had been fixed.)

recce/cli.py
```diff
@@ -2908,6 +2914,336 @@ def mcp_server(state_file, sse, host, port, **kwargs):
...
+            "[[red]Error[/red]] mcp-config-install currently supports macOS only (iter 1 scope).\n"
+            "Manual config: edit your MCP client config to add both `recce` and `recce-widgets`\n"
+            "entries with RECCE_MCP_WIDGETS=1. See docs/mcp-widgets.md for the JSON snippet."
+        )
+        exit(1)
+
+    # ------------------------------------------------------------------
+    # Validate project dir
+    # ------------------------------------------------------------------
+    project_dir_path = Path(project_dir).resolve()
```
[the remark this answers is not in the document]
iamcxa: Fixed in `65ca0739` — `--project-dir` now uses `Path(project_dir).expanduser().resolve()`, so `~/my_dbt_project` expands correctly.

recce/cli.py
```diff
@@ -2908,6 +2914,336 @@ def mcp_server(state_file, sse, host, port, **kwargs):
...
+            "Example: recce mcp-config-install --project-dir /path/to/my_dbt_project"
+        )
+        exit(1)
+
+    # ------------------------------------------------------------------
+    # Resolve Claude Desktop config path
+    # ------------------------------------------------------------------
+    if claude_config:
+        config_path = Path(claude_config).resolve()
+    else:
```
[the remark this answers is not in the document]
iamcxa: Fixed in `65ca0739` — `--config` now uses `Path(claude_config).expanduser().resolve()`, same `~` expansion fix as `--project-dir`.

recce/mcp_server.py
```diff
@@ -50,6 +50,34 @@
...
+    "impact_analysis",
+    "lineage_diff",
+}
+
+DEFAULT_CLOUD_REQUEST_TIMEOUT = 30
+
+
+def _widgets_enabled() -> bool:
+    """Read RECCE_MCP_WIDGETS env at call time (not import time) so tests can monkeypatch."""
+    return os.environ.get("RECCE_MCP_WIDGETS", "").strip() in ("1", "true", "True")
```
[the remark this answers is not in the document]
iamcxa: Fixed in `dbde3247` — `_widgets_enabled()` now compares `.strip().lower()` against `("1", "true")`, so `TRUE` / `True` / `true` / `1` all work.

recce/data/mcp/get_server_info.html
```diff
@@ -0,0 +1,349 @@
...
+            <div class="kv-value kv-mono">${escapeHtml(g.base_sha.slice(0, 8))}</div>`);
+        }
+      }
+
+      // pull request
+      if (data.pull_request && data.pull_request.title) {
+        const pr = data.pull_request;
+        const prLink = pr.url
+          ? `<a href="${escapeHtml(pr.url)}" target="_blank">${escapeHtml(pr.title)}</a>`
+          : escapeHtml(pr.title);
```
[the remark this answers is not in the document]
iamcxa: Fixed in `c6d3506e` — added `rel="noopener noreferrer"` to the `target="_blank"` PR-link anchor.

tests/test_mcp_config_install.py
```diff
@@ -0,0 +1,284 @@
+"""Tests for `recce mcp-config-install` CLI subcommand.
+
+Covers:
+- Writes both recce and recce-widgets MCP entries with RECCE_MCP_WIDGETS=1
+- Preserves existing MCP server entries (other servers untouched)
+- Validates --project-dir (must contain dbt_project.yml)
+- Dry-run mode does not write to disk
+- Backup file is created before writing
+"""
```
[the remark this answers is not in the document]
iamcxa: Fixed in `c533be93` — the module docstring `Covers:` list now includes the backup-pristine-on-rerun and `python -m recce.cli` fallback tests (7 total).

.gitignore
iamcxa: Fixed in `c533be93` — switched to `recce/data/**` + `!recce/data/mcp/` negation. Verified with `git check-ignore`: `recce/data/mcp/*.html` and `.gitkeep` are tracked; `recce/data/lineage/index.html`, `recce/data/_next/`, and `recce/data/index.html` are ignored.

recce/cli.py
iamcxa: Deferred to the follow-up discussion (POC scope) — leaving this thread open as a tracked TODO. The opt-in SHA change is intentional behavior-wise (base artifacts come from the base branch, so the base SHA always differs from feature HEAD → the old auto-check was a false-positive generator). But `stale_sha` / `current_sha` / the exit-code-2 branch are now dead for both callers. Decision pending: clean up the dead path vs. wire a caller that passes `expected_base_sha`.

recce/widget_server.py
iamcxa: Deferred to the follow-up discussion (POC scope) — leaving open as a tracked TODO. The widget delegates bypass the central `call_tool` classifier/telemetry (DRC-2754). Errors still surface (FastMCP wraps them); what's lost is expected-error classification + Sentry metrics for the 15 widget tools. Tracking a shared classify/telemetry wrapper for when the two-server architecture is consolidated.

recce/mcp_server.py
iamcxa: Deferred to the follow-up discussion — leaving open as a tracked test-coverage TODO. The fix is mock-tested (asserts drifted columns never reach the generated SQL) plus the documented manual Snowflake A/B, and even-wei independently verified 156 tests pass. A real-DuckDB e2e with a PK + a column present only in `current` is the gap to close; not a blocker for the POC.

recce/cli.py
iamcxa: Fixed in `65ca0739` — config write is now temp-file + `os.replace()` (atomic), guarded by `except OSError` with temp cleanup. A mid-write failure leaves the existing config intact and points the user at the backup.

recce/widget_server.py
iamcxa: Fixed in `dbde3247` — the guard now checks `cloud_session` (the actual `--session` dest) in addition to `session`, so it is no longer silently dead.

recce/widget_server.py
iamcxa: Deferred to the follow-up discussion (POC scope) — leaving open as a tracked TODO. Likely fix: add a `warning` field to `SchemaDiffOutput` + extract `_warning`, plus the single-env notice, so the empty-state ("No models found") isn't read as a clean diff when there's simply no base to compare against.

recce/widget_server.py
iamcxa: Fixed in `dbde3247` — broadened the `except` and added `logger.error` on load failure, so a packaging defect that drops `recce/data/mcp/*.html` is visible in server logs. Stub message updated (these are tracked source files, so a miss means a broken install/package, not a missing frontend build).

iamcxa:
@even-wei — addressing your NO-GO (reviewed at `87f22fd3`; current head is `c533be93`, which has 3 commits after your review's SHA + this resolve pass):

**Blocker (format gate)** — already resolved before the current head: `230af0f0` reformatted `tests/test_widget_server.py`. Verified now on `c533be93`: `black --check`, `isort --check-only`, and `flake8` all exit 0.

**Note 1 / 4 (backup clobbered on re-run, undo lost)** — fixed in `eeb5002d`: the backup is skip-if-exists, so `.recce.bak` keeps the pristine pre-recce original across re-runs. Atomic write also added (`65ca0739`): a mid-write failure now leaves the existing config intact.

**Note 2 (escapeHtml doesn't escape `'`)** — fixed in `c6d3506e`: all 15 widgets now escape `'` → `&#x27;`.

**Note 3 (double base-side introspection in impact_analysis)** — acknowledged; this is the tradeoff for the binder-error correctness fix. Tracking a cache/batch of the base-side `get_model` as a follow-up.

**Note 5 (15 copy-pasted escapeHtml/svgEl helpers)** — acknowledged; the POC accepts per-file helpers. Shared-asset extraction tracked as a follow-up.

Also note your "Limits": the `uv run pytest` stale-env failures (older `mcp` lacking the `resource(meta=...)` kwarg) are an environment artifact — confirmed; under the pinned `mcp>=1.23.0` all suites pass, which your forced-3.13 run verified (156 passed).

Re-requesting your review on `c533be93`.

recce/widget_server.py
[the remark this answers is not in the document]
iamcxa: Hardened in `d089e737` — `_read_widget_html` now anchors on `importlib.resources.files("recce")` (a regular package, always importable) and traverses to `data/mcp/{name}.html`, instead of importing `recce.data.mcp` as a namespace package. This removes the install-layout dependency you flagged: a `pip install recce` wheel serves the widgets the same as an editable/source checkout. Combined with the broadened `except` + `logger.error` (`dbde3247`), the load path is now both robust and observable. Resolving.

recce/cli.py
iamcxa: On reflection, holding this rather than cleaning it up inside this widget PoC. The `stale_sha` / `expected_base_sha` path is not pure dead code — it's a *tested opt-in API*: `test_status_stale_sha_when_expected_base_sha_mismatch` exercises the mismatch → `stale_sha` branch, and `test_status_stale_sha` pins the no-arg = `fresh` behavior. So it's "no production caller yet", not "unreachable".

Removing it would touch the `recce check-base` command (color-map, exit-code-2, docstring) plus those tests — a cross-feature refactor outside this widget PoC's scope, and one I can't validate locally right now (the full pytest suite OOMs in this env; only the lint tools run). Proposal: either leave the opt-in as-is (harmless + tested), or do a focused follow-up that wires a caller passing `expected_base_sha` (giving it a purpose) or removes SHA-freshness entirely. Leaving this thread open for that decision.

recce/widget_server.py
iamcxa: TODO marker added in `d080819e` (above `_recce_server`) and tracked in the PR description's "Review follow-up — deferred items" section. Deferred per iter-1 POC scope; leaving open until the shared classify/telemetry wrapper lands (two-server consolidation).

recce/mcp_server.py
iamcxa: TODO marker added in `d080819e` (at the value_diff intersection) and tracked in the PR description's deferred section. Deferred; leaving open until the real-DuckDB e2e fixture (PK + drifted column) lands.

recce/widget_server.py
iamcxa: TODO marker added in `d080819e` (at the `SchemaDiffOutput` build) and tracked in the PR description's deferred section. Deferred; leaving open until the `warning` field + single-env notice are added.

even-wei:
# Code Review: PR #1397

**SHA** `6fb3bff` · **Verdict** NO-GO

## Blockers

1. `recce/widget_server.py:84` — `PullRequestInfo.id: Optional[str]` rejects the `int` id the state actually carries, so `get_server_info` crashes whenever PR metadata is present.
   Evidence: `recce/pull_request.py:14` is `Optional[Union[int, str]]` populated from `pr.number` (int); reproduced in this venv — `ServerInfoOutput(..., pull_request={'id': 15, ...})` → `ValidationError: pull_request.id Input should be a valid string`. Tests only cover `pull_request is None` (`tests/test_widget_server.py:427`).
   Pass C.

2. `recce/cli.py:3132-3135` — `--dry-run` (and a declined confirm prompt) still creates `claude_desktop_config.json` + parent dirs when the file doesn't exist, violating the documented "without writing" contract.
   Evidence: the skeleton `write_text` runs before the `if dry_run:` check at `cli.py:3199`; reproduced — `mcp-config-install --dry-run` with a nonexistent `--config` exits 0 and leaves the file on disk. The dry-run test pre-creates the file so the suite can't catch it (`tests/test_mcp_config_install.py:166`).
   Pass A.

## Issues

1. `recce/widget_server.py` (all 15 tools) — single `args: <Model>` parameter makes FastMCP emit a nested inputSchema (`properties.args.$ref`) while every docstring documents flat parameters and the same-named tools on `recce mcp-server` use flat schemas.
   Evidence: nested `{"properties": {"args": {...}}, "required": ["args"]}` reproduced with `mcp.server.fastmcp` in this venv — description ≠ schema breaks the repo's "MCP tool description = LLM agent contract" rule, and flipping `RECCE_MCP_WIDGETS` silently changes the calling convention.
   Pass C.

2. `recce/widget_server.py:692-699` (also `:1162-1170`) — `query`/`query_diff` declare `readOnlyHint: True, destructiveHint: False` but execute the raw `sql_template` (a `DELETE` runs); hosts use these hints to skip confirmations.
   Evidence: `QueryTask` → `execute_sql_with_limit` (`recce/tasks/query.py:43`) passes the statement straight to `adapter.execute`; relatedly `tests/test_widget_server.py:336` pins `openWorldHint: False` for `row_count_diff`, which runs `COUNT(*)` against the warehouse.
   Pass B.

3. `recce/data/mcp/get_server_info.html:317` — PR link `href` has no scheme allowlist; a `javascript:` URL executes on click (escapeHtml doesn't neutralize schemes).
   Evidence: `` `<a href="${escapeHtml(pr.url)}" target="_blank" ...>` `` — only href-from-data sink in the PR; gate on `http(s):`.
   Pass B (MEDIUM — `pr.url` originates from state/PR metadata, not direct user input).

4. `recce/data/mcp/impact_analysis.html:546-561` — the mini-graph fabricates edges: every modified model is connected to every downstream model (cartesian product), visually asserting lineage relationships that may not exist.
   Evidence: `for (const srcName of modifiedNames) { ... for (const dstName of downstreamNames) { ... appendChild(path)` (comment admits "simplified").
   Pass A.

5. `recce/data/mcp/query_diff.html:424-436` / `value_diff_detail.html:384-397` — `in_a`/`in_b` are located by `key || name` but filtered out by `name` only → header/data column misalignment when they diverge; and when the columns are absent (`inAIdx === -1`) every row is labeled "Removed".
   Evidence: `findIndex(c => c.key === "in_a" || c.name === "in_a")` vs `filter(c => c.name !== "in_a")`; `const status = (!inB) ? "removed" : "added"` with `inB=false` fallback.
   Pass A.

6. `recce/data/mcp/value_diff.html:401,426` — "All values match" banner renders when every `matched_p` is null (false-quiet, against the repo's false-loud principle), and the derived mismatched-row count can render negative.
   Evidence: null-excluded `mismatchedCols` + `added === 0 && removed === 0` → "Every column is 100% identical"; `Math.round(common - matchedCount)` unclamped.
   Pass A.

7. 11 of 15 widgets call `render()` with no try/catch — any throw leaves the widget stuck on "Loading…" forever.
   Evidence: bare `app.ontoolresult = ({ structuredContent }) => { render(structuredContent ?? {}); }` in query, query_diff, row_count_diff, value_diff_detail, schema_diff, list_checks, top_k_diff, histogram_diff, value_diff, get_model, get_server_info; only get_cll/impact_analysis/lineage_diff/profile_diff wrap it.
   Pass D.

8. `recce/widget_server.py:295-301` — `schema_diff` runs `get_lineage_diff()` + `adapter.select_nodes` synchronously inside the async handler (no executor), blocking the FastMCP event loop for all concurrent tool/resource calls; the `get_model`/`get_cll`/`histogram_diff` delegates share the pattern (inherited, but newly exposed to FastMCP concurrency).
   Pass H.

9. `recce/widget_server.py:531,617-628` — the documented `not_found: bool` miss-signal is dead code: `_tool_get_model` raises `ValueError` when the model is missing (`mcp_server.py:2179-2183`), so the branch never fires and the docstring contract is wrong.
   Pass C.

10. `recce/widget_server.py:360-370` — `get_server_info` docstring promises `base_status` "reveals stale or missing artifacts", but only `run_mcp_server` runs the freshness check; the widget server always reports `"unknown"`/`"single_env"`.
    Evidence: `_base_status` set only at `mcp_server.py:2815`; `run_widget_server` (`widget_server.py:2528-2573`) never sets it.
    Pass C.

11. Tests — `schema_diff` is the only widget tool whose handler path is never executed (the very tool that had the serialization regression); zero error-path tests across the 15 delegates (handler raising, `None` raw results); the `call_tool` widget-rejection guard (`mcp_server.py:1319-1321`) and the `run_widget_server`/`mcp_widget_server` bootstrap are untested (every test injects `ws._recce_server` directly).
    Evidence: no `ws.schema_diff(...)` call in `tests/test_widget_server.py`; only error tests are missing-HTML (line 143) and lineage over-cap (line 2290).
    Pass E.

12. `CLAUDE.md:101-137` — machine-local GBrain guidance ("GBrain is set up and synced on this machine", `.gbrain-source`, `~/.gstack/`) committed into the shared repo's agent instructions; false/noise for every other contributor's environment.
    Pass F.

## Notes

1. `recce/widget_server.py:60-64` — `SchemaChange` fields typed `List[Dict[str, str]]` but `_compute_schema_changes` emits `.get("type")` values that can be `None` (the documented present-but-None CLAUDE.md gotcha); use `Optional[str]`. Pass C.
2. `recce/widget_server.py:71-77` — `GitInfo.base_branch/base_sha/current_sha` can never populate (state `GitRepoInfo` has only `branch`); the matching render block in `get_server_info.html:296-309` is dead. Pass C.
3. `recce/widget_server.py:1975` — `get_cll` pops `_warning` but `_tool_get_cll` never adds one, and `lineage_diff` has no warning field at all — in single-env mode it renders an empty diff that reads as clean (same false-quiet class as the acknowledged schema_diff deferral, which covers only schema_diff). Pass A.
4. `recce/data/mcp/profile_diff.html:283,333` — string stats are escaped twice (`R&D` displays as `R&amp;D`); `strStatRow` null-checks miss `undefined`. Pass A.
5. `recce/data/mcp/lineage_diff.html:193,255,389` — layout keyed on `n.idx` with no validation; missing/duplicate `idx` stacks all nodes at one position. Pass A.
6. `recce/data/mcp/row_count_diff.html:296-311` — `=== null` guards miss `undefined` (safe under the current Pydantic dump contract, fragile to it), and the summary "Changed" counter counts null-vs-value rows the badge shows as "N/A". Pass A.
7. `recce/data/mcp/query_diff.html:485` / `value_diff_detail.html:444` — inline `onclick` handlers break silently under a strict host CSP; prefer `addEventListener`. Pass A.
8. `recce/data/mcp/top_k_diff.html:359,410` — only `base.values` is iterated (current-only categories depend on an unasserted server union contract) and counts aren't coerced to Number (string counts concatenate in the "shown" stat). Pass A.
9. `tests/test_check_base.py:77+` — seven dead `patch("recce.git.current_commit_hash")` contexts and the stale `test_status_stale_sha` name survive the opt-in change; `expected_base_sha` has no production caller (acknowledged deferral). Pass E/F.
10. PR description says "18-vs-20 tools"; the real split is 20 widgets-off / 5 widgets-on / 15 widget tools. Stale test name `..._six_tools_and_six_resources` (`tests/test_widget_server.py:89`) asserts 15. Pass F.
11. Verification: black/isort/flake8 clean at `6fb3bff`; all PR test files pass; the 19 failing tests in this env (test_cli_cache, test_server SPA routes, …) fail identically on `origin/main` — pre-existing/environmental, not introduced here.

---
*Multi-pass adversarial review (recce-dev:claude-code-review) — passes A–I + security skill; widget_server.py, 15 HTML widgets, and test suite each independently reviewed and cross-refuted.*

even-wei (changes requested):
Claude Code Review: Critical issues found (2 blockers, 12 issues). See review comment: https://github.com/DataRecce/recce/pull/1397#issuecomment-4687177256

PR #1403, opened 2026-05-27
fix(adapter): match DuckDB 1.5.x external-access denial wording

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix`

DuckDB 1.5.x reworded its external-access denial message:

| Version | Message |
|---------|---------|
| ≤ 1.4.x | `Cannot change enable_external_access setting while database is running` |
| ≥ 1.5.x | `Cannot enable external access while database is running` |

The pinned matcher in `recce/exceptions.py` only knew the old wording, so once CI resolved `duckdb==1.5.2` the raw error escaped unwrapped instead of becoming `DuckDBExternalAccessBlocked`, turning `main` red. Adds the new wording (keeps the old for DuckDB ≤ 1.4.x).

Refs DRC-3578

Verified against both duckdb 1.4.4 and 1.5.2 — the `SET bypass` case fails without this change and passes with it.

NONE

### DIFF
--- a/recce/exceptions.py
+++ b/recce/exceptions.py
@@ -37,7 +37,10 @@ class DuckDBExternalAccessBlocked(RecceException):
 _DUCKDB_EXTERNAL_ACCESS_DENIAL_MESSAGES = (
     "file system operations are disabled by configuration",
     "Loading external extensions is disabled through configuration",
+    # DuckDB <= 1.4.x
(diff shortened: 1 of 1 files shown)


kentwelcome (approved):
LGTM

PR #1405, opened 2026-05-28
feat(ui): export StalenessBanner from @datarecce/ui (DRC-3508)

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`feat` / refactor — promotes an existing OSS component to a public package export so the cloud shell can stop carrying a duplicate copy.

Promote `StalenessBanner` (and the companion `FirstTimePopover`) to public exports of `@datarecce/ui/components`, behind a prop-driven seam that supports both the OSS strip surface and the cloud floating-card surface from the same source file.

DRC-3506 mounted a local copy of `StalenessBanner` in the cloud-mode shell at `recce-cloud-infra/recce-cloud/src/common/components/recce-web/StalenessBanner.tsx` as a stopgap, because the OSS component was internal to `@datarecce/ui` and not part of the published surface. That copy has since diverged from OSS on several axes — visuals, dismissibility, toaster API, success-toast semantics, wording, the `cloudMode` gate. Bringing it back under one component requires more than a toaster swap, so the prop surface here mirrors each divergence point:

- `variant: 'banner' | 'card'` — OSS strip vs. cloud floating `Paper`
- `dismissible` + `sessionId` — opt-in close button with per-session sessionStorage signature
- `toastAdapter` — overrides the default OSS module-singleton; cloud passes its `useToast()` wrapped
- `successToastOnlyOnUserRefresh` — suppresses spurious "Base refreshed" toasts from unrelated cache transitions
- `requireCloudMode` — cloud shell sets `false` because it doesn't sit inside `LineageGraphContext` at the mount level
- `messageVariant: 'data' | 'metadata'` — wording diff
- `showFirstTimePopover` — cloud opts out (the popover is OSS-only per DRC-3508 acceptance)
- `cardSx` — additive `sx` for shell-specific positioning on the floating variant

The OSS callsite (`js/packages/ui/src/components/app/MainLayout.tsx`) stays `<StalenessBanner />` — every default matches current OSS behavior, so all 16 existing tests pass unchanged. 8 new tests cover the cross-shell prop surface (cloud-mode gate bypass, metadata wording, dismissibility + per-session key scoping, toast adapter routing, user-initiated success-toast gate, popover opt-out).

Resolves DRC-3508 (OSS half). A paired recce-cloud-infra PR will bump `@datarecce/ui`, swap the import in `RecceWebLayout.tsx` to `@datarecce/ui/components`, and delete the local copy + its test.

- **`FirstTimePopover` decision** (issue acceptance criterion): exported but cloud opts out via `showFirstTimePopover={false}`. The component is small, has no cloud-specific dep beyond a toaster, and the cloud-side intro decision is tracked separately per the issue's "out of scope" note. Exporting it makes the cross-shell option available without forcing it.
- **`successToastOnlyOnUserRefresh` default is `false`**: keeps OSS behavior unchanged. Cloud opts in. This is the only behavioral knob that the issue's "no behavioral difference for users" criterion is sensitive to — defaulting it off preserves the OSS contract.
- **`cardSx` instead of a `cardTopOffset` number prop**: cloud's `top: 180` overlay offset is shell-chrome-dependent (clears PreviewTopBar + PreviewNavBar + the in-canvas banner row). Taking arbitrary `sx` is more flexible than baking in a single offset, and the cloud-side TODO around moving to a ReactFlow `<Panel>` mount still applies — that future change won't need a prop surface revision.
- The dismissal signature logic (sessionStorage keyed on `sessionId::staleness-shape`) is lifted verbatim from the cloud-infra implementation, including the resync-on-key-change effect that handles Next.js layout preservation across `[sessionId]` URL changes.
- Build verified: `@datarecce/ui/dist/components.d.ts` exposes `StalenessBanner`, `FirstTimePopover`, `StalenessBannerProps`, `StalenessBannerVariant`, `StalenessMessageVariant`, `StalenessToastAdapter`. Type-check, lint, and the full 3,787-test suite all pass.

NONE

(The OSS banner continues to render identically. The cloud-infra companion PR will keep cloud users on the existing floating-card overlay — same visual, same wording, same dismissibility — but now from the shared source so future tweaks land in both shells at once.)

### DIFF
--- a/js/packages/ui/src/components/index.ts
+++ b/js/packages/ui/src/components/index.ts
@@ -52,8 +52,18 @@ export type {
   LineageCanvasProps,
   LineageViewProps,
   LineageViewRef,
+  StalenessBannerProps,
+  StalenessBannerVariant,
+  StalenessMessageVariant,
+  StalenessToastAdapter,
 } from "./lineage";
-export { LineageCanvas, LineagePageOss, LineageView } from "./lineage";
+export {
+  FirstTimePopover,
+  LineageCanvas,
+  LineagePageOss,
+  LineageView,
+  StalenessBanner,
+} from "./lineage";
 // Notification components
 export type { NotificationProps } from "./notifications";
 export { LineageViewNotification } from "./notifications";
--- a/js/packages/ui/src/components/lineage/FirstTimePopover.tsx
+++ b/js/packages/ui/src/components/lineage/FirstTimePopover.tsx
@@ -12,7 +12,7 @@ import { LOCAL_STORAGE_KEYS } from "../../api/storageKeys";
  *
  * @param anchorEl - DOM element to anchor the popover to (typically the banner).
  */
-interface FirstTimePopoverProps {
+export interface FirstTimePopoverProps {
   anchorEl: HTMLElement | null;
 }
 
--- a/js/packages/ui/src/components/lineage/StalenessBanner.tsx
+++ b/js/packages/ui/src/components/lineage/StalenessBanner.tsx
@@ -3,12 +3,20 @@
 import Box from "@mui/material/Box";
 import Button from "@mui/material/Button";
 import CircularProgress from "@mui/material/CircularProgress";
+import IconButton from "@mui/material/IconButton";
+import Paper from "@mui/material/Paper";
 import Stack from "@mui/material/Stack";
+import type { SxProps, Theme } from "@mui/material/styles";
 import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
+import { IoClose } from "react-icons/io5";
 import { MdInfo } from "react-icons/md";
-import { isSessionBaseOutdated, refreshSessionBase } from "../../api";
+import {
+  isSessionBaseOutdated,
+  refreshSessionBase,
+  type SessionStaleness,
+} from "../../api";
 import { useLineageGraphContext } from "../../contexts";
 import { useApiConfig, useServerInfo } from "../../hooks";
 import { toaster } from "../ui/Toaster";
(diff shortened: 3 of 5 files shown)


even-wei:
# Code Review: PR #1405

**SHA** `592a42cc9` · **Verdict** NO-GO

## Issues

1. `js/packages/ui/src/components/lineage/StalenessBanner.tsx:317-355` — The new `variant="card"` branch (floating `Paper` with `cardSx` merging) has zero direct test coverage. The 8 new tests at `__tests__/StalenessBanner.test.tsx:401-556` never pass `variant: "card"` (`grep -c variant` returns 0). The cloud companion PR is the only consumer and the only thing that will catch a regression in this path — `position: "fixed"`, the `zIndex: (theme) => theme.zIndex.drawer + 1` callback, and the `Array.isArray(cardSx) ? cardSx : [cardSx]` spread at `StalenessBanner.tsx:341` are all untested. Add a card-variant test that asserts the `Paper` renders, that a `cardSx` literal overrides `top`/`left`, and that an array-form `cardSx` flattens (the variant the cloud shell is most likely to pass).
   Pass E.

## Notes

1. `js/packages/ui/src/components/lineage/StalenessBanner.tsx:168-171` — `dismissible: true` without `sessionId` silently degrades: `dismissedKey` is `null`, sessionStorage persistence no-ops, the close button still works in-memory but the dismissal is lost on remount. No dev warning. Consider a `console.warn` (or a TS overload that makes `sessionId` required when `dismissible` is true) so a misconfiguration surfaces.
   Pass A.

2. `js/packages/ui/src/components/lineage/StalenessBanner.tsx:317-355` vs `:357-379` — the card and banner branches duplicate the `<Stack direction="row">` message row (icon + Typography + refreshButton + dismissButton + popover). About 12 lines per branch. A future layout change (second action button, alternate ordering, accessibility tweak) has to update both. Extract an inner `<MessageRow>` once a third caller appears.
   Pass F.

## Verification

- Tree clean; checked out worktree at `feature/drc-3508-promote-stalenessbanner-to-a-shared-datarecceui-export` (HEAD `592a42cc9`).
- `pnpm install --frozen-lockfile` — clean.
- `pnpm type:check` — clean.
- `pnpm lint` — clean.
- `pnpm test packages/ui/src/components/lineage/__tests__/StalenessBanner.test.tsx` — 24/24 pass (16 prior + 8 new).
- Full `pnpm test` — 3787 pass / 5 skipped / 164 files. No regressions.
- OSS callsite `js/packages/ui/src/components/app/MainLayout.tsx:84` unchanged (`<StalenessBanner />`); defaults match prior behavior.
- Security pass: no new auth/network/PII surfaces. `sessionStorage` writes use server-supplied signature, no injection vector. OWASP backstop clean.

---

Posted by `recce-dev:claude-code-review`.

even-wei (commented):
Review posted: https://github.com/DataRecce/recce/pull/1405#issuecomment-4561450174 — Verdict: NO-GO (1 issue, 2 notes). Card-variant has zero test coverage; blocking until that lands.

even-wei:
Addressing the review comment at https://github.com/DataRecce/recce/pull/1405#issuecomment-4561450174.

**Issue — card variant test coverage**: fixed in 9eb6e7d69. Adds 5 tests covering the previously-untested `variant='card'` branch:

- Renders a `MuiPaper-root` element (proves the branch executes vs. falling through to the default Box).
- End-to-end cloud-shell prop bundle (`variant='card'` + `dismissible` + `sessionId` + `toastAdapter` + `successToastOnlyOnUserRefresh` + `messageVariant='metadata'` + `showFirstTimePopover=false` + `cardSx={top:180}`) verifies metadata wording, refresh-base API call, dismissal hides the card, popover stays suppressed.
- `cardSx` object override produces a different emotion-generated className than the defaults — proves the merge reaches the Paper element.
- `cardSx` array form (`[sx1, sx2]`) flattens without throwing — guards against someone collapsing the `Array.isArray(cardSx) ? cardSx : [cardSx]` branch to `[cardSx]` always (which would nest the array and warn at runtime).
- `zIndex` theme callback resolves cleanly (spies on `console.error` so any "theme not in scope" MUI warning would fail the test).

Tests: 24 → 29 passing. Type-check, lint, and full `pnpm test` still all green.

**Notes**:
- **`dismissible: true` without `sessionId`**: agree the silent no-op is a footgun. Tracking as a follow-up — leaning toward a runtime dev warning (matches the existing `vi.spyOn(console, "error").mockImplementation(() => {})` pattern in OSS test conventions) rather than a TS discriminated union, since the prop interface is already wide and a union would propagate complexity. Will file separately.
- **~12 lines of Stack duplication between branches**: deferring per the "extract once a third caller appears" principle. Currently only two surfaces (OSS strip + cloud floating card).

even-wei:
**Dismissible guard — addressed in 707da5f69**.

Added a dev-time `useEffect` that fires `console.warn` when `dismissible=true` is passed without `sessionId`. Gated on `process.env.NODE_ENV !== "production"` so it's tree-shaken out of prod bundles. Three new tests cover: warning fires in the misuse case, stays silent when both props are provided, stays silent when `dismissible` is false.

Picked dev-warning over a TS discriminated union because the props interface is already wide (8 cross-shell knobs) and a union would force narrowing dance inside the component body. The warning catches the misuse at first dev render without changing the type-level contract.

Tests: 29 → 32 passing. Lint + type-check + full suite still green.

Final state for the three review items:
- ✅ Card variant test coverage — fixed in 9eb6e7d69
- ✅ Dismissible+sessionId footgun — fixed in 707da5f69
- 🟡 ~12 lines Stack duplication — deferred per the third-caller rule (still only OSS strip + cloud floating card today)

gcko:
# Code Review: PR #1405

**SHA** `b7d8ef7b` · **Verdict** GO

Promotes `StalenessBanner` and `FirstTimePopover` to public `@datarecce/ui/components` exports behind a prop-driven seam for the OSS strip vs. cloud floating-card surfaces. Quality gates green: Biome lint clean, `tsc --noEmit` clean, 3,795 tests pass, `tsdown` build emits the six new public type/value exports as the PR body claims.

OSS callsite (`MainLayout.tsx:84`) unchanged, every new prop defaults to OSS-current behavior, and the 16 pre-existing tests still pass alongside 17 new cross-shell tests (cloud-mode gate bypass, metadata wording, dismissible + per-session scoping, toast adapter routing, success-toast gate, popover opt-out, card variant + `cardSx` merge + zIndex theme callback). Race conditions in the timeout/refresh/dismiss interleavings are handled: the `if (!timeoutRef.current) return;` guard in the catch block prevents double error toasts, and `handleDismiss` clears the timeout + the user-initiated flag.

## Notes

1. `StalenessBanner.tsx:413-428` — `defaultToastAdapter` calls module-level `toaster.success/error`, which is an event-emitter. In the cloud bundle the OSS `<Toaster>` component is never mounted, so its `listeners` Set is empty and the default adapter silently no-ops. The paired cloud PR commits to passing `toastAdapter`, so this is mitigated in practice — but it's a sharp edge for any future caller who forgets. Consider a dev-time warning when `toastAdapter` is omitted *and* `requireCloudMode === false`, analogous to the existing `dismissible && !sessionId` guard.
   Pass A.

2. `StalenessBanner.tsx:183-185` — The dismissed-signature lazy `useState(...)` reads sessionStorage on initial render; the effect immediately reads it again and calls `setDismissedSignature(...)`. On a `sessionId` URL flip with a preserved layout (Next.js App Router), there's a one-frame window where `dismissedSignature` reflects the old session before the resync effect runs — the banner can flash visible for one paint before the effect hides it again. `useSyncExternalStore` (or re-deriving in render) would close the gap; not urgent because dismissible is cloud-only and shells generally remount on session change.
   Pass H.

3. `StalenessBanner.test.tsx:401-720` — Tests in the cross-shell `describe` block don't clear jsdom's `sessionStorage` between cases (only `localStorageMock.clear()`). Today this is harmless because every `sessionId` is unique (`sess-1`, `sess-A`, `sess-B`, `sess-cloud-1`, `sess-guarded`), but the suite is one ID collision away from order-dependent failures. Adding `sessionStorage.clear()` to the `beforeEach` would harden it.
   Pass E.

4. `StalenessBanner.tsx:136-139` — `useMemo<StalenessToastAdapter>(() => toastAdapter ?? defaultToastAdapter, [toastAdapter])` is effectively pass-through: identity equals `toastAdapter` when supplied, equals the module-stable `defaultToastAdapter` otherwise. The memo doesn't add stability the caller hasn't already delivered; an inline ternary would be equivalent and read more clearly. The success-toast effect's `toast` dep then carries whatever the caller passes — if cloud's `useToast()` returns a fresh wrapper object per render, the effect re-runs every render (its body is correctly idempotent under `transitioned`, so this is wasteful, not buggy).
   Pass A.

5. `lineage/index.ts:16` — `FirstTimePopover` is exported but `FirstTimePopoverProps` is not, while the `StalenessBanner` block exports its props type. External consumers needing to type a `<FirstTimePopover>` callsite have to reach for `React.ComponentProps<typeof FirstTimePopover>`. Minor public-surface inconsistency.
   Pass C.

6. `StalenessBanner.tsx:329-367` — The `card` variant uses `position: fixed` with `left: 50%; transform: translateX(-50%)` but no default `top`. Callers without a `cardSx={{ top: ... }}` get the card pinned to the viewport top under any sticky chrome. The PR body acknowledges this (cloud passes `top: 180`), and it's only reachable via `variant="card"`, so the OSS surface is safe — but worth a `// callers must position via cardSx` comment or a default `top: 16` to make the foot-gun less subtle.
   Pass F.

gcko (approved):
Approved with notes — see review comment for non-blocking observations.

even-wei:
@gcko Thanks for the thorough pass — addressed the actionable items in `9f2e21eb0`. Disposition per note:

- **Note 1 (`StalenessBanner.tsx:413-428`, `defaultToastAdapter` no-op outside OSS shell):** Fixed. Added a dev-time `useEffect` warning when `requireCloudMode={false}` is passed without a `toastAdapter`. Symmetric with the existing `dismissible && !sessionId` guard you referenced. 3 new tests cover it (warning fires on misuse, silent when adapter supplied, silent when `requireCloudMode` defaults to true — i.e. OSS shell). `StalenessBanner.tsx:206-222`.

- **Note 2 (`StalenessBanner.tsx:183-185`, one-frame flash on sessionId flip):** **Parked**. Fixing properly wants `useSyncExternalStore` (or re-deriving the dismissed signature in render rather than via state + effect), which is a larger shape change than the rest of this PR warrants. Agree with your "not urgent because dismissible is cloud-only and shells generally remount on session change" framing — the cloud shell remounts on `[sessionId]` navigation, so the flash window doesn't reach a user in practice today. Will file a follow-up if a preserved-layout cloud surface lands.

- **Note 3 (`StalenessBanner.test.tsx:401-720`, missing `sessionStorage.clear()`):** Fixed. Added `sessionStorage.clear()` to the cross-shell suite `beforeEach`. `StalenessBanner.test.tsx:412`.

- **Note 4 (`StalenessBanner.tsx:136-139`, pass-through `useMemo`):** Fixed. Replaced with an inline ternary; removed the unused `useMemo` import. `StalenessBanner.tsx:141-146`.

- **Note 5 (`lineage/index.ts:16`, `FirstTimePopoverProps` not exported):** Fixed. `FirstTimePopoverProps` is now exported from both `FirstTimePopover.tsx` and re-exported via `lineage/index.ts:16-19`. Aligns the public surface with the `StalenessBanner` block.

- **Note 6 (`StalenessBanner.tsx:329-367`, no default `top` foot-gun):** Fixed via doc comment rather than a default. Adding a default `top: 16` would conflict with cloud's explicit `top: 180`, and the OSS strip never reaches this branch. Documented the positioning contract on the `cardSx` prop instead so the requirement is visible at the call site. `StalenessBanner.tsx:103-113`.

Verification on the rebased state (`9f2e21eb0`):
- `pnpm type:check` clean.
- `pnpm lint` clean.
- `pnpm test packages/ui/src/components/lineage/__tests__/StalenessBanner.test.tsx` — 35/35 pass (32 prior + 3 new toastAdapter-guard tests).
- Full `pnpm test` — 3,798 pass / 5 skipped / 164 files.

PR #1407, opened 2026-05-28
fix(lineage): show impact coloring in the minimap (DRC-3250)

danyelf:
## Type
- [x] Bug fix

## Description

The lineage minimap colored nodes by **change status only**, so in the new CLL experience (`--new-cll-experience`) **impacted-but-unchanged** nodes — amber on the canvas — showed up neutral grey in the minimap. The minimap and canvas disagreed.

Root cause: in the OSS stack, a node's impact state isn't on `node.data` — `GraphNodeOss` reads it from the lineage context (`impactedNodeIds`). The minimap's `getNodeColor` had no access to that set, so it couldn't reflect impact.

### Fix

Single source of truth for a node's accent color:

- **`styles.tsx`** — new `getNodeChangeStyle({ changeStatus, isImpacted, newCllExperience })` that resolves the impacted-amber vs change-status color (muted "cll" palette in the new experience). This is the decision that previously lived inline in `LineageNode`.
- **`LineageNode.tsx`** — now renders from `getNodeChangeStyle` (no behavior change).
- **`config/nodeTypes.ts`** — `makeGetNodeColor({ impactedNodeIds, newCllExperience })` *copies* the node's color via the same helper; `getNodeColor` kept as the impact-agnostic default.
- **`LineageViewOss.tsx`** — memoizes a minimap color fn over `impactedNodeIds` + `newCllExperience`.

So the minimap no longer re-derives palette logic — it asks the same helper the node renders from, and the two can't drift.

## How tested

- New unit test `config/__tests__/nodeTypes.test.ts` asserts `makeGetNodeColor(...) === getNodeChangeStyle(...).color` across impacted / changed / unchanged + legacy-experience cases.
- `pnpm type:check` clean; `pnpm lint` clean; full frontend suite green (1924 passed).
- **Verified live** against jaffle_shop with `--new-cll-experience --whole-model-impact --impact-at-startup`: minimap renders **51 amber impacted** + 5 brown (modified) + 5 grey (unchanged), all matching `getNodeChangeStyle` and the canvas. Previously the 51 impacted were neutral.

## User-facing changes

In the new CLL experience, the lineage minimap now reflects impacted (amber) nodes, matching the main graph and the legend.

## Scope note

This fixes the **OSS production** view (`LineageViewOss`). The parallel `@datarecce/ui` `LineageCanvas` minimap is unchanged — it has no live consumer (neither the OSS app nor cloud render it), and `LineageView` doesn't wire impact data into nodes.

## Checklist
- [x] Tests pass
- [x] DCO sign-off

Closes DRC-3250

image

### DIFF
--- a/js/packages/ui/src/components/lineage/LineageCanvas.tsx
+++ b/js/packages/ui/src/components/lineage/LineageCanvas.tsx
@@ -24,7 +24,7 @@ import {
 import { LineageColumnNode } from "./columns";
 import { LineageEdge, type LineageEdgeData } from "./edges";
 import { LineageNode, type LineageNodeData } from "./nodes";
-import { cllChangeStatusColors } from "./styles";
+import { getNodeChangeStyle } from "./styles";
 
 export interface LineageCanvasProps {
   /** Nodes to display */
@@ -155,8 +155,17 @@ export function LineageCanvas({
         {showMiniMap && (
           <MiniMap
             nodeColor={(node) => {
+              // Copy the node card's accent color from the shared source of
+              // truth so the minimap can't drift from the canvas. Impact lives
+              // on `node.data` in this stack (the OSS stack reads it from
+              // context instead), so impacted-but-unchanged nodes are amber
+              // here too (DRC-3250). Mirrors LineageNode's resolution.
               const data = node.data as LineageNodeData;
-              return cllChangeStatusColors[data.changeStatus ?? "unchanged"];
+              return getNodeChangeStyle({
+                changeStatus: data.changeStatus,
+                isImpacted: data.isImpacted,
+                newCllExperience: data.newCllExperience,
+              }).color;
             }}
           />
         )}
--- a/js/packages/ui/src/components/lineage/LineageViewOss.tsx
+++ b/js/packages/ui/src/components/lineage/LineageViewOss.tsx
@@ -97,9 +97,9 @@ import {
   EXPLORE_MIN_ZOOM,
   edgeTypes,
   FIT_VIEW_PADDING,
-  getNodeColor,
   initialNodes,
   LEGIBLE_MIN_ZOOM,
+  makeGetNodeColor,
   nodeTypes,
 } from "./config";
 import {
@@ -508,6 +508,14 @@ export function PrivateLineageView(
     wholeModelChangedNodeIds,
     publish: publishImpactSets,
   } = usePublishedImpactSets();
+
+  // MiniMap node coloring. Mirror the canvas: in the new CLL experience,
+  // impacted-but-unchanged nodes are amber, so thread the impacted set through
+  // instead of coloring by change status alone (DRC-3250).
+  const minimapNodeColor = useMemo(
+    () => makeGetNodeColor({ impactedNodeIds, newCllExperience }),
+    [impactedNodeIds, newCllExperience],
+  );
 
   // biome-ignore lint/correctness/useExhaustiveDependencies: Intentionally only run when lineageGraph changes (initial load/refetch).
   useLayoutEffect(() => {
@@ -1546,7 +1554,7 @@ export function PrivateLineageView(
             </Panel>
             {nodes.length <= MINIMAP_NODE_THRESHOLD && (
               <MiniMap
-                nodeColor={getNodeColor}
+                nodeColor={minimapNodeColor}
                 nodeStrokeWidth={3}
                 zoomable
                 pannable
--- a/js/packages/ui/src/components/lineage/config/__tests__/nodeTypes.test.ts
+++ b/js/packages/ui/src/components/lineage/config/__tests__/nodeTypes.test.ts
@@ -8,12 +8,15 @@
  * `getNodeChangeStyle(...).color` for the equivalent state (the single source
  * of truth `LineageNode` also renders from). The one thing the minimap can't
  * read off `node.data` is impact, so we also check the impacted set is honored.
+ * The unchanged-node case is additionally pinned against a literal palette
+ * value (not just `getNodeChangeStyle`) so a future drift is caught.
  */
 
 import { describe, expect, it } from "vitest";
 import type { LineageGraphNode } from "../../../../contexts/lineage/types";
+import { colors } from "../../../../theme";
 import { getNodeChangeStyle, getStyleForImpacted } from "../../styles";
-import { getNodeColor, makeGetNodeColor } from "../nodeTypes";
+import { makeGetNodeColor } from "../nodeTypes";
 
 // Minimal node factory — only `id` and `data.changeStatus` matter here.
 function makeNode(id: string, changeStatus?: string): LineageGraphNode {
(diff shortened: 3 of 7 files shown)


gcko:
# Code Review: PR #1407

**SHA** `ed8dc9b5` · **Verdict** NO-GO

Reviewed as a fan-out of passes A (correctness), B (security), C (cross-reference), D (error-handling), E (tests), F (diff-specific), H (React hooks), each with an independent adversarial verification pass. `pnpm type:check` (workspace `tsc --noEmit`) is green for the changed files.

The core refactor — a single `getNodeChangeStyle` source of truth that both `LineageNode` and the MiniMap render from — is sound and well-documented. Two issues block:

## Issues

1. `config/nodeTypes.ts:57-76` (live caller `LineageViewOss.tsx:516`) — Unchanged minimap nodes change color in **both** experiences as an out-of-scope side effect, and the docstring claims the opposite.
   Evidence: old `getNodeColor` returned `colors.neutral[400]` (`#A3A3A3`) for a node with no `changeStatus` (merge-base `nodeTypes.ts:38-40`). Unchanged nodes carry `changeStatus: undefined` (`coerceCllChangeStatus` returns undefined), so they now route `getNodeChangeStyle → getIconForChangeStatus(undefined) → status="unchanged" → changeStatusColors.unchanged = colors.neutral[500]` (`#737373`) (`styles.tsx:507,681`; `colors.ts:63-64`). The `nodeTypes.ts:70` comment "equivalent to the original change-status-only coloring" is therefore false. Cosmetic (two neutral grays), and possibly intended — but if so, pin it in a test and fix the comment; if not, preserve `neutral[400]`.
   Passes A, C, F.

2. `LineageCanvas.tsx:157-160` — The migrated CLL canvas (a public `@datarecce/ui` export used by `LineageView` and the `CllExperience` Storybook) still has the exact DRC-3250 minimap bug this PR fixes for `LineageViewOss`: its `nodeColor` reads `cllChangeStatusColors[data.changeStatus ?? "unchanged"]` with no impacted-set lookup, so impacted-but-unchanged nodes are not amber in its minimap.
   Evidence: `nodeColor={(node) => cllChangeStatusColors[(node.data as LineageNodeData).changeStatus ?? "unchanged"]}` — no `getNodeChangeStyle`, no `impactedNodeIds`. `CllExperience.stories.tsx` feeds it `newCllExperience:true` + `isImpacted` and shows the minimap, so the card-vs-minimap mismatch is directly reproducible. Either fix it here with the new `makeGetNodeColor`/`getNodeChangeStyle` or file a follow-up so DRC-3250 isn't half-shipped.
   Pass F.

## Notes

- `config/__tests__/nodeTypes.test.ts` — Most assertions compare `makeGetNodeColor(...)` to `getNodeChangeStyle(<same args>).color`, but the factory *is* a thin wrapper around `getNodeChangeStyle`, so those cases verify argument plumbing, not palette values, and can't catch the `neutral[400]→[500]` drift above. Only the impacted-amber case (`:34-38`) has an independent oracle (`getStyleForImpacted`). Palette values are pinned in `styles.test.ts`, so this is scoping, not a gap — but a literal pre-PR constant for the unchanged case would have caught issue #1. Passes A, E, F.
- `config/nodeTypes.ts:76` — `getNodeColor` is now dead production code: the only live minimap caller uses `minimapNodeColor` (`LineageViewOss.tsx:1557`); `getNodeColor` is referenced only by its test and the `config/index.ts` re-export. Its "Used by MiniMap" / "kept for callers that don't have the CLL impacted set" docstring is stale. Passes C, F.
- `config/nodeTypes.ts:63` — `isImpacted: newCllExperience ? impactedNodeIds?.has(node.id) : false` duplicates a guard `getNodeChangeStyle` already applies (`styles.tsx:548-549` gates on `newCllExperience && isImpacted && isUnchanged`), so the inline ternary can't change the result. Pass F.
- `config/nodeTypes.ts:59` — The callback is typed `(node: LineageGraphNode)` but the MiniMap is parameterized over `LineageGraphNodes` and column nodes are pushed as top-level entries. No crash (column nodes carry `data.changeStatus` and `id`), matches pre-PR typing — clarity only. Pass D.
- `config/__tests__/nodeTypes.test.ts:19-23` — The mock factory types `changeStatus` as `string` (real type is the `"added"|"removed"|"modified"` union) and casts `as unknown as LineageGraphNode`, omitting required fields. Test-local looseness; the production consumer is fully typed. Pass E.
- `LineageViewOss.tsx:515` — The `minimapNodeColor` `useMemo` wiring has no test; deps `[impactedNodeIds, newCllExperience]` match the closure inputs exactly, so the correctness-bearing logic lives in the tested factory. Pass E.

gcko (changes requested):
NO-GO — 2 issues (minimap color regression + LineageCanvas left unfixed). Full review in the issue comment above.

danyelf:
Thanks for the review — both blockers addressed in `f90b821a` (rebased on latest `main`).

**Blocker 2 — `LineageCanvas` minimap.** The CLL canvas minimap now resolves its node color through the same `getNodeChangeStyle` source of truth the node card uses, reading impact off `node.data` (this stack carries impact on the node, unlike the OSS stack which pulls it from context). So impacted-but-unchanged nodes are now amber in its minimap too — DRC-3250 is fully shipped across both stacks rather than half-fixed. The `CllExperience` story that reproduced the mismatch now agrees.

**Blocker 1 — unchanged-node color shift.** The `neutral[400] → neutral[500]` change is intentional and we're keeping it: routing through the shared helper is the whole point of the refactor, and a single resolved palette is preferable to preserving a one-off legacy gray. To stop it from being silent, it's now pinned with a literal-constant test (not a self-referential comparison to `getNodeChangeStyle`), with a comment recording the deliberate shift. The misleading "equivalent to the original" docstring is gone.

**Dead-code note.** Removed `getNodeColor` entirely — it had no production consumer (the live minimap uses `makeGetNodeColor`/`minimapNodeColor`), along with its re-export and test.

Verified: `type:check`, `lint`, and the lineage + nodeTypes suites all green.

gcko:
# Code Review: PR #1407

**SHA** `ed8dc9b5` · **Verdict** GO

A tight, single-source-of-truth refactor. The core claim holds: the minimap now derives its accent color from the same `getNodeChangeStyle` the node card renders from, and the impact lookup matches the card's exactly.

## Blockers
None.

## Issues
None.

## Notes
1. `config/nodeTypes.ts:76` — the default `getNodeColor` is **not** strictly "equivalent to the original change-status-only coloring" as the PR description states. Old code returned `colors.neutral[400]` for a falsy `changeStatus`; the new path routes `undefined → "unchanged" → changeStatusColors.unchanged === colors.neutral[500]` (`styles.tsx:681`). The effect is sub-perceptible (neutral[400] vs neutral[500] in a minimap) and actually *aligns* the minimap with the canvas card — which has always used `neutral[500]` for no-status nodes — so it's in-spirit with the PR's goal. Flagging only because the description claims exact equivalence and the new test (`getNodeColor(makeNode("m")) === getNodeChangeStyle({}).color`) encodes the new value rather than guarding the old one.

2. `config/nodeTypes.ts:76` — the default `getNodeColor` now has **no live render consumer**: `LineageViewOss` switched to `makeGetNodeColor`, and the only other reference is the `config/index.ts` re-export plus the test. It's retained as public API "for callers that don't have the CLL impacted set," but none exist today. Fine to keep (tested, reasonable surface), just noting it's currently dead for rendering.

## Verification
- `pnpm exec vitest run …/config/__tests__/nodeTypes.test.ts` → 8/8 passed.
- `pnpm type:check` (`tsc --noEmit`) → clean.
- `pnpm exec biome check` on all 6 changed files → clean.
- Cross-ref: minimap impact derivation (`newCllExperience ? impactedNodeIds?.has(node.id) : false`) matches `GraphNodeOss.tsx:304` (`newCllExperience ? impactedNodeIds.has(id) : false`) exactly. `getNodeChangeStyle` replicates the prior inline logic in `LineageNode` verbatim; `color === hexColor` always (`styles.tsx:513-514`), so the field swap in `getNodeColor` is a no-op.

## Limits
Pass B (security): the diff is pure presentational color mapping — no input source, sink, auth boundary, URL/DOM injection, or data flow from a request. No security surface to trace; full security-review skill not invoked as it would have nothing to analyze.

gcko (approved):
Claude Code Review: GO. No blocking issues — two non-blocking notes in the review comment.

PR #1411, opened 2026-05-29
feat: UI wiring for inline paired distributions (DRC-3390 Stage C)

danyelf:
## Stage C — UI wiring for inline paired distributions (DRC-3390)

Connects Stage A's paired-histogram cells (#1398) to Stage B's `profile_distribution` data (#1399) behind the `--inline-profile` flag. Under the new-CLL experience, the schema **diff** view renders a per-column **Distribution** column with paired histograms.

### What's in it

- **`useInlineProfileDistribution`** — submits one `profile_distribution` run per node **synchronously** (`POST /api/runs` without `nowait`, so the completed run comes back in a single response — no polling loop), narrows the `ProfileDistributionResult` envelope, and emits one Amplitude timing event per resolved run. Transport failures (network, HTTP 5xx, and client timeouts) are reported to Sentry so slow-run timeouts surface. The run is keyed by `nodeId` and uses `placeholderData` so the prior histograms stay on screen while a wider query loads (see scoping below).
- **`InlineProfileDistributionCell`** — per-column state machine: loading (pending dot) / run error / blank / per-column failure / histogram (continuous) / top-K ranks (discrete). Owns the snake→camel payload mapping and the datetime epoch-seconds formatting seam. Never renders synthetic data. (The discrete **counts** path + `baseTotal`/`currentTotal` are wired to the full `run.ts` payload contract now, even though Stage B/DuckDB only emits **ranks** — Stage D's full adapters emit counts, so that stays a backend-only change.)
- **Changed-column scoping** — under new-CLL, the run is scoped to the columns that actually changed: the union of the two signals the grid already colors from — `columnChanges` (this node's own added/removed/modified columns) and `impactedColumns` (downstream-impacted). A **whole-model change** (read from the lineage's `wholeModelChangedNodeIds` set) profiles every column. A **"Profile all columns"** button widens the single run to all columns on demand.
- **Grid** — `toSchemaDataGrid` appends a "Distribution" column (colId `distribution`) only when distribution data is threaded in.
- **`PrivateSchemaView`** — fires the hook, threads data into the grid, renders `ProfileDistributionUnsupportedBanner` on the `{status:"unsupported"}` envelope.
- **Telemetry / CSS** — `trackProfileDistribution` event + `schema_view` explore-source; `.schema-column-profile-distribution`, `.schema-distribution-pending`, `.schema-distribution-error`.

### Notes for reviewers

- **User-facing:** `--inline-profile` now turns on `--new-cll-experience` automatically (the feature only renders there); previously the flag alone did nothing.
- The feature is gated behind **new-CLL** (`new_cll_experience`); the hook additionally self-gates on the `inline_profile` flag. With both off it's a no-op.
- The synchronous-submit choice fits the DuckDB `approx_all` path that ships here (fast). If slow remote-warehouse adapters land later and start hitting the client timeout, those surface in Sentry — follow-up **DRC-3629** tracks giving the OSS API client a request timeout so that signal fires in OSS too.
- "Profile all columns" re-requests every column; the backend cache is keyed on the exact column set, so it recomputes the already-shown columns — cheap on DuckDB, and **DRC-3630** tracks per-column backend caching if it ever matters.

### Verification

- `pnpm type:check`, `pnpm lint`, `pnpm test`, `pnpm run build` — all green.
- Backend `tests/tasks/test_profile_distribution.py` passing (the ranks-payload assertions live in #1399).
- **Live browser verify (DuckDB + `--inline-profile --whole-model-impact`):** Distribution column renders; scoped to the impacted column by default with the "Profile all columns" button; widening fills every column with no pending-dot flash; backend trace shows one `POST /api/runs` per run and zero `/wait` polls; no console errors.

## Checklist
- [x] Tests added/updated
- [x] DCO sign-off

image
image
image


even-wei:
# Code Review: PR #1411

**SHA** `4f742fd7` · **Verdict** NO-GO

Verified: `pnpm type:check` exit 0; `pnpm test` 173 files, 3911 passed / 5 skipped / 0 failed; `pnpm install --frozen-lockfile` exit 0. Stage B contract (`run.ts`) consumed correctly — variable-length edges handled (`computeContinuousLayout` validates `edges.length === density.length + 1` per env, never index-aligns or hard-indexes); datetime epoch-seconds formatting gated on out-of-band dbt type; all-null `current_ranks` + `trimmed` render gap-on-absent without crashing; `kind:null` / `status:"unsupported"` narrowed exhaustively; `inline_profile` flag short-circuits the hook to `disabled`; no `findByRunType("profile_distribution")` calls; no `dangerouslySetInnerHTML` / XSS surface (Pass B clean).

## Blockers
None.

## Issues
1. `js/packages/ui/src/hooks/useInlineProfileDistribution.ts:122,169-182` — On a cache-hit remount (an explicitly designed scenario — see the hook's own docstring lines 34-37: "memoizes across re-renders and remounts e.g. toggling between schema tabs"), the timing event re-fires with a garbage wall time and double-counts the run.
   Evidence: submit `queryFn` (line 128) sets `submitStartRef` but does NOT re-run on remount (`staleTime: Infinity`, cached `runId` returned). `submitStartRef`=`useRef(0)` and `trackedRunRef`=`useRef(null)` both re-init on the fresh mount, so the terminal-state effect emits `total_wall_ms = round(performance.now() - 0)` ≈ page lifetime, and the per-mount guard does not prevent re-emit for an already-tracked run.
   Pass H / A.
2. `js/packages/ui/src/components/schema/SchemaView.tsx:279-387` — The most complex new logic in the PR (changedColumns scoping, primary+complement dual-run merge, `restFailures` complement-only-failure path, "Profile all columns" expansion) has zero tests.
   Evidence: `components/schema/__tests__/` contains only `ColumnNameCell.test.tsx`; no test exercises `distributionData` merge, the `distributionRest.status === "error"` → per-column `{kind:null}` fan-out, or the opt-in expansion.
   Pass E.
3. `js/packages/ui/src/hooks/__tests__/useInlineProfileDistribution.test.tsx` — Hook `status:"error"` path and `emitTiming`'s error branch are untested.
   Evidence: tests cover ok / unsupported / disabled / polling only; no `mockRejected` or `run.error` case, so `track.ts:245-252` (error timing emit) and the hook's `error` envelope (`useInlineProfileDistribution.ts:195-205`) are uncovered.
   Pass E.

## Notes
1. `js/packages/ui/src/lib/api/track.ts:189-191` — `trackProfileDistribution` is Amplitude-only; every other wrapper here dual-emits via `capturePosthog`. Given the active Amplitude→PostHog migration this event will go dark once Amplitude is removed. Intentional per the PR description, so non-blocking.
   Pass C.
2. `js/packages/ui/src/components/data/PairedHistogramContinuous.tsx:60-67` (pre-existing, outside this diff) — `PairedHistogramContinuousData` doc comments still claim "Length 12"/"Length 11", contradicting the variable-length edge contract the code itself correctly handles. Worth a follow-up cleanup so the next consumer doesn't trust the stale comment and hard-index.
   Pass C.

even-wei (changes requested):
Claude Code Review: NO-GO. 3 issue(s). See review comment.

gcko (approved):
## Review: APPROVE — no blocking issues

No blocking issues remain. The pass-1 **HIGH** (untested SchemaView orchestration around a two-run primary+complement model) and the pass-1 **MEDIUM** (a primary-run error masking successfully-profiled complement columns) were both **resolved by the post-review refactor** — commit `22fd6e9f` ("synchronous submit + single-run widening", followed by `aeea4e25` / `97cc6d6a`) collapsed the two-run model into a single widening run. The high-risk complement-computation + three-way merge logic those findings were anchored on no longer exists in the code. What remains is one non-blocking **MEDIUM** plus a few **NIT**-level follow-ups, all listed below.

### MEDIUM (non-blocking) — moderate-complexity SchemaView wiring is untested

The SchemaView orchestration still has no integration test. The high-risk logic that justified a HIGH was removed before this pass: current `SchemaView.tsx` makes a **single** `useInlineProfileDistribution` call (line 317), "Profile all columns" merely widens that same run's `scopedColumns` to `undefined` (line 303), and `distributionData` is now a flat status→fields mapping (lines 339–353) — there is no `restColumns`, no complement run, no `{...primary, ...rest, ...restFailures}` merge.

What is left untested at the SchemaView level is moderate-complexity wiring: the `changedColumns` set derivation (union of `Object.keys(columnChanges)` + nodeId-prefix-stripped `impactedColumns`, lines 280–290), the `hasChangedScope` / `scopedColumns` / `profileEnabled` / `profilingAll` gates, and the button gate. The leaf cell, the hook (incl. the error branch), and the grid generator are unit-tested; the schema `__tests__/` dir still holds only `ColumnNameCell.test.tsx`, and the derivation is inline (no `selectChangedColumns` helper exists). A bug in the inline derivation degrades gracefully — wrong/extra columns profiled, with the "Profile all columns" escape hatch — rather than corrupting data or crashing, so this is a real but non-blocking follow-up. A SchemaView-level test is still recommended ("wiring is where small bugs live").

### NIT (non-blocking follow-ups)

- **Hook error-path edge coverage.** The consequential transport-failure branch is now covered (`useInlineProfileDistribution.test.tsx:232–253` asserts `status === "error"`, the surfaced error, and a single `status:"error"` timing event). Residual gap: the rare resolved-run-with-`run.error`-field path (`useInlineProfileDistribution.ts:175`, via `runError`) is still uncovered — minor, as the synchronous path usually throws instead.

- **"Profile all columns" on an unchanged model is mild UX, not a defect.** Clicking flips `profileEnabled` (`SchemaView.tsx:307-309`) and fires an all-columns run, so the button works as the intended entry point. The only nit is presenting the action before any change exists — a deliberate product choice.

- **React Query cache key can collide for column names containing a comma.** `useInlineProfileDistribution.ts:113` uses `keySuffix = columns?.join(",") ?? "*"`; a subset `["a,b"]` and `["a","b"]` map to the same key. dbt identifiers don't contain commas in practice (theoretical), but a separator that can't appear in a column name (or hashing the sorted set) removes the footgun.

- **`isDatetimeType` substring match is loose, though safe on type strings.** `InlineProfileDistributionCell.tsx:340-349` flags any type whose lowercased string `includes("date"|"time"|...)`. It runs on the dbt *type* (`row.currentType ?? row.baseType`), not the column name, so standard SQL types are fine; a future user-defined/aliased type containing one of those substrings would get epoch-seconds date formatting applied to plain numeric edges.

- **"Profile all columns" yields two timing events across the journey.** A single hook instance fires two sequential queries (initial changed-columns run, then the widened all-columns run on click), each emitting one per-run Amplitude event. Consistent with the per-task telemetry contract, but downstream funnel/duration analysis should expect two events for the expand journey.

### Positive

Never renders synthetic data (explicit `if (!payload) return null`); good accessibility on the new states (`role="status"`/`aria-label` on the pending dot, `role="img"`/`aria-label`/`title` on the error icon, `aria-hidden` on the inner SVG); unmount halts polling cleanly; the frontend conforms to the frozen `run.ts` payload contract (both ranks and counts modes handled).

danyelf:
@gcko Thanks for the approve + the detailed follow-up list. Addressed the actionable items in `f89fdb55`:

**MEDIUM — SchemaView scoping wiring untested.** Extracted the inline-profile scoping decision into a pure `selectInlineProfileScope` helper (`components/schema/selectInlineProfileScope.ts`) — the `selectChangedColumns`-shaped seam you noted was missing — and unit-tested it (`__tests__/selectInlineProfileScope.test.ts`): the changed-column union, nodeId-prefix-stripped `impactedColumns` + dedup, the `hasChangedScope`/`scopedColumns`/`profileEnabled`/`profilingAll` gates, and the "Profile all columns" expand→hide gate. `SchemaView` now calls the helper from one `useMemo`. Behavior-preserving; full suite green (3927 passed).

**NIT — hook `run.error`-field path.** Added a test for the resolved-run-with-task-`error` path (`useInlineProfileDistribution.ts:175` via `runError`): asserts `status === "error"`, no Sentry capture (no throw), and a single error timing event.

**NIT — cache key comma collision.** `keySuffix` now uses `JSON.stringify(columns)` instead of `join(",")`, so `["a,b"]` and `["a","b"]` map to distinct keys.

Intentionally left as-is, per your read:

- **`isDatetimeType` substring match** — runs on the out-of-band dbt *type* string (`currentType ?? baseType`), where every real SQL type containing `date`/`time`/`timestamp`/`datetime` *is* a temporal type; tightening risks dropping legitimate types (`timestamp without time zone`, `timestamptz`) for a hypothetical aliased type. Keeping the DRC-3504 behavior.
- **"Profile all columns" on an unchanged model** — deliberate entry point.
- **Two timing events on the expand journey** — consistent with the per-run telemetry contract; noted for downstream funnel analysis.

even-wei:
# Code Review: PR #1411

**SHA** `03e75d1` · **Verdict** NO-GO

## Blockers
None.

## Issues
1. `js/packages/ui/src/components/schema/selectInlineProfileScope.ts:79-81` — Prefix-stripping the **global** impacted-column set mis-attributes sibling models' columns whenever one node id is a prefix of another (`model.shop.orders` absorbs `model.shop.orders_summary_total` → bogus column `summary_total`), firing spurious profile runs and rendering an empty Distribution column on unchanged models.
   Evidence: `computeImpactedColumns` returns ids for *every* node in the CLL graph (`computeImpactedColumns.ts:43-48`); `id.startsWith(`​`${nodeId}_`​`)` matches sibling ids; the backend silently intersects unknown columns away (`recce/tasks/profile_distribution.py:688`), so the run returns `{columns: {}}` and an empty Distribution column appears. Underscore-separated dbt model names (`orders` / `orders_summary`, `stg_x` / `stg_x_y`) make this collision routine.
   Pass A.

2. `js/packages/ui/src/components/schema/SchemaView.tsx:313-315` — Resetting `profileAllColumns` in an effect leaves one committed render of the **new** node with the stale opt-in, firing a spurious all-columns `profile_distribution` run before the scoped one.
   Evidence: `NodeView` is not keyed by node id (`LineageViewOss.tsx:1592`), so `SchemaView` stays mounted across node switches; on the first render after a switch, `profileEnabled` is true via stale `profileAllColumns` and the hook's query (`["profile_distribution", newNodeId, "*"]`, hook called at `SchemaView.tsx:304`) initiates its fetch in effects declared *before* the reset effect; the `queryFn` (`useInlineProfileDistribution.ts:132`) ignores React Query's abort signal, so the backend run completes regardless. Cheap on DuckDB today; an expensive stray warehouse query once Stage D adapters land. Key the state by node id (or store `{nodeId, optIn}` and derive) instead of reset-in-effect.
   Pass H.

3. `js/packages/ui/src/components/schema/selectInlineProfileScope.ts:89-90` — A whole-model change does **not** widen the scope to all columns when any changed/impacted column exists, contradicting the function's own docstring and the PR description.
   Evidence: `scopedColumns = profileAllColumns || !hasChangedScope ? undefined : changedColumns` — `wholeModelChange` feeds only `profileEnabled`; the docstring (line 58) says "A whole-model change profiles all columns" and the PR body repeats it, while `SchemaView.tsx`'s comment says the opposite ("with a whole-model change **and no scoped columns**"). The `wholeModelChange && columnChanges` combination — a breaking SQL change that also adds a column, the common case — is untested (`selectInlineProfileScope.test.ts` covers whole-model only with `columnChanges: null`). Either the scoping is wrong, or two docs are wrong and the deciding test is missing.
   Pass A.

4. `js/packages/ui/src/lib/api/track.ts:190` — `trackProfileDistribution` emits only the Amplitude `track()` call, omitting the `capturePosthog()` dual-emit every other wrapper in this file performs.
   Evidence: `track.ts:35` documents `capturePosthog` as "used by ALL wrappers alongside the (kept) Amplitude track() call"; every sibling wrapper (e.g. `trackColumnLevelLineage`, lines 163-166) dual-emits. With the Amplitude→PostHog consolidation in flight, this event silently never reaches PostHog and dies with Amplitude's removal.
   Pass C.

## Notes
1. `js/packages/ui/src/components/data/InlineProfileDistributionCell.tsx:66-75` — `isDatetimeType` matches bare `TIME` columns, whose epoch values are seconds-since-midnight; `formatEpochSeconds` renders every histogram tooltip edge as "Jan 1, 1970". Backend classifies `time` as datetime too (`profile_distribution.py:215-220`), so bins are correct — only the tooltips are meaningless. Pass A.
2. `js/packages/ui/src/hooks/__tests__/useInlineProfileDistribution.test.tsx:167` — The placeholder-drop on node **switch** ("navigating to a different node must NOT show the previous node's distributions", the `placeholderData` comment's stated contract) is asserted nowhere; only same-node widening is tested. Pass E.

---
*Quality gate on `03e75d1`: `pnpm lint` ✓ · `pnpm type:check` ✓ · `pnpm test` 3927 passed ✓ · `pnpm run build` ✓*

even-wei (changes requested):
Claude Code Review: NO-GO — 4 issues found (scope mis-attribution on prefix-named models, stale opt-in spurious run, whole-model scoping doc/code contradiction, missing PostHog dual-emit). See review comment for details.

danyelf:
Thanks for the thorough review. All four issues and both notes are addressed in `14272756`.

## Issues

**1. Sibling-column mis-attribution (`selectInlineProfileScope.ts`)**
Replaced prefix-stripping with exact `${nodeId}_${column}` membership over the node's *own* column names. `SchemaView` now computes `nodeColumnNames` (base ∪ current columns) and threads it in; impacted ids are attributed only when `nodeColumnNames` contains the stripped column, so `model.shop.orders` can no longer absorb `model.shop.orders_summary`'s columns. When the node's column list is unavailable we fall back to own-changes-only rather than risk a bogus column. Added a sibling-prefix regression test and an unknown-column-names fallback test.

**2. Stale opt-in fires a spurious all-columns run (`SchemaView.tsx`)**
The "Profile all columns" opt-in is now keyed by node id and *derived* (`profileAllColumnsNodeId === nodeId`) rather than stored as a bare boolean reset in an effect. Switching nodes clears it within the same render, so there's no longer a stale frame that kicks off an all-columns `profile_distribution` run before the reset lands. Removed the reset effect entirely.

**3. Whole-model change didn't widen to all columns (`selectInlineProfileScope.ts`)**
Resolved in favor of the docstring/PR intent: a whole-model change now widens `scopedColumns` to every column even when specific columns also changed, because the change isn't pinned to specific columns (any column's values may have shifted). Fixed the contradicting `SchemaView` comment and added the previously-missing `wholeModelChange && columnChanges` test.

**4. Missing PostHog dual-emit (`track.ts`)**
`trackProfileDistribution` now calls `capturePosthog("profile_distribution", props)` alongside the Amplitude `track()`, matching every other wrapper. (We considered a single `emit()` helper to prevent this class of bug, but left it out of scope since the Amplitude→PostHog consolidation will collapse every wrapper to a single call anyway.)

## Notes

**1. `TIME` tooltips (`InlineProfileDistributionCell.tsx`)**
Split the type predicate: `isDatetimeType` now matches only `date`/`timestamp`/`datetime` (epoch seconds → calendar date), and a new `isTimeOfDayType` matches bare `TIME`/`TIME WITH/WITHOUT TIME ZONE` (seconds-since-midnight → `HH:MM:SS`) while excluding `timestamp`/`datetime`. Tooltips on `TIME` columns now read as clock times instead of "Jan 1, 1970". Covered by unit tests (`TIME` → `HH:MM:SS`, no 1970; `timestamp without time zone` → still date-formatted) and a new storybook story (`InlineProfileDistributionCellDatetime.stories.tsx`) contrasting the two. Exported `InlineProfileDistributionCell` from `@datarecce/ui` to support the story.

**2. Missing node-switch placeholder test (`useInlineProfileDistribution.test.tsx`)**
Added the regression test for the documented contract: navigating to a different node drops the previous node's distributions (the placeholder is reused only when `previousQuery.queryKey[1] === nodeId`).

## Verification
`pnpm lint` ✓ · `pnpm type:check` ✓ · `pnpm test` (full suite) ✓ · `pnpm run build` ✓ · `storybook build` ✓.

One unrelated thing surfaced: the storybook *vitest* harness (`packages/storybook/vitest.config.ts`) is missing the `@datarecce/ui/theme` / `/advanced` / `/api` aliases that `.storybook/main.ts` has, so story play-functions don't run under that harness (affects all stories, pre-existing). Filing separately rather than bundling it here.

danyelf:
Self-review pass pushed (`bc069f84`, `0682d7e7`):

**Fixed**
- `--inline-profile` alone was a silent no-op — now implies `--new-cll-experience`.
- Datetime histogram tooltips no longer show the wrong calendar day off-UTC.
- Corrupt (non-finite) histogram payloads now blank the cell instead of drawing a meaningless chart.
- "Profile all columns" no longer shows alongside the unsupported banner.
- Added coverage for the per-row distribution cell (pending vs. blank vs. error, and removed-column type fallback) — the previously-untested grid seam.

**Deferred (not blocking this pilot)**
- Per-cell error fan-out and uncapped widening on wide warehouse models — tracked by DRC-3629 / DRC-3630.
- Rare sibling-model column-name collision in scoping: low-probability, degrades to profiling an extra column.

even-wei:
# Code Review: PR #1411

**SHA** `daf66d7` · **Verdict** GO

No blockers or issues found across passes A–I (correctness, security, cross-reference, error handling, tests, diff, performance, async, architecture).

Key contracts verified against the codebase, not just the diff: `submitRun` without `nowait` does return the completed `Run` synchronously; the `ProfileDistributionOkResult`/TopK payload shapes (incl. `mode?: "counts"`, `(number | null)[]` counts) match what `InlineProfileDistributionCell` narrows on; network/timeout faults wrap as `HttpError(0, …)` so the hook's Sentry gating (capture 5xx/status-0, skip 4xx) holds; `cli.py` sets `new_cll_experience` in the last flag block with no later overwrite, mirroring the existing `whole_model_impact` pattern; and the storybook `/api/flag` mock fix is correct — `getServerFlag` returns the parsed body directly, so the old `{data: …}` envelope was double-wrapped.

## Notes
1. `js/packages/ui/src/components/data/InlineProfileDistributionCell.tsx:80` — `isDatetimeType`/`isTimeOfDayType` add a second dbt-type classification mechanism alongside `components/ui/DataTypeIcon/classifyType.ts`.
   Evidence: the substring heuristic is justified today — `classifyType`'s exact-match table returns `unknown` for DuckDB `TIMESTAMP_S/MS/NS`, which would break date formatting on the primary path — but two classifiers in one package can drift; consolidation candidate (e.g. extend `classifyType` and switch on its `date|datetime|time` categories). Pass A.
2. `js/packages/ui/src/utils/formatTime.ts:119` — `formatEpochSeconds` renders date-only, so a timestamp column whose spread is intra-day produces identical tooltips on every bin (e.g. "Jun 5, 2026 – Jun 5, 2026").
   Evidence: `toLocaleDateString` with year/month/day only; common for event-timestamp columns loaded in one batch. Graceful but information-free; consider adding time components when the edge span is sub-day. Pass A.

## Limits
- `tests/test_cli.py::test_cli_shows_update_available_warning` fails in this environment on **both** the PR branch and `origin/main` (Rich panel truncation splits "99.0.0" in the captured output) — pre-existing environment flake, not introduced by this PR. All other backend CLI tests pass (24/25), `flake8` clean.
- Live-browser verification was not re-performed here (author's DuckDB browser verify noted in the PR description).

## Verification
`pnpm type:check` ✓ · `pnpm lint` ✓ (650 files) · `pnpm test` ✓ (180 files, 3972 passed / 5 skipped) · `pnpm run build` ✓ · `uv run flake8 recce/cli.py` ✓

even-wei (approved):
Claude Code Review: No critical issues found. See review comment for details.

PR #1412, opened 2026-05-31
feat(telemetry): dual-write OSS CLI + frontend events to PostHog (dual-run)

even-wei:
**PR checklist**

- [x] Ensure you have added or ran the appropriate tests for your PR.
- [x] DCO signed

**What type of PR is this?**

`feat` (telemetry) — dual-run migration step.

**What this PR does / why we need it**:

Adds PostHog emission alongside Amplitude (**dual-run**) across the Recce OSS CLI and frontend, so product-usage data starts flowing into PostHog while Amplitude stays live. Part of the Amplitude→PostHog consolidation; the goal is to **collect data first** — no Amplitude removal in this PR.

- **P4 — OSS CLI dual-write:** new `recce/event/posthog_telemetry.py` fans out alongside the Amplitude collector for `command`, `api_event`, `load_state`, `codespace_instance`, `single_environment`, `performance`, `oss_connect_to_cloud` — tagged `event_source="oss-cli"`.
- **P6 — OSS frontend dual-emit:** `track.ts` adds a `posthog-js` capture next to each of the ~20 Amplitude wrappers — `event_source="oss-web"` (registered as a super-property so autocapture events are tagged too). Wrapper signatures unchanged → call sites untouched.

**Which issue(s) this PR fixes**:

None (migration workstream; no Linear issue).

**Special notes for your reviewer**:

- **Dual-run / safety:** Amplitude is fully intact — PostHog is purely additive. Removal is a later, gated phase.
- **Opt-out parity:** PostHog reuses the existing `anonymous_tracking` opt-out (CLI) and `recce_user_id` consent (frontend) — **no previously-untracked user starts being tracked**. `RECCE_DISABLE_TELEMETRY` kill switch added. `$process_person_profile=False` (CLI) / `person_profiles:'identified_only'` (FE) — anonymous, no person bloat.
- **autocapture ON, session replay OFF** (matches Amplitude autocapture; replay deliberately dropped).
- **geoip ENABLED on the CLI** (`disable_geoip=False`) so `country` resolves from the user IP — parity with Amplitude, powers analytics' `country_distribution`. ⚠️ This means PostHog ingests the **end-user machine IP** for geo on an OSS-installed CLI; flagging for privacy sign-off (one-line toggle if undesired).
[...]

PR #1415, opened 2026-06-02
refactor(telemetry): drop _cli suffix on OSS PostHog event names

even-wei:
## What & why

Follow-up to the merged #1412. Drops the `_cli` suffix on two OSS-CLI PostHog event names so every cross-surface event uses **one canonical name disambiguated by `event_source`**:

- `single_environment_cli` → **`single_environment`** (matches the oss-web `single_environment`)
- `oss_connect_to_cloud_cli` → **`oss_connect_to_cloud`**

This aligns them with the already-consistent shared events (`lineage_view_render`, `environment_config`, `navigation_change`) — bare names split by `event_source`. One-line change to `_PH_EVENT_MAP` + its test.

## Why now (timing)

#1412 merged with the `_cli` names, but **these two events have not been ingested by PostHog yet** (only `command` has so far). Landing this before the next OSS release avoids emitting `*_cli` to prod and then splitting the series across two names later.

## Test

`uv run pytest tests/event/` → 46 passed. Dual-write / opt-out / kill-switch behavior unchanged; only the two emitted names change.

Relates to DRC-3615, DRC-3621.


gcko (approved):
lgtm

PR #1421, opened 2026-06-08
refactor(lineage): fold whole-model-impact into the CLL experience

danyelf:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`refactor`

`--whole-model-impact` was a separate server flag whose only way to be enabled
also forced `--new-cll-experience` on. In practice whole-model impact is a
**subset of the CLL experience** — there's never a reason to have one without the
other. This PR removes the standalone flag and gates the whole-model-impact
surfaces on `new_cll_experience` instead, so the two always travel together:

- CLL on  → new-CLL visuals **and** whole-model impact render
- CLL off → neither renders

**Backend**
- Drop the `--whole-model-impact` Click option, its `flag` dict entry, and the
  implied-flag block in `recce/cli.py`.
- Remove the two obsolete CLI tests.

**Frontend** (full merge — the `wholeModelImpact` identifier is gone, not aliased)
- `RecceServerFlags` no longer carries `whole_model_impact`.
- The lineage context drops its `wholeModelImpact` field (the computed
  `wholeModelChangedNodeIds` / `wholeModelImpactedNodeIds` data sets stay).
- `computeImpactedSets` always computes the whole-model sets — it's only ever
  called when `newCllExperience && cll`.
- The title chip + left stripe (NodeView), graph badge (LineageNode), and
  change-category text-label suppression now gate on `newCllExperience` — which
  every one of those sites already had in scope, so it's a direct substitution.

- No behavior change for users already running `--new-cll-experience`: they now
  get the whole-model surfaces automatically (previously gated behind the extra
  flag).
- `recce/data/` (embedded frontend bundle) is **not** included here — it's a
  generated artifact regenerated by the normal build/release step.
- Verified end-to-end against a DuckDB two-env project with
  `--new-cll-experience`: `/api/flag` returns `new_cll_experience: true` with no
  `whole_model_impact` key; the graph renders `ADD`/`COLUMN` badges, suppresses
  the category text labels, and NodeView shows the amber whole-model-impacted
  title chip + stripe on a downstream-impacted model.
- Checks: `pnpm type:check` / `pnpm lint` / `pnpm run build` clean; full frontend
  suite 3979 passed; `pytest tests/test_cli.py` 23 passed; `black`/`isort`/`flake8`
  clean.

```
The whole-model impact highlighting (title chip, left stripe, column badges, and
suppression of change-category text labels) is now part of the new CLL experience
and enabled by `--new-cll-experience`. The separate `--whole-model-impact` flag
(and its `RECCE_WHOLE_MODEL_IMPACT` env var) has been removed.
```


danyelf:
# Code Review: PR #1421

**SHA** `1b39e111` · **Verdict** GO

Mechanical, well-scoped refactor. Verified the central claim, the rename completeness, and all quality gates locally.

## Verified
- `computeImpactedSets` (now always computing the whole-model sets) is reached only under `newCllExperience && cll` at both call sites — `LineageViewOss.tsx:635-638` and `:963-966`. No whole-model work runs when CLL is off; no perf regression for the CLL-off path.
- Rename is complete: no straggling `wholeModelImpact` prop / `whole_model_impact` flag / `--whole-model-impact` / `RECCE_WHOLE_MODEL_IMPACT` references remain in `js/` or `recce/`. The retained `wholeModelImpactedNodeIds` / `wholeModelChangedNodeIds` are the data-set identifiers the PR intentionally keeps.
- `LineageViewContextType` carries a single `newCllExperience` field — the rename collapsed the old pair without leaving a duplicate (`types.ts:261`).
- `GraphNodeOss.tsx:363` still threads `newCllExperience={newCllExperience}` into `LineageNode`, so graph badges keep rendering after the prop swap.
- Checks: `pnpm type:check` exit 0; the 5 affected lineage suites pass (132 tests); `pytest tests/test_cli.py` 23 passed.

## Notes
1. PR description — "No behavior change for users already running `--new-cll-experience`" is slightly misleading. A user who previously ran `--new-cll-experience` *without* the old `--whole-model-impact` now gets the whole-model surfaces (title chip, stripe, badges, label suppression) where before they did not. The next sentence ("they now get the whole-model surfaces automatically") states the real, intended change. Both flags are experimental, so this is acceptable — just noting the framing.
2. `usePublishedImpactSets.ts` — the stable `EMPTY_SET` reference now guards only the *initial* state. Post-`publish`, an empty whole-model result is a fresh `new Set()` from `computeWholeModelImpact`, so downstream memos keyed on `wholeModelChangedNodeIds` / `wholeModelImpactedNodeIds` recompute on every impact publish even when the sets are empty. Negligible (cheap lookups, fires only during impact analysis, and the component re-renders regardless because `nodeIds` always changes) — but it's a deliberately-removed half of the original optimization, worth a conscious confirm.

No blockers, no issues.

gcko (approved):
LGTM

PR #1423, opened 2026-06-09
perf(api): cache merged lineage served by /api/info

even-wei:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`perf` — backend performance optimization, no behavior change.

The `/api/info` endpoint rebuilt the merged lineage (`build_merged_lineage(...)`) on **every** request — ~7.3 ms at ~1058 models — even though its inputs are immutable between refreshes and the frontend refetches `/api/info` frequently (on window focus, and after every check run that invalidates the lineage query). This caches the merged lineage at the dbt adapter layer, keyed on the same artifact identity as the existing `get_lineage_diff()` cache and invalidated together on `refresh()`.

Audit context: the *payload-size* wins this issue originally targeted were already shipped (DRC-3263 stripped `raw_code`/`checksum`/`config`; DRC-3258 moved to the merged wire format). The `/api/info` wire payload is already lean (~0.43 MB @ 1058 models) and every remaining field is consumed by the frontend graph builder — so the remaining cost was compute, not bytes. This PR addresses that.

Resolves DRC-3326

- Wire format is **byte-identical** — same object, same `model_dump(exclude_none=True, by_alias=True)`. No API or frontend change.
- Cache lives at the adapter layer (`get_merged_lineage()`): a non-cached default in `BaseAdapter` (sqlmesh/other adapters unaffected) + a `@lru_cache(maxsize=1)` override in the dbt adapter, cleared in `refresh()` alongside the sibling caches.
- Regression test `test_get_merged_lineage_cached` asserts equivalence to a fresh merge, cache-instance identity on repeat, and invalidation on `cache_clear`.
- `+68 / −4` across 5 files.
- Deferred (out of scope, noted for follow-up): edge-payload id-interning and server-side gzip/zstd — both need a wire-contract change + frontend coordination.

NONE


even-wei:
# Code Review: PR #1423

**SHA** `6d02379` · **Verdict** GO

Multi-pass adversarial review (A correctness, B security, C cross-reference, D error-handling, E tests, F diff-specific, G performance, H concurrency, I architecture). No blockers, no issues.

## Notes
1. `recce/adapter/dbt_adapter/__init__.py:743` — The new `_get_merged_lineage_cached` mirrors the existing `_get_lineage_diff_cached` exactly: same `id()`-derived cache key, same `@lru_cache(maxsize=1)`, and it is additionally `cache_clear()`-ed in `refresh()` (line 2016). Invalidation is therefore *doubly* guaranteed — the `id()` key changes when `refresh()` reassigns the manifest/catalog objects, AND the explicit clear runs. Strictly safer than the sibling caches, which rely on the `id()` key alone. Pass A/D.
2. `tests/test_core.py:204` — The regression test asserts equivalence, instance identity, and invalidation via an explicit `cache_clear()` call. It does not exercise invalidation through a real `refresh()` (the actually-interesting `id()`-key behavior), but that path is the same as the already-shipped `get_lineage_diff` cache and is belt-and-suspendered by the explicit clear. Acceptable coverage. Pass E.

## Verification
- `flake8` on all 5 changed files: clean.
- `pytest tests/test_core.py`: 9 passed (incl. new `test_get_merged_lineage_cached`).
- `pytest tests/adapter/`: 135 passed.
- True PR diff (`origin/main...HEAD`): 5 files, +68/-4.

Pure internal compute-path refactor: `/api/info` wire payload is byte-identical (`build_merged_lineage` is a deterministic pure function of `LineageDiff`). `BaseAdapter.get_merged_lineage` is a concrete default, so non-dbt adapters (sqlmesh) inherit working behavior — adapter contract intact. No new error paths, no new input/attack surface, no novel concurrency race (the watcher-thread vs request-handler read pattern is identical to the existing lineage caches).

even-wei (commented):
GO — full review posted as an issue comment. Pure compute-path refactor, wire payload byte-identical, invalidation doubly-guaranteed; flake8 clean, adapter + core tests pass. 2 non-blocking notes. (Self-approve not permitted by GitHub; verdict is GO.)

wcchang1115:
# Code Review: PR #1423

**SHA** `f807a9ca` · **Verdict** GO

Cache usage and invalidation verified correct. The merged-lineage cache is keyed
on the same artifact identity (`id()` of the four manifest/catalog objects) as the
existing `_get_lineage_diff_cached`, and invalidates on every path that replaces
those artifacts: `refresh(path)` clears it explicitly (`dbt_adapter/__init__.py:2016`)
*and* the key changes; `load_artifacts()` and `import_artifacts()` invalidate via the
key alone (new objects → new `id()`), consistent with the sibling caches. No
stale-serving path found. Wire format is byte-identical — same `build_merged_lineage`,
same input, same `model_dump`. Lint clean; `tests/test_core.py` 9/9 pass.

## Notes
1. `recce/server.py:595` — `get_merged_lineage()` now returns a *shared* cached
   `MergedLineage` instance to every `/api/info` response (previously each request
   built a fresh object). Safe today: the only consumer is the read-only
   `model_dump(...)` at `:604`. Latent risk if a future caller mutates the returned
   object — that mutation would persist into the cache and leak across requests.
   Consider a comment on `_get_merged_lineage_cached` noting the result must be
   treated as immutable.
2. `tests/test_core.py:206` — the regression test exercises invalidation via a direct
   `_get_merged_lineage_cached.cache_clear()`, not via `refresh()`. The explicit
   `cache_clear()` in `refresh()` (line 2016) is therefore not guarded by a test.
   It's redundant with the `id()`-based key (refresh would invalidate anyway), so this
   is low-risk, but a `refresh()`-driven assertion would lock in the integration point.

wcchang1115 (approved):
Claude Code Review: No critical issues found. Cache usage and invalidation verified correct. See review comment for two non-blocking notes.

PR #1425, opened 2026-06-11
fix(grid): stable row IDs and live API handle

gcko:
## Summary

Two complementary fixes for the AG Grid wrapper component to improve row selection stability and handle API correctness:

1. **Use _index for stable row IDs in query results**: Query result rows are now tagged with `_index` by `dataFrameToRowObjects`, but the `getRowId` function wasn't checking for it. This caused AG Grid to lose row selections when a query re-ran and data was transformed with new random IDs. Now using `_index` as the second priority (after `__rowKey` but before `rowIndex`) to ensure stable IDs across re-renders.

2. **Make handle API live via getters**: Changed the handle API to return live values via getters instead of snapshots, ensuring consumers always get the current API reference and DOM element.

## Test Plan

- ✅ All 76 related test files pass (1724 tests, 5 skipped)  
- ✅ Added comprehensive unit tests for row ID priority levels: `__rowKey`, `_index`, `rowIndex`, and random fallback
- ✅ Full frontend test suite passes (3983 tests)
- ✅ Type checking passes
- ✅ Linting passes


kentwelcome (approved):
LGTM

PR #1428, opened 2026-06-16
fix(schema): resolve column change cause once (impacted vs changed)

danyelf:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` / `refactor` — frontend (schema view).

A column whose type changed **only because an upstream column's type changed** was badged `~` ("changed") instead of `!` ("impacted").  This PR fixes that.

Root cause: the changed-vs-impacted classification was re-derived independently at every surface that paints a column — the name-cell badge, the row background, and the tooltip — so they drifted (you could see an "impacted" signal in one place and "changed" in another on the same column).

This introduces **`getColumnChangeStatus`** as the single source of truth for a column's change *cause*, and wires the badge, the row background (`SchemaView.getRowClass`), and the tooltip to all consume it:

- A type delta on a downstream-impacted column with no local definition change is now **impacted (`!`)**, not a structural change (`~`).
- Reorder-only and locally definition-changed columns keep `~`.
- The type-**icon** transition stays intentionally decoupled — it renders on any type delta regardless of cause ("what the type became" ≠ "who caused it").

While in the same surface (found during visual verification):

- **HUGEINT icon** — DuckDB `HUGEINT` / `UHUGEINT` / `INT128` and unsigned integers now map to the integer type icon instead of the `[...]` unknown placeholder.
- **Type-transition alignment** — the base / arrow / current chips now center on a single row.
- **Column width** — the column-name column flexes so long names (`credit_card_amount`) are no longer truncated.

Closes DRC-3626

- The de-brittling is the heart of this: `getColumnChangeStatus` is tested once with a case table, and the three rendering surfaces now map from its result rather than re-deriving. This is the "classify once, render many" fix.
- The badge render collapses to one badge per column driven by the resolved status (previously a column could render both a structural `~` and a definition-changed `~`); single-badge behavior is the intended simplification.
- The datetime classifier in `InlineProfileDistributionCell` (`isDatetimeType`/`isTimeOfDayType`) is a *separate* parallel type-classifier — intentionally left to DRC-3669, not folded in here.
- Verification: `pnpm type:check`, `pnpm lint`, and `pnpm test` (4001 passing) are green. Visually confirmed on jaffle_shop (mega-snowflake / DuckDB): HUGEINT icon, full column names, and the type-transition alignment. The `~`→`!` flip is covered by unit tests; live it requires impact analysis to be engaged (the impacted-column set is populated interactively).

```release-note
Schema view: a column whose type changed only because of an upstream change is now marked "impacted" (!) rather than "changed" (~), consistently across the badge, row highlight, and tooltip. DuckDB HUGEINT and related wide/unsigned integer types now show the integer type icon. Long column names are no longer truncated.
```

Generated with [Claude Code](https://claude.ai/code)


gcko:
# Code Review: PR #1428

**SHA** `5e862920` · **Verdict** GO

This refactor extracts column change-cause resolution into a single
`getColumnChangeStatus` helper consumed by the name-cell badge, the tooltip,
and the row background — closing the divergence the PR targets (an "impacted"
badge sitting on a "changed"-coloured row).

Traced and confirmed:
- **Single source verified.** The badge's `isImpacted` prop (`schemaCells.tsx:42`,
  `isImpacted={row.isImpacted}`) and the row background (`SchemaView.tsx:468`,
  `getColumnChangeStatus(row, row.isImpacted)`) draw from the same field, so badge
  and background cannot disagree by construction.
- **Precedence equivalence holds.** `changeUnknown + type delta` → `changed`
  (matches old `hasStructuralChange` guard), `changeUnknown + impacted` → `unknown`,
  inherited type shift on impacted → `impacted` (the intended behavior change).
  All branches are covered by the table-driven `getColumnChangeStatus.test.ts`.
- **Double-`~` fixed.** Old code could render two `~` badges when a column was both
  structurally type-changed and definition-changed; the new `changed` branch renders
  exactly one.
- **No singleEnv divergence.** Single-env uses a separate component
  (`PrivateSingleEnvSchemaView`) whose `getRowClass` always returns `row-normal` and
  never calls `getColumnChangeStatus`; the cell's `singleEnv → "unchanged"`
  short-circuit only applies there, where no diff badges render.

Quality gates on `5e862920`: `pnpm type:check`, `pnpm lint`, and `pnpm test`
(4001 passed / 5 skipped) all green. `classifyType` additions (DuckDB wide/unsigned
integers) are additive and covered.

## Blockers
None.

## Issues
None.

gcko (approved):
Claude Code Review: No critical issues found. See review comment.

PR #1434, opened 2026-06-18
feat(schema): document [ALL] and [ADD] badges in schema legend (DRC-3466)

danyelf:
## Type

- [x] Enhancement

## Description

The Columns tab shows a small legend above the schema grid that explains the change markers a column can carry. Recently we added two newer lineage markers that render on the grid when the new CLL experience is enabled — `[ALL]` (whole-model) and `[ADD]` (additive) — but the legend never documented them. Users saw those badges with no in-product explanation of what they meant.

This PR extends the legend to cover both. The existing `+ added` / `- removed` / `~ changed` entries are unchanged; two new entries appear alongside them:

- **`[ALL]` whole-model** — every row in this model can be affected
- **`[ADD]` additive** — adds a column; existing rows unchanged

The chips reuse the exact styling and color tokens the lineage graph and the NodeView header already use (via `pickTitleChip` / `pickGraphBadge` + `TreatmentChip`), so the legend matches the badges a user actually sees on the canvas — no new colors were introduced. Both new entries are gated on the `new_cll_experience` server flag, the same flag that controls whether the badges render at all, so the legend never documents a marker the user can't see.

This is documentation for the CLL badge set that is being redesigned in DRC-3813 / DRC-3814 / DRC-3815; the copy and tokens here will track that work.

## Notes for reviewers

Two supporting fixes ride along with the legend change:

1. **Wire-enum lint allowlist.** The `check_wire_enum_literals.sh` guard (DRC-3553) flagged the legend's `changeCategory: "non_breaking"` as a legacy enum literal. It genuinely has to stay legacy: `pickGraphBadge`'s classifier keys the additive badge off the legacy `ChangeCategory` union (`"non_breaking"` → `additive`), and that union only accepts `breaking` / `non_breaking` / `partial_breaking` / `unknown` — the v2 term isn't a valid value there. The line is annotated with the sanctioned `// wire-enum-ok` marker rather than changed.
2. **Storybook QueryClient provider.** `SchemaLegend` now reads the server flag through `useRecceServerFlag` (react-query), and the Schema story's decorator renders `<SchemaLegend />` directly. Without a `QueryClientProvider` in the Storybook preview, every schema story threw "No QueryClient set". The preview now wraps all stories in a shared `QueryClient`, which restores the schema stories and lets the new legend entries render for review.

## Linked issues

- Closes DRC-3466 — Schema legend should cover the `[ALL]` and `[ADD]` badges
- Documents badges being redesigned in DRC-3813 / DRC-3814 / DRC-3815
- Follow-up from DRC-3341

## How tested

- `SchemaView` / `SchemaLegend` unit tests (8/8 pass): the always-on added/removed/changed entries; the whole-model and additive entries present with the correct copy and aria-labels when `new_cll_experience` is on; both hidden when it's off.
- `pnpm type:check` — clean.
- `biome check` — clean on all changed files.
- `check_wire_enum_literals.sh origin/main` — exits 0.
- Legend rendered in Storybook (new `LegendWithWholeModelAndAdditive` story) and captured below.

## Screenshots

The Columns-tab legend with `new_cll_experience` enabled, now including the `[ALL]` and `[ADD]` entries.

## User-facing changes

The Columns-tab legend now explains the `[ALL]` and `[ADD]` badges when the new CLL experience is enabled. No behavior change to the badges themselves.

## Checklist

- [x] Tests added/updated and passing
- [x] DCO sign-off on commits

### DIFF
--- a/js/packages/storybook/.storybook/preview.tsx
+++ b/js/packages/storybook/.storybook/preview.tsx
@@ -2,6 +2,7 @@ import { theme } from "@datarecce/ui/theme";
 import CssBaseline from "@mui/material/CssBaseline";
 import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
 import type { Preview } from "@storybook/react-vite";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import "chartjs-adapter-date-fns";
 import { useEffect } from "react";
 
@@ -21,6 +22,13 @@ if (typeof window !== "undefined") {
 // The @datarecce/ui theme uses CSS-variables mode with `colorSchemeSelector:
 // "class"`, so a single theme drives both light and dark modes — the global
 // decorator below toggles the `.dark` class on <html>.
+
+// Components that read server flags (e.g. SchemaLegend via useRecceServerFlag)
+// use react-query, so stories need a QueryClient in context. Retries off so
+// mocked-endpoint stories settle immediately.
+const queryClient = new QueryClient({
+  defaultOptions: { queries: { retry: false } },
+});
 
 const preview: Preview = {
   parameters: {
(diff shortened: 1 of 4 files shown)


danyelf:
Closing — wrong legend surface. Per captain review, the `[ALL]`/`[ADD]` badges are lineage-node badges, so their legend entry belongs on the **lineage view** (the `LineageLegend` `changeStatus` variant on the lineage canvas), not the Columns-tab `SchemaLegend` this PR modified. Redoing against the correct target under a fresh PR. The legend copy here is reusable. DRC-3466 has been rewritten to point at the lineage view.

PR #1435, opened 2026-06-18
chore(deps): consolidate dependabot updates

gcko:
## Summary

Consolidates all open Dependabot PRs into a single tested branch (each taken to
its latest compatible version), **plus** the safe tier of an out-of-band
dependency audit. Major bumps are intentionally deferred to dedicated PRs.

- Closes #1426 — Bump the all-minor-patch group in /js with 27 updates
- Closes #1429 — Bump pyjwt from 2.12.1 to 2.13.0
- Closes #1430 — Bump python-multipart from 0.0.29 to 0.0.31
- Closes #1431 — Bump cryptography from 46.0.7 to 48.0.1
- Closes #1432 — Bump starlette from 1.2.1 to 1.3.1

### Dependabot-driven — Python (`uv.lock`)

| Package | From | To | Security |
|---------|------|----|----------|
| pyjwt | 2.12.1 | 2.13.0 | **5 CVEs** (algorithm-confusion, allow-list bypass, DoS, SSRF, cache-wipe) |
| cryptography | 46.0.7 | 49.0.0 | — |
| python-multipart | 0.0.29 | 0.0.32 | — |
| starlette | 1.2.1 | 1.3.1 | — |

`cryptography` (direct use in `connect_to_cloud.py`: RSA-OAEP, key-gen, PEM)
and `starlette` (middleware/routing in `server.py`/`mcp_server.py`) use only
stable APIs untouched by the major/minor breaking changes.

### Dependabot-driven — Node (`#1426` group)

react/react-dom 19.2.6→19.2.7 · @types/react 19.2.15→19.2.17 ·
@tanstack/react-query 5.100.14→5.101.0 · @mui/material+@mui/system 9.0.1→9.1.1 ·
@amplitude/unified 1.1.9→1.1.16 · @amplitude/analytics-core 2.49.0→2.50.0 ·
@codemirror/merge 6.12.1→6.12.2 · @codemirror/view 6.43.0→6.43.1 ·
@sentry/nextjs+@sentry/react 10.56.0→10.58.0 · next+@next/third-parties 16.2.7→16.2.9 ·
posthog-js 1.382.0→1.387.0 · write-excel-file 4.1.0→4.1.1 · read-excel-file 9.1.1→9.2.0 ·
import-in-the-middle 3.0.1→3.0.2 · @types/node 25.9.2→25.9.3 ·
baseline-browser-mapping 2.10.34→2.10.37 · storybook+@storybook/* 10.4.2→10.4.6

[...]

PR #1440, opened 2026-06-18
chore(deps): bump rich, websockets, packaging to latest majors

gcko:
## Type

chore (dependencies)

## Description

Applies the **safe subset** of the major-version bumps deferred from the
Dependabot consolidation (#1435). `uv.lock` only — no `pyproject.toml`
constraint changes were needed; the existing floors already admit these.

### Applied

| Package | From | To | Notes |
|---------|------|----|-------|
| rich | 14.2.0 | 15.0.0 | Console/print/Table usage is stable across the major |
| websockets | 15.0.1 | 16.0 | used transitively via uvicorn; server uses `starlette.websockets`, not the library API |
| packaging | 25.0 | 26.2 | only `Version`/`parse`/`is_prerelease`/`base_version` used |

Each was verified **in isolation** to leave the dbt stack untouched (dbt-core
stays 1.10.15). `make test` passes.

### Deferred (with reasons)

| Package | Target | Why deferred |
|---------|--------|--------------|
| deepdiff | 9.1.0 | **Blocked upstream:** `dbt-common` requires `deepdiff<9.0,>=7.0`. Bumping forces dbt-core down 1.10.15 → 1.8.8. Hold at 8.x until dbt relaxes the cap (Recce's own `<10.0` floor already allows 9). |
| @babel/helpers + @babel/runtime | 8.0.0 | The OSS app pins these as a **dedup anchor**; all transitive consumers require `^7`, so bumping the floor splits the tree into two copies (7 + 8) rather than consolidating. Forcing 8 via override would run babel-7-compiled deps on an 8.0.0 runtime released only days ago. Defer to a dedicated PR after it soaks. |

## How root causes were found

`uv lock --upgrade-package <pkg>` was run one major at a time: only `deepdiff`
downgraded dbt-core (to 1.8.8); `rich`, `websockets`, and `packaging` each kept
dbt-core at 1.10.15. The deepdiff conflict was confirmed via
`dbt-common` metadata: `deepdiff<9.0,>=7.0`.

## Test plan

- [x] `uv lock --check` passes
- [x] Python `make test` passes (rich 15.0.0, websockets 16.0, packaging 26.2)
- [x] `recce.cli` / `recce.server` import cleanly under the new majors
- [x] dbt-core remains 1.10.15 (no stack downgrade)


kentwelcome (approved):
LGTM

PR #1441, opened 2026-06-18
chore(deps): unpin date-fns + reconcile biome guard

gcko:
## Type

chore (dependencies / CI)

## Description

Unpins `date-fns` (held at `~4.2.1` since #1435) now that the original test
break is no longer reproducible, and reconciles a biome guard that #1435 left
stale.

### date-fns: the hold is obsolete

`date-fns` was pinned because 4.3.0's "modularized optimization fallback" broke
`vi.mock("date-fns", …)` in `TimelineEvent.test.tsx`. I re-verified empirically
— the break **no longer reproduces**:

| date-fns | `TimelineEvent.test.tsx` | Full suite (4005 tests) |
|---|---|---|
| 4.3.0 (the blamed version) | ✅ passes | ✅ passes |
| 4.4.0 (latest) | ✅ passes | ✅ passes |

Root cause of the fix: #1435 bumped **Vite → 8** and **Vitest → 4.1.9**, whose
resolver handles date-fns's subpath `exports` correctly. The pin outlived the
bug.

- `js/pnpm-workspace.yaml`: `date-fns: ~4.2.1` → `^4.1.0` (keeps the dedup-anchor
  override per `js/CLAUDE.md`, drops the clamp) → resolves to **4.4.0**
- `.github/dependabot.yml`: removed the now-obsolete `date-fns` ignore
- `@datarecce/ui` dependency floor left untouched (`^4.1.0`)

### biome guard reconciliation (incidental, but necessary)

`js/.husky/pre-commit` pins `@biomejs/biome` via `EXPECTED_BIOME` and blocks any
`js/package.json` commit where the staged value differs. #1435 bumped biome to
**2.5.0** (above the buggy 2.4.12–2.4.16 stack-overflow range) but **did not
update `EXPECTED_BIOME`**, which the hook's own docs require ("update
EXPECTED_BIOME in the same commit as the bump"). As a result the guard was
stale — it blocked *this* PR and would block any `js/package.json` change on
current main.

Set `EXPECTED_BIOME=2.5.0` after running the hook's prescribed macOS-arm64
verification: `biome check src/` and `biome check packages/ui/src/` both exit 0
with no stack overflow.

## Test plan

- [x] `pnpm test` — 182 files / 4005 tests pass (mock unchanged, date-fns 4.4.0)
- [x] Also verified the suite passes at date-fns 4.3.0 (the originally-blamed version)
- [x] `pnpm type:check`, `pnpm lint`, `pnpm run build` pass
- [x] `CI=true pnpm install --frozen-lockfile` passes
- [x] `biome check` clean on macOS arm64 (no stack overflow) at 2.5.0


kentwelcome (approved):
LGTM

PR #1446, opened 2026-06-23
feat(adapter): fail loud on dbt v2/Fusion (v20) manifests

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`feat`: a fail-loud guard for unsupported dbt artifact schemas.

dbt v2 / Fusion emits manifest schema v20; Recce's bundled dbt 1.x reads up to manifest v12 / catalog v1. Feeding a Fusion artifact to Recce today fails with dbt's internal schema error, or an unrelated `ValueError` on the cloud path; nothing tells the user that Fusion isn't supported.

This PR detects the unsupported schema at artifact load (both local `target/` files and cloud-loaded state) and raises `UnsupportedDbtSchemaError`:

> dbt v2 / Fusion manifests (schema v20) are not yet supported by Recce. Recce supports dbt 1.x artifacts (manifest schema up to v12, catalog v1). Re-generate the manifest with dbt 1.x.

This PR only detects and reports. It does not parse v20 artifacts or handle v12-vs-v20 cross-environment diffs.

NONE

- If dbt 1.x ever bumps its manifest schema past v12, `_DBT1X_MAX_SCHEMA` needs a manual bump. The rationale for the fixed ceiling is in the comment above the constant.
- The `path=` branch catches and translates dbt's `IncompatibleSchemaError` instead of pre-checking the version, to avoid re-parsing large manifests and to leave artifacts dbt accepts untouched.
- No separate Cloud-side change is needed: Recce Cloud precompute runs OSS Recce in a container.

Yes. Feeding a dbt v2 / Fusion (v20) artifact now raises a Recce error naming Fusion and the supported versions, instead of dbt's internal schema-version error.


kentwelcome (approved):
LGTM

PR #1447, opened 2026-06-24
fix(lineage): guard base/current in useModelColumns for added/removed nodes

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Bug fix

Clicking a newly-added model in the lineage view crashed with `TypeError: Cannot read properties of undefined (reading 'columns')`.

The model-info API omits a side's key when a node is missing from one environment of the diff (added → no `base`, removed → no `current`). `useModelColumns`'s `useMemo` guarded only `modelInfo` with optional chaining, then read `modelInfo.base.columns` directly — so `modelInfo.base` was `undefined` for an added model and reading `.columns` threw. TypeScript didn't catch it because `ModelInfoResult.model.base/current` were typed non-optional while the wire format omits one.

This PR builds the base/current column lists independently with optional chaining and unions them, and makes `ModelInfoResult.base/current` optional so the type matches the API. Added/removed models now render their columns instead of crashing (or showing an empty list).

Linear DRC-3744

- No api_server change is needed: omitting the absent side's key is intentional and documented, and most frontend code already treats a missing key as no data (e.g. `modelDetail?.base?.columns ?? {}`). This hook was the one place that didn't.
- Reproduced the crash locally against the real base-less response shape, then confirmed the fixed build renders the added model's columns. `vitest related` green incl. the 2 new regression tests; `pnpm type:check` clean.

Fix a crash when selecting an added or removed model in the lineage view; such models now display their columns.

### DIFF
--- a/js/packages/ui/src/api/info.ts
+++ b/js/packages/ui/src/api/info.ts
@@ -252,8 +252,9 @@ export interface ModelEnvDetail {
  */
 export interface ModelInfoResult {
   model: {
-    base: ModelEnvDetail;
-    current: ModelEnvDetail;
+    // optional: added nodes have no base, removed nodes have no current
+    base?: ModelEnvDetail;
+    current?: ModelEnvDetail;
   };
 }
 
--- a/js/packages/ui/src/hooks/__tests__/useModelColumns.test.tsx
+++ b/js/packages/ui/src/hooks/__tests__/useModelColumns.test.tsx
@@ -120,6 +120,56 @@ describe("useModelColumns", () => {
     expect(result.current.error).toBe(null);
   });
 
+  // Regression: guard against the crash on an added model (no base key).
+  it("returns current columns (no throw) for an added model with no base", async () => {
+    mockGetModelInfo.mockResolvedValue({
+      model: {
+        current: {
+          columns: {
+            id: { name: "id", type: "INT" },
+            name: { name: "name", type: "TEXT" },
+          },
+          primary_key: "id",
+        },
+      },
+    });
+
+    const { result } = renderHook(() => useModelColumns(MODEL_NAME), {
+      wrapper: createWrapper(),
+    });
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(result.current.columns.map((c) => c.name)).toEqual(["id", "name"]);
+    expect(result.current.primaryKey).toBe("id");
+    expect(result.current.error).toBe(null);
+  });
+
+  it("returns base columns (no throw) for a removed model with no current", async () => {
+    mockGetModelInfo.mockResolvedValue({
+      model: {
+        base: {
+          columns: { id: { name: "id", type: "INT" } },
+          primary_key: "id",
+        },
+      },
+    });
+
+    const { result } = renderHook(() => useModelColumns(MODEL_NAME), {
+      wrapper: createWrapper(),
+    });
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(result.current.columns.map((c) => c.name)).toEqual(["id"]);
+    expect(result.current.primaryKey).toBe("id");
+    expect(result.current.error).toBe(null);
+  });
+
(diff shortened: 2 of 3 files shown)


gcko (approved):
**Pass 1 — adversarial review (first pass, no prior captain feedback).**

### Scope match vs DRC-3744 — DELIVERS the scoped fix ✅

DRC-3744 prescribes exactly three recce-repo changes, and the PR delivers all three with no scope creep:

1. **Hook fix** — `useModelColumns.tsx` lines 112–124 now builds `baseColumns` and `currentColumns` independently, each guarded with optional chaining (`modelInfo?.base?.columns ? Object.values(...) : []`), then unions them. Matches the ticket's prescription verbatim.
2. **Type fix** — `info.ts` lines 253–259 makes `ModelInfoResult.model.base` and `.current` optional (`base?`/`current?`), so the type now matches the wire format (added node omits `base`, removed node omits `current`). Matches ticket item 2.
3. **Regression tests** — `useModelColumns.test.tsx` lines 121–170 add the added-model (no `base` key) and removed-model (no `current` key) cases. Matches ticket item 3.

No api-server change was made, consistent with the ticket's explicit "No api server change is needed" note. The PR is correctly scoped — it fixes the crash and nothing else.

### CRITICAL / HIGH

None.

### MEDIUM

None.

### LOW

- **LOW — `useModelColumns.tsx:122-123` primaryKey precedence is sound but untested for the both-sides-present mismatch case.** `primaryKey: modelInfo?.current?.primary_key ?? modelInfo?.base?.primary_key` correctly prefers current, falls back to base for removed nodes. Behavior is right; there is no test asserting the fallback-to-base path actually returns the base PK when `current.primary_key` is `undefined` but `base.primary_key` is set. Minor coverage gap, not a correctness defect.
- **LOW — `useModelColumns.tsx:113-119` the empty-payload edge (`model: {}` — neither side present) is handled but untested.** With both keys absent the memo returns `columns: []`, `primaryKey: undefined`, no throw — correct and safe, but no regression test pins it. The two added tests cover the two ticket-named cases only.

### NIT

- **NIT — `unionColumns` docstring vs behavior mismatch (pre-existing, NOT touched by this PR).** `useModelColumns.tsx:18-20` docstring says "Columns from current take precedence if both exist," but the implementation (lines 24–33) pushes base first and adds a current column only if its name is not already present — i.e. base-first-wins on duplicate names. For nodes present in both envs whose column metadata differs, the base entry's metadata is kept. This predates the PR (the function body is unchanged in the diff) and is out of scope; flagging only because it sits in the touched file. No action required for this PR.

### Notes (positive / defensive verification)

- The OLD guarded branch was itself crash-prone: `primaryKey: modelInfo?.current.primary_key` would throw for a removed node (no `current` key), because the optional chain protected only `modelInfo`. The new code removes that latent second crash as well — the fix is more complete than a minimal patch.
- Verified the type widening (non-optional → optional) does **not** break the other four `getModelInfo` consumers: `NodeViewOss.tsx:181-182`, `SchemaDiffView.tsx:260-277`, and `SchemaSummary.tsx:61-77` all already read `?.base?.columns` / guard with `?.base &&`. The change validates their existing defensive code rather than introducing new type errors. No non-optional `model.base.columns` / `model.current.columns` reads remain anywhere in `js/packages/ui/src` (the only two `current.columns` hits outside tests are unrelated types — a run DataFrame in `utils/csv/extractors.ts` and column-level-lineage in `components/lineage/lineage.ts`).
- The two new tests are genuine regressions: their mocked payloads (`{ model: { current: {...} } }` and `{ model: { base: {...} } }`) reproduce the exact wire shape that threw on the old code, so they fail pre-fix and pass post-fix.

### Summary

This is a tight, correctly-scoped bug fix that delivers DRC-3744 exactly as written: it guards the `base`/`current` reads in `useModelColumns` with optional chaining and unions both sides, makes the `ModelInfoResult` base/current properties optional to match the wire format, and adds the two prescribed regression tests. The fix also incidentally closes a second latent crash in the old guard branch's `primaryKey` read. No correctness, security, or coverage-drift issues were found; the only observations are LOW/NIT (two untested edge cases that are nonetheless handled, and a pre-existing docstring/behavior mismatch in the unmodified `unionColumns`). Recommended verdict: APPROVE.

PR #1450, opened 2026-06-25
fix(mcp): cap lineage_diff view_mode=all and compact tool-result JSON (DRC-3758)

iamcxa:
## Problem (root cause of DataRecce/recce-cloud-infra DRC-3756)

`lineage_diff` with `view_mode: "all"` returns the entire project DAG. On a large warehouse this overflowed the Claude Agent SDK MCP output cap (`MAX_MCP_OUTPUT_TOKENS`, default 25,000 tokens) at **527,044 characters**. The Recce summary agent runs MCP-only (no file tools), so it cannot read the SDK's "output saved to file" fallback — the lineage silently vanished and the PR summary broke.

## Fix — two bounds, applied to BOTH backends

1. **Cap `view_mode: "all"`** at `VIEW_ALL_MAX_NODES` (300) in **both** `RecceMCPServer._tool_lineage_diff` (local mode) **and** `CloudBackend._tool_lineage_diff` (cloud mode) — `call_tool` delegates to the cloud backend before the local path, so both must be capped. Keeps the nodes that matter for analysis — **changed + impacted first** — then fills to the cap. When capped, the result carries `truncated`, `total_nodes`, and `returned_nodes`. Edges to dropped nodes are excluded by the existing `id_to_idx` guard, so the graph stays consistent. `changed_models` is **never** capped (already bounded to the impact set).
2. **Compact tool-result JSON** — drop `indent=2` for tight separators in the single `call_tool` serialization point. Consumers `json.loads` the payload, so it is runtime-safe; two `test_mcp_cloud_backend.py` tests that string-matched the pretty form were updated to the compact form (the full mcp suite was checked for other string-match breakage).

## Why cap = 300 (measured, not guessed)

Hard limit = Claude Code `MAX_MCP_OUTPUT_TOKENS` default **25,000 tokens**. Measured serialized size of a capped 300-node result with the compact separators this PR ships:

| node_id length | chars | est. tokens (~3 chars/tok) | % of 25k cap |
|---|---|---|---|
| realistic ~45 | 32,726 | ~10,900 | 44% |
| worst-case ~80 | 43,226 | ~14,400 | 58% |

Even worst-case stays at ~58% of the cap — comfortable headroom, cannot overflow.

## Tests

223 mcp tests pass (`test_mcp_server.py` + `test_mcp_cloud_backend.py` + `test_mcp_e2e.py`), including new coverage that:
- `view_mode: "all"` truncates to the cap, keeping changed + impacted and dropping unrelated nodes, with the `truncated` / `total_nodes` / `returned_nodes` markers — for **both** the local and cloud backends;
- `changed_models` is not capped even when its count exceeds the cap.

`black` / `isort` / `flake8` clean.

## Review

codex review, 2 rounds: round 1 caught (a) `CloudBackend._tool_lineage_diff` was uncapped and (b) the compact-JSON change broke two string-match tests — both fixed; round 2 found **0 findings**.

Known edge case (accepted): if more than 300 nodes are themselves changed/impacted, some are still dropped; the markers report the counts but not which class was trimmed.

## Cross-repo

This is the root-cause fix; it pairs with two recce-cloud-infra changes for the same incident (Linear DRC-3756):
- the summary-agent runner guard that makes an oversized-output run fail loud instead of silently empty (DRC-3756, recce-cloud-infra #1479);
- the agent prompt that steers `lineage_diff` to `changed_models` and forbids `view_mode: "all"` (DRC-3757, recce-cloud-infra #1480).

Resolves Linear DRC-3758.

### DIFF
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -55,7 +55,8 @@ SINGLE_ENV_WARNING = (
 # (MAX_MCP_OUTPUT_TOKENS, default 25,000 tokens). A full-project lineage on a large
 # warehouse overflowed at 527K chars (DRC-3756); the summary agent has no file
 # tools to read the SDK's file-pointer fallback, so the lineage silently vanished.
-# ~300 nodes keeps the worst-case serialized result under ~75% of the 25k cap.
+# ~300 nodes keeps the worst-case serialized result at ~58% of the 25k cap
+# (measured: ~43k chars / ~14.4k tokens for 300 long-id nodes, edges included).
 VIEW_ALL_MAX_NODES = 300
 
 
@@ -285,10 +286,8 @@ class CloudBackend:
 
         selected_nodes = {node_id: node for node_id, node in nodes.items() if node_id in selected}
 
-        # DRC-3758: bound view_mode="all" so the serialized result cannot exceed the
-        # Claude Agent SDK MCP output cap (see VIEW_ALL_MAX_NODES). Mirrors the
-        # local RecceMCPServer path. Keep changed + impacted nodes first; edges to
-        # dropped nodes are excluded by the id_to_idx guard below.
+        # DRC-3758: bound view_mode="all" (changed and impacted first) — see
+        # VIEW_ALL_MAX_NODES. Mirrors the local path; edges to dropped nodes excluded below.
         view_mode = arguments.get("view_mode", "changed_models")
         truncated = False
         total_node_count = len(selected_nodes)
@@ -651,7 +650,7 @@ class RecceMCPServer:
                                 "type": "string",
                                 "enum": ["changed_models", "all"],
                                 "default": "changed_models",
-                                "description": "View mode: 'changed_models' for only changed models (default), 'all' for all models. On large projects, 'all' is capped to the most relevant nodes (changed and impacted first); when capped, the result carries 'truncated', 'total_nodes', and 'returned_nodes'. Prefer 'changed_models', which already includes the impacted models.",
+                                "description": "View mode: 'changed_models' for only changed models (default), 'all' for all models. On large projects, 'all' is capped to the most relevant nodes (changed and impacted first); when capped, the result carries 'truncated', 'total_nodes', and 'returned_nodes'. Edges to dropped nodes are omitted to keep the graph consistent, so a returned node may have upstream/neighbors that are not shown — a capped graph's connectivity is not complete lineage. Prefer 'changed_models', which already includes the impacted models.",
                             },
                         },
                     },
@@ -1495,11 +1494,8 @@ class RecceMCPServer:
                         if materialized is not None:
                             nodes[node_id]["materialized"] = materialized
 
-        # DRC-3758: bound view_mode="all" so the serialized result cannot exceed the
-        # Claude Agent SDK MCP output cap (see VIEW_ALL_MAX_NODES). Keep the nodes
-        # that matter for analysis — changed + impacted — first, then fill to
-        # the cap. Edges to dropped nodes are excluded by the id_to_idx guard in the
-        # edge-building loop below, so the returned graph stays internally consistent.
+        # DRC-3758: bound view_mode="all" (changed and impacted first) — see
+        # VIEW_ALL_MAX_NODES. Edges to dropped nodes are excluded by the id_to_idx guard.
         truncated = False
         total_node_count = len(nodes)
         if view_mode == "all" and total_node_count > VIEW_ALL_MAX_NODES:
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -284,6 +284,25 @@ class CloudBackend:
         impacted = set((await self._request("POST", "select", json={"select": "state:modified+"})).get("nodes", []))
 
         selected_nodes = {node_id: node for node_id, node in nodes.items() if node_id in selected}
+
+        # DRC-3758: bound view_mode="all" so the serialized result cannot exceed the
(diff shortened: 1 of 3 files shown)


wcchang1115 (changes requested):
# Code Review: PR #1450

**SHA** `4dbcea1` · **Verdict** NO-GO

## Blockers
None.

## Issues
1. `recce/mcp_server.py:~654` (tool description) — Truncated `view_mode="all"` drops edges to removed nodes, so a returned node can have hidden upstream/neighbors, but the description doesn't say so. An agent reading the graph will treat the connectivity as complete lineage. Add a caveat after the existing markers sentence:
   `"...when capped, the result carries 'truncated', 'total_nodes', and 'returned_nodes'. Edges to dropped nodes are omitted to keep the graph consistent, so a returned node may have upstream/neighbors that are not shown — connectivity is not a complete lineage."`
   The fix is the caveat text only — do not change truncation behavior or synthesize bridge edges. Per CLAUDE.md "tool description = agent contract." Pass A.

## Notes
1. Headroom number — the `VIEW_ALL_MAX_NODES` comment (`:~58`) says worst case is `under ~75% of the 25k cap`, while the PR body's measured table says ~58%; the two disagree. Confirm the measured 32,726 / 43,226-char sample included the `edges` array (not just node rows), then make the comment state the verified value, e.g. `"~300 nodes keeps the worst-case serialized result at ~58% of the 25k cap."` (Edges are small integer pairs, <10k chars for 300 nodes, so the cap stays safe either way.) Pass F / G.

2. Comments — the two per-site comments (cloud `:~290`, local `:~1498`) repeat the rationale already at `VIEW_ALL_MAX_NODES` (both already say "see VIEW_ALL_MAX_NODES") and have drifted from each other ("that matter for analysis" vs "internally consistent"). Trim each to 2 lines, delegating the WHY to the constant; keep the module-level comment. Fold in the wording fixes while trimming: "silently vanished" → "was dropped from the result"; drop "the nodes that matter for analysis" and "the most relevant nodes" (redundant with the "(changed and impacted first)" parenthetical). Pass F.

## Limits
Reviewed from the PR diff plus current `main` source; did not check out the branch or run the 223-test suite. Index/consistency and definition-order (`view_mode`, `impacted`, `diff_info` all defined before use) verified by reading; not by execution.

iamcxa:
Thanks for the thorough review @wcchang1115 — all addressed in `fdd6161b`.

**Issue 1 (tool-description caveat) — done.** Added the connectivity caveat to the `lineage_diff` description: a capped `view_mode="all"` graph omits edges to dropped nodes, so a returned node may have upstream/neighbors that aren't shown and connectivity is not complete lineage. Caveat text only — no behavior change, no synthesized bridge edges.

**Note 1 (headroom number) — done.** Corrected the `VIEW_ALL_MAX_NODES` comment from `~75%` to the measured `~58%`. Confirmed the measurement included the `edges` array (450 edge pairs for the 300-node sample), so the 32,726 / 43,226-char figures already account for edges — as you noted, the edge pairs are tiny either way.

**Note 2 (comment trim) — mostly done, one wording kept on purpose.** Trimmed both per-site comments to 2 lines, delegating the WHY to the constant. One exception: I kept the module comment's "the lineage silently vanished" instead of "was dropped from the result" — that sentence describes the **DRC-3756 incident** (a silent failure where the agent lost the lineage), not the cap's deliberate truncation, so "was dropped from the result" would misstate what happened. The cap's intentional drop is described separately.

**codecov — done.** Removed the unreachable fallback branch in the cloud cap test's `_request` stub.

Re your Limits note: the full mcp suite (incl. the new cap coverage on both the local and cloud backends) plus black/isort/flake8 pass locally; CI will confirm. Ready for another look.

wcchang1115:
# Code Review: PR #1450

**SHA** `fdd6161` · **Verdict** GO · **Incremental** (after `4dbcea1`)

`fdd6161` is the only commit since `4dbcea1`: doc/comment refinement plus one test-mock simplification. No production logic changed — the capping, edge-exclusion, and compact-JSON code from `0a1fd78`/`4dbcea1` is byte-identical. I re-verified that the strengthened tool-description and comment claims accurately describe the existing implementation in both the `CloudBackend` and `RecceMCPServer` paths:

- "Edges to dropped nodes are omitted" — confirmed at `mcp_server.py:334` (`source in id_to_idx and target in id_to_idx`) and `:1546-1548`.
- "result carries `truncated`, `total_nodes`, `returned_nodes`" — confirmed at `:341-344` and `:1565-1568`.
- "~58% of the 25k cap (measured ~43k chars / ~14.4k tokens)" — internally consistent (14.4k / 25k token cap ≈ 58%).
- Caveat correctly scoped to the `lineage_diff` tool (`:649`); the other `view_mode` at `:867` is `select_nodes` (returns IDs, no edges) and is correctly left untouched.

All 39 tests in `tests/test_mcp_cloud_backend.py` pass. No regression, no new issue.

## Blockers
None.

## Issues
None.

## Notes
1. `tests/test_mcp_cloud_backend.py:165` — The mock was loosened from a `path == "select"` branch (with a `return {}` default) to returning `{"nodes": ["b"]}` for *every* non-`info` path. Behaviorally identical today because `_tool_lineage_diff` only issues `info` and `select` requests, but if a future `_request("…", "<other>")` is added to that method, this mock will silently feed it the impacted-set payload instead of an empty dict. Non-blocking; the inline comment documents the intent.

wcchang1115 (approved):
Claude Code Review: No critical issues found (incremental review after 4dbcea1).

PR #1452, opened 2026-06-30
test(lineage): hygiene cleanup from PR #1381 review (DRC-3525)

danyelf:
## What this PR does

Clears the leftover hygiene items from the PR #1381 review (Linear DRC-3525). These are low-risk, self-contained chores — two new tests, one color-hardcoding cleanup, and a guard against a palette that was being kept in sync by hand. **No product behavior changes.**

## The items

**1. Test the whole-model badge colors.** The helper that picks the palette tokens for the lineage badges (`tokensForKind`) had no tests at all. Added one that locks its output for every badge kind in both light and dark mode, so a future color tweak can't silently break them.

**2. (Deferred) The `RECCE_WHOLE_MODEL_IMPACT` CLI test.** Skipped on purpose. That environment variable was never actually adopted — the real feature flag is `RECCE_NEW_CLL_EXPERIENCE` / `--new-cll-experience`. Writing a test for a name that doesn't exist would just bake in confusion. This needs a small product decision (point the test at the real flag, or drop the item), not a mechanical test, so it's left out for now.

**3. Two missing edge-case tests for impact propagation.** `computeWholeModelImpact` was covered for the common cases but not for (a) a lineage graph with disconnected pieces — impact shouldn't leak across them — or (b) a changed node that isn't in the graph at all, which shouldn't crash. Both added.

**4. Remove hardcoded colors from the lineage node.** `LineageNode.tsx` had raw hex values (`#1e1e1e`, `#ffffff`, `#000000`) repeated inline for the card background and text. Replaced them with the shared design tokens the rest of the lineage UI already uses, selected by the existing dark-mode flag. One small, deliberate normalization: the dark card moves from `#1e1e1e` to the shared neutral `#262626`, which matches the sibling node and MUI's dark surface — the old value was the odd one out. Visually imperceptible.

**5. Stop the CLL palette from silently drifting.** The palette is currently written twice — once as TypeScript constants for the React canvas, once as CSS variables for the schema grid — with a comment admitting the two are kept in sync by hand. Rather than the bigger refactor (generating one side from the other), this adds a test that parses the CSS and asserts the two match, so any future drift fails CI. The proper single-source fix (emit the CSS variables from the TS constants) touches the schema view, so it's left as a follow-up; the guard makes the duplication safe in the meantime.

## Tests
All green: 554 lineage tests pass (including the three new/updated files), `pnpm type:check` clean, Biome lint clean.

## User-facing changes
None — tests and internal color sourcing only. The dark lineage-node card shifts `#1e1e1e` → `#262626`, which is negligible and now consistent with its sibling nodes.

### DIFF
--- a/js/packages/ui/src/components/lineage/__tests__/cllPaletteSync.test.ts
+++ b/js/packages/ui/src/components/lineage/__tests__/cllPaletteSync.test.ts
@@ -1,0 +1,168 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+import {
+  cllAdditiveAccent,
+  cllAdditiveBadgeBg,
+  cllAdditiveBadgeFg,
+  cllChangedAccent,
+  cllChangedBadgeBg,
+  cllChangedBadgeFg,
+  cllChangeStatusBackgroundsDark,
+  cllChangeStatusBackgroundsLight,
+  cllChangeStatusColors,
+  cllImpactedAccent,
+  cllImpactedBadgeBg,
+  cllImpactedBadgeFg,
+} from "../styles";
+
+/**
+ * Single-source-of-truth guard for the CLL palette.
+ *
+ * The CLL colors are declared twice: as TS consts in `lineage/styles.tsx`
+ * (consumed by the React lineage canvas) and as `--schema-*` CSS custom
+ * properties in `schema/style.css` (consumed by the ag-grid schema view).
+ * Both files carry "keep in sync by hand — no build-time check" comments.
+ *
+ * This test IS that build-time check: it parses `schema/style.css` and asserts
+ * every TS const matches its CSS counterpart, so drift on either side fails CI.
+ * The TS consts are treated as canonical; the CSS values are verified against
+ * them. (A fuller consolidation — deriving the CSS vars from TS at runtime — is
+ * deferred; see DRC-3525.)
+ */
+
+const cssPath = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  "../../schema/style.css",
+);
+const css = readFileSync(cssPath, "utf8");
+
+/** Extract `--name: value` declarations from the first block matching a header. */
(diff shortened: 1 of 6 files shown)


gcko (approved):
- **LOW**: DRC-3525 item 2 remains intentionally unimplemented. The ticket asked for a CLI env-var test for `RECCE_WHOLE_MODEL_IMPACT=true`, while the PR body says that name was never adopted and defers the item pending a product decision. I did not see any changed file add equivalent CLI flag/env coverage, so this is acceptable only if the verdict keeps that decision tracked instead of treating the ticket as fully done.
- **LOW**: DRC-3525 item 5 says to consolidate the duplicated CLL palette, but this PR leaves the duplication in place and adds a drift guard instead. The new test documents that the colors are still declared twice in `js/packages/ui/src/components/lineage/__tests__/cllPaletteSync.test.ts` line 21 and that the fuller consolidation is deferred around lines 26-30. The guard substantially lowers drift risk, but it is still a scope tradeoff rather than the requested single-source fix.

danyelf:
Follow-up on the two **LOW** notes from the approving review — both are accepted scope tradeoffs, and the deferrals are now tracked on the linked Linear issue **DRC-3525** so the ticket isn't treated as fully done:

- **Item 2 (CLI env-var test):** deferred pending a product decision. Recorded on DRC-3525 that the requested `RECCE_WHOLE_MODEL_IMPACT` env var was never adopted (live gate is `RECCE_NEW_CLL_EXPERIENCE` / `--new-cll-experience`), so the item needs a decision on whether to test the real flag or close as covered-in-spirit by existing flag-propagation tests.
- **Item 5 (CLL palette consolidation):** deferred as a follow-up. Recorded on DRC-3525 that this PR ships a drift guard (`cllPaletteSync.test.ts`) rather than the single-source consolidation; the guard lowers drift risk and the fuller fix is tracked.

The failing `wire-enum-lint` check is fixed in 4f9e5f3 — the two legacy `changeCategory` test inputs get the documented `// wire-enum-ok` allowlist marker, since `wholeModelTreatment.ts` still branches on those legacy values during the deprecation window (rewriting them to v2 vocabulary would stop exercising the real code path).

PR #1457, opened 2026-07-03
DRC-3025 Float-Precision Rework

danyelf:
The data-diff grid no longer flags float-precision noise like `0.1+0.2` vs `0.3` as a change, so users stop chasing phantom diffs.

## What changed
- Route live diff-grid row status, counts, and the `changedOnly` filter through epsilon-aware `isCellChanged`.
- Route side-by-side cell highlighting (`createCellClassBase`/`Current`) through the same comparison.
- Drop raw `_.isEqual` value comparisons in `rowBuilders.ts` and `toDiffColumn.tsx`.
- Add float-noise regression tests on the live `buildDiffRows` and cell-class paths.

## Evidence
- Frontend suite: 4040/4040 passed (5 skipped), independently re-run at aggressive-self-review.

## Review guidance
- `isCellChanged` is now the single source of truth for value diffs — confirm no comparison path bypasses it.

---
Audit: DRC-3025 Float-Precision Rework · entity `fg` (tracked in local spacedock docs)
Closes DRC-3025


danyelf:
Superseding this PR per captain review. Reworking `isCellChanged` over lodash `isEqualWith(a, b, floatComparator)` — applies the epsilon at every numeric leaf (fixes a latent nested-float gap where object/array cells still compared exactly) and internalizes the epsilon constant, dropping a dead public export (`FLOAT_RELATIVE_EPSILON` was re-exported through two barrels with zero consumers). A fresh PR with a clean diff will follow. Closing to keep the review thread readable.

PR #1458, opened 2026-07-04
Float-precision-aware cell change detection in the data-diff grid (DRC-3025)

danyelf:
The data-diff grid no longer flags float-precision noise like `0.1+0.2` vs `0.3` as a change — including floats nested inside JSON/array/struct cells — so users stop chasing phantom diffs.

## What changed
- Rebuild `isCellChanged` on lodash `_.isEqualWith` with a float-leaf comparator: a magnitude-relative epsilon applied at every numeric leaf, scalar and nested.
- Route the live diff grid — row status, counts, the `changedOnly` filter, and side-by-side cell highlighting — through it, replacing raw `_.isEqual`.
- Keep the near-zero absolute floor; strings/booleans/mixed-types/null/NaN stay exact.
- Add float-noise regression tests across the live grid paths.

## Evidence
- Frontend suite: 4081/4081 passed (5 skipped), independently re-run at validation and adversarial review.

## Review guidance
- `isCellChanged` is the single source of truth for value diffs; the epsilon is a private leaf comparator, no longer exported.

---
Audit: DRC-3025 Float-Precision Rework · entity `fg` (tracked in local spacedock docs)
Closes DRC-3025


danyelf:
Closing without merge. DRC-3025 has been rediagnosed: the root cause is that numeric Decimals serialize to the frontend as untyped JSON strings, not the profile float-noise this PR addressed. The profile-scoped float coercion here is superseded by a new entity that fixes numeric type serialization at the DataFrame layer (see the reframed DRC-3025 description). The epsilon-comparison logic will be migrated forward into that work.

PR #1462, opened 2026-07-10
fix(mcp): warn about dbt selector grammar in select param descriptions

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix`

The MCP `select` string feeds dbt's node-selection grammar, where a comma means intersection. Agents comma-join distinct model names (e.g. `a,b`), so dbt intersects them → zero nodes → an empty diff indistinguishable from "no changes" → silent under-reporting (seen in production). Since `select` is passed to dbt verbatim, the fix is in the description contract: warn on every `select` param that a comma is intersection, a space is union, and bare names are exact.

DRC-3834

- Grammar verified against dbt docs (Core 1.11): space=union, comma=intersection, bare name=exact, wildcard needs `*`.

NONE

### DIFF
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -53,6 +53,15 @@ SINGLE_ENV_WARNING = (
     "Run `dbt docs generate --target-path target-base` to enable diffing."
 )
 
+# Guards against agents comma-joining distinct model names in `select`, which dbt reads
+# as an intersection and silently resolves to an empty diff (looks like "no changes").
+SELECTOR_SYNTAX_NOTE = (
+    "Combine selectors with a SPACE for union (OR) or a COMMA for intersection (AND); "
+    "this applies to all selector arguments. "
+    "When targeting by model name: comma-joining distinct names matches nothing "
+    "(no node is all of them at once), and bare names match exactly (use 'name*' for prefix matching)."
+)
+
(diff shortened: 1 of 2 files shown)


iamcxa (approved):
Reviewed together with the agent-side sibling DataRecce/recce-cloud-infra#1522 (same DRC-3834 fix; the MCP-description ⇄ prompt contract). Grammar guidance is consistent across both.

**Verified correct ✅**
- `SELECTOR_SYNTAX_NOTE` is appended to all 5 tool-schema `select` params (lines 730 / 765 / 948 / 1023 / 1380) — complete coverage, not a partial rollout.
- The new test enforces it well: the per-tool `"intersection (AND)" in select["description"]` assertion means *every* tool exposing a `select` param must carry the note (not merely ≥5), so a future tool that adds `select` without it will fail the test. Good regression guard.
- The note's grammar matches the agent-side prompt in #1522 (space = union, comma = intersection, bare = exact, `*` for prefix).

**Minor (NIT) 💡**
- test_mcp_server.py L3163: the diff also collapses an unrelated `with patch.object(...)` from multi-line onto one line (looks like a `ruff format` sweep caught in the same commit). Harmless — just unrelated to the selector fix.

Approving — clean, well-tested description fix. The NIT is optional.

PR #1464, opened 2026-07-14
fix(cli): support duckdb_external_access opt-in in recce mcp-server

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

fix

`recce mcp-server` always ran DuckDB with `enable_external_access=false` and offered no way to turn it on, unlike `recce server`. On a local dbt/duckdb project whose sources use `external_location: read_csv(...)` and whose staging models are views, any `row_count_diff` / `profile_diff` touching a VIEW model failed with `Permission Error: Scanning CSV files is disabled through configuration`. `recce/exceptions.py` tells users to set `duckdb_external_access: true` in recce.yml, but mcp-server ignored that key, so the hint was misleading.

This makes mcp-server resolve the setting identically to `server`: the `--duckdb-external-access` flag, OR the `duckdb_external_access` key in recce.yml. Default stays blocked (opt-in only).

Fixes DRC-3890

- Default stays `False` (blocked) by design: `recce mcp-server --sse` is HTTP-exposed, so this is a security boundary. Production (recce-cloud-infra's `recce_instance_launcher`) runs it with no opt-in, so prod behavior is unchanged; enabling it there is a separate infra decision.
- The `--duckdb-external-access` option block and merge line are duplicated verbatim from `server` for exact parity; factoring into a shared list would also touch `server`, out of scope here.

Yes. `recce mcp-server` now accepts `--duckdb-external-access` and honors `duckdb_external_access: true` in recce.yml, matching `recce server`. Default is unchanged (external access blocked).


kentwelcome (approved):
LGTM

PR #1466, opened 2026-07-16
DRC-3025: stamp catalog float/decimal types so decimals compare exactly

danyelf:
## What & why

Fixes [DRC-3025](https://linear.app/recce/issue/DRC-3025) — numeric values lost their type on the wire, so the frontend could not compare them as numbers and fell back to exact string-inequality. Unchanged tables reported phantom changes.

The issue braids two defects together:

1. **Systemic** — a numeric column arrives untyped, so nothing tells the UI whether it is exact (`DECIMAL`) or approximate (`DOUBLE`).
2. **Narrow** — computed profile aggregates (`avg`, `median`, the proportions) genuinely carry IEEE-754 noise, so they are legitimately close-but-unequal.

**Approach:** the backend stamps each column's comparison type onto the diff payload — from `catalog.json` for model-backed dbt results, from the dtype for pandas results, and by construction for computed profile aggregates. The frontend then routes every cell-compare site through a single `isCellChanged(base, current, colType)` that dispatches on that type: `DOUBLE`/`FLOAT` tolerate a magnitude-relative epsilon, while `DECIMAL`/`NUMERIC`/`INT` compare exactly — as numbers, not as opaque strings, and with no `float64` anywhere in that decision.

Per the issue's explicit "Do NOT": nothing blanket-coerces `Decimal → float`. Exact comparison is done on canonical decimal digit strings, so `12345678901234567.89` and `…67.90` stay distinguishable even though they are the same `double`.

## Changes

**Backend**
- `recce/adapter/dbt_adapter/__init__.py` — `catalog_column_types()` resolves each column's true DB type from the already-loaded catalog (no new I/O, works offline, all adapters).
- `recce/tasks/dataframe.py` — new `FLOAT` comparison type; `from_db_type()` maps warehouse type strings to comparison semantics; `stamp_column_types()`; `from_pandas` labels `float64` as `FLOAT` (it *is* an IEEE-754 double — this is the SQLMesh path, which never stamps from a catalog).
- `recce/tasks/query.py`, `valuediff.py` — stamp catalog types on model-backed query-diff and value-diff-detail results. Ad-hoc SQL and catalog-absent cases keep the exact `from_agate` type.
- `recce/tasks/profile.py` — stamp the four computed aggregates `FLOAT`.

**Frontend**
- `js/packages/ui/src/utils/dataGrid/gridUtils.ts` — `isCellChanged` as the single comparison entry point, with `floatComparator` (epsilon, deep — also covers floats nested in JSON/ARRAY/STRUCT cells) and `exactNumericComparator` (canonical decimal digits).
- `rowBuilders.ts`, `toDiffColumn.tsx`, `inlineRenderCell.tsx` — every compare site routed through it, side-by-side and inline alike.
- `DataFrameColumnGroupHeader.tsx`, `DataFrameColumnHeader.tsx` — treat `float` like `number` for the render-mode menu.
- `api/types/base.ts` — `"float"` added to `ColumnType`.

**Tests** — `tests/tasks/test_drc3025_serialization.py` (backend), plus `gridUtils`, `rowBuilders` and `inlineRenderCell` unit tests.

## Reviewer notes

Two design points that look like bugs but are deliberate, both recorded in the code:

- **Comparison type is read from the current catalog for both sides of a diff.** A comparison type belongs to the column *pair*, not to one side — `isCellChanged` takes one type and could not reconcile two. The model-backed path is a preview-change diff whose two queries both already run against the current environment.
- **A number on one side and a string on the other is never bridged.** That pairing means the column's data type changed between base and current, which the grid reports as a modification on purpose (see the integer→text and text→number cases in `toDataDiffGrid.test.ts`).

The epsilon is `1e-9` *relative* with a `1e-12` absolute floor near zero. Float noise runs ~1e-15/1e-16 relative, and the exact forms these aggregates can take are far coarser than 1e-9 (`numeric(28,6)` resolves 1e-6), so no representable exact difference is swallowed.

**Supersedes #1457 and #1458** — the abandoned float-precision approach on `explore/drc-3025-float-precision`. Those are dead ends; this branch is the live fix.

**Not covered here:** `toRenderedValue` still display-collapses high-precision decimals to 2 dp via `parseFloat` (pre-existing for `number`; changing default rendering is a product decision). SQLMesh gets `FLOAT` from the dtype but no `DECIMAL` awareness, though `SqlmeshAdapter.get_lineage` exposes `columns_to_types` if we want that later.

## Verification

- Python: 1578 passed, 5 skipped. `black`, `isort`, `flake8` clean.
- Frontend: 187 files / 3973 tests passed. `biome` and `tsc --noEmit` clean. `pnpm run build` succeeds.
- Not done: no end-to-end pass against a live `recce server` — `integration_tests/dbt/target/manifest.json` is absent in a fresh checkout, so that needs a dbt run first.

## Does this PR introduce a user-facing change?

```
Data diffs no longer report phantom changes on unchanged numeric columns. Float-typed
columns (DOUBLE/FLOAT/REAL) and computed profile aggregates are compared with a
tolerance for floating-point noise, while exact decimals are compared exactly — a
one-cent change in a DECIMAL column is still reported as a change.
```


gcko:
# Code Review: PR #1466

**Reviewed at** `85e8d9cb` · **Verdict at that SHA** NO-GO · **Fixed at** `996d414b`

Adversarial review run on the inherited state of this branch, recorded here because the findings are the reason for the four commits that follow it. Every finding below was found by reading, not by a failing check: at `85e8d9cb` the suite was fully green (Python 1577 passed, frontend 3958 passed, lint and type-check clean) and stayed green with all of these present.

## Blockers

**1. `js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx:124` — the fix never reached the bug it was written for.**
The "did this cell change?" decision was a raw `===` on the two values, bypassing `isCellChanged`, so it never saw the column's comparison type:

```js
if (row[baseKey] === row[currentKey]) {
```

Inline is the **default** display mode for the three affected views — `ProfileResultView.tsx:192`, `ValueDiffDetailResultView.tsx:106`, `QueryDiffResultView.tsx:122`, all `?? "inline"`. So a profile `avg` of `"0.3"` vs `"0.30000000000000004"` still rendered as a `base → current` change: the exact screenshot on DRC-3025. The type stamping reached the side-by-side cell classes and the row status, and nothing else.

It also disagreed with itself — `determineRowStatus` already routed through `isCellChanged`, so the row was reported unmodified while its cells rendered a diff. `columnType` was already in scope at line 89 and read at line 141 for delta mode.
Pass A/C. → fixed in `59580936`.

## Issues

**2. `recce/tasks/dataframe.py:195` — `from_pandas` labelled genuine floats exact.**
`float64 → NUMBER`. A pandas `float64` *is* an IEEE-754 double, making this the one path where the value is really a float rather than a decimal string — and it was the path marked exact. That path is SQLMesh (`query.py:132`, `381-382`, `438`; `SqlmeshAdapter.support_tasks` registers `QUERY_DIFF`), which also never stamps from a catalog, so this label was its only signal. Every SQLMesh float column kept comparing by string-inequality.
Pass C. → fixed in `0d97e064`.

**3. The stated approach was not the implemented approach.**
The PR body claimed the frontend "routes all cell-compare sites through a single `isCellChanged(…, colType)`". Finding 1 was one uncentralised site; `getCellClass` is a second. A claim like that is worth verifying rather than trusting — it was false in exactly the way that mattered.
Pass F. → body rewritten; `isCellChanged` now documents the requirement in code.

**4. Decimals were still compared as opaque strings — half of the issue's fix direction was missing.**
`isCellChanged` sent everything non-float to `_.isEqual` on the raw wire string. That *is* the "the UI cannot compare it as a number and falls back to exact string-inequality" behaviour DRC-3025 set out to remove; only the float half had been addressed. Two spellings of one value (`"19.99"` vs `"19.990"`, which a scale change produces) still read as a change.
Pass A. → implemented in `d0ccabd7`, comparing canonical decimal digit strings with no `Number` on the path, so a 1-cent difference in a `DECIMAL(38,2)` survives.

**5. `recce/tasks/valuediff.py:56` — dead, untested parameter.**
`_stamp_catalog_types(..., base: bool = False)` had no call site passing `base=True`, so the base-catalog branch was unreachable. Reading the current catalog for both sides is in fact correct, but nothing recorded why, and the parameter advertised a capability that did not exist.
Pass F. → fixed in `996d414b`.

## Notes

**6. `gridUtils.ts:228` — `parseFloat` parses a numeric prefix.** `parseFloat("0.3abc") === 0.3`, so two different strings compared equal inside a float cell. → fixed in `d0ccabd7` (whole-string numeric parse, screening the blank input `Number` maps to 0).

**7. `recce/tasks/profile.py:11` — the comment justified the stamp with facts that are not true in general.** "avg/median are DOUBLE; the proportions are ratios": `avg` over a `DECIMAL` comes back exact on most adapters, and both proportions divide by `cast(count(*) as {{ dbt.type_numeric() }})`, which is exact `NUMERIC`. The stamp is still right, for a different reason — these *may* be approximate, so the epsilon has to be safe for the exact case too, and it is. → reworded in `996d414b`.

**8. `gridUtils.ts:434` — `getCellClass` gained `colType?` but has no in-repo caller.** It is public API via `utils/index.ts:91`, so external consumers keep exact compare until they pass the new argument. Left as-is; both branches of the parameter are now pinned by tests rather than a call site being invented for it.

**9. `gridUtils.ts:417` — `toRenderedValue` display-collapses high-precision decimals to 2 dp via `parseFloat`.** Pre-existing for `"number"`, not a regression, but it means high-precision `DECIMAL`s are already collapsed *for display* — worth not mistaking for solved.

**10. Test gaps.** Nothing covered the inline render path (the default), `from_pandas`'s type mapping, or `getCellClass`'s new parameter. → all three added.

## One finding the tests caught on me

The first attempt at finding 4 also bridged number↔string comparison, on the reasoning that agate emits strings while pandas emits JSON numbers. `toDataDiffGrid.test.ts` rejected it, correctly: this codebase *deliberately* reports an integer→text type change with the same value as modified. Both comparators are now string-to-string only, with that constraint written into the code so it does not get re-broken.

## Verification after the fixes

Python 1578 passed / 5 skipped (+1). Frontend 187 files, 3973 passed (+15). `biome`, `tsc --noEmit`, `black`, `isort`, `flake8` clean; `pnpm run build` succeeds. The blocker's regression test was confirmed to fail against the old `===` before the fix was restored.

Not verified: no end-to-end pass against a live `recce server` — `integration_tests/dbt/target/manifest.json` is absent in a fresh checkout. The label chain was instead traced by reading it end to end (backend `float` → `buildMergedColumnMap` → `columnBuilders` → `toDiffColumn` context → `inlineRenderCell`) with unit tests at each hop.

---

🤖 Review and fixes generated with [Claude Code](https://claude.com/claude-code)

gcko (approved):
LGTM

PR #1467, opened 2026-07-16
DRC-3809: resolve PIVOT output columns to real source columns (CLL)

danyelf:
_Note: this is Claude being very excited about the PIVOT operation in particular. In general, it's important for both our system and our demos to show that CLL properly transitions in things like a parent model with a column change may become a child with a whole-model change, or vice versa._

## Background (for a reader with no context)

Recce shows **column-level lineage (CLL)** — for each column of a dbt model, which upstream columns feed it. It computes this by parsing each model's compiled SQL with [sqlglot](https://github.com/tobymao/sqlglot) and walking the query's scopes to connect every output column back to its real source column(s).

This PR fixes CLL for SQL **`PIVOT`** queries (common in Snowflake).

## The problem

When a query uses `PIVOT`, sqlglot rewrites every projected column to reference a **synthetic pivot table alias** (e.g. `P`, or an auto-generated `_q_0` / `_0` depending on sqlglot version). But sqlglot never registers that alias as a real source in the query's scope — the pivot hangs off the *base* table node instead. Recce's resolver (`recce/util/cll.py`) had no handling for this, so it fell through to a "phantom" fallback that emitted a dependency on a table (`P` / `_q_0`) that doesn't exist in the DAG.

Downstream, the dbt adapter couldn't map that phantom table to any model, so it **dropped the edges entirely**. The result: a pivoted model lost **all** of its lineage — not just the pivoted columns' column-level parents, but the model-level provenance too.

### Before

For the query `select id, jan, feb from t1 pivot (sum(rev) for month in ('jan','feb')) as p` (schema `t1(id, month, rev)`), CLL produced:

```
model-level lineage (m2c):  []                    ← all source provenance to t1 LOST
column 'ID':   passthrough ← [(P, ID)]            ← phantom: P is not a real node
column 'JAN':  passthrough ← [(P, JAN)]           ← phantom
column 'FEB':  passthrough ← [(P, FEB)]           ← phantom
```

In the running app the pivoted model showed no real upstream lineage — the columns pointed at a pivot alias that isn't in the graph, so the edges vanished.

### After

```
model-level lineage (m2c):  [(T1, REV), (T1, MONTH)]     ← provenance to t1 restored
column 'JAN':  derived    ← [(T1, REV), (T1, MONTH)]     ← real pre-pivot sources
column 'FEB':  derived    ← [(T1, REV), (T1, MONTH)]
column 'ID':   passthrough ← [(T1, ID)]                  ← real pass-through source
```

The generated pivot value columns (`JAN`, `FEB`) resolve to the columns the pivot actually consumes — the **measure** (`t1.rev`) and the **dimension** (`t1.month`) — and unchanged pass-through columns (`id`) resolve straight to their base column. No phantom node anywhere.

### Seen in the running app

Demonstrated on a real jaffle-shop dbt-duckdb project with a `payments_by_method` model = `stg_payments` pivoted (`sum(amount) for payment_method in (credit_card, coupon, bank_transfer, gift_card)`). _(There is no "before" screenshot because the pre-fix state produced no resolvable lineage at all — the phantom node dropped the edges; the "before" is the empty/broken lineage described above.)_

A pivot **value column** (`credit_card`) correctly resolves as **derived** from `stg_payments.payment_method` + `stg_payments.amount`:

Column lineage for payments_by_method.credit_card — derived from payment_method + amount

A **pass-through column** (`order_id`) correctly resolves as **passthrough** from `stg_payments.order_id`:

Column lineage for payments_by_method.order_id — passthrough from order_id

And the model-level lineage graph is intact — `raw_payments → stg_payments → payments_by_method`, with **no phantom pivot-alias node**:

Lineage graph with the pivoted model resolving through stg_payments, no phantom node

## How

A small, surgical resolver in `_cll_select_scope` / `source_column_dependency` (`recce/util/cll.py`). Before falling through to the phantom path, it builds a per-scope **pivot map** from the base table's pivot node and resolves:
- **value output columns** → `derived` on the measure + dimension source columns,
- **everything else** → `passthrough` on the base table,
- and extends model-level lineage (`m2c`) with the consumed source columns.

Resolution defers to the existing base-source logic. The full fix applies when the pivoted relation is a raw table; for a CTE/subquery base, sqlglot registers the pivot alias as a real source, so column deps resolve to the CTE name and are dropped downstream — only the model-level (m2c) extension takes effect there (see Known boundary below). UNPIVOT nodes, which share the same `args["pivots"]` slot, are explicitly skipped and stay on the pre-existing phantom path. Crucially it keys off the **pivot node's structure, never the alias string** — so it's immune to the `_0` vs `_q_0` synthetic-alias drift between sqlglot versions (verified: ideation ran sqlglot 27.29.0 → `_q_0`; implementation ran 30.11.0 → `_0`; both pass).

## Verification

- New test `tests/util/test_cll.py::ColumnLevelLineageTest::test_pivot` — explicit-alias **and** `select *` (no alias) cases, asserting the exact edges above with **no** phantom node. Genuinely test-first: reverting the resolver reddens it (m2c empty, `JAN`→`P` phantom).
- Full CLL suite **49/49**; full backend suite **1548 passed** (once the frontend is built). The parked DRC-3806 uncorrelated-subquery case is undisturbed.
- **Running-app demo** on a real jaffle-shop dbt-duckdb project: a `payments_by_method` PIVOT model renders in the live Recce app with its value columns resolving to `stg_payments.payment_method` + `stg_payments.amount` and `order_id` passing through — no phantom node. (Screenshots in the workflow evidence folder `docs/spacedock/recce-lineage-bugs/_evidence/drc-3809/`.)

## Known boundary (not a regression — possible follow-up)

The fix covers a `PIVOT` hanging directly off a base table. A `PIVOT` wrapped in a subquery/CTE (`... from (select ...) pivot ...`) still falls to the old phantom path. Out of scope for this AC; worth a follow-up if subquery-pivots matter.

Part of the DRC-2891 complex-SQL-pattern audit. Sibling UNNEST/LATERAL gap tracked separately (DRC-3807/3808). Opening as **draft** — validated in the workflow; pending final review.


gcko:
# Code Review: PR #1467

**SHA** `48b521ea` · **Verdict** NO-GO

## Blockers

1. `recce/util/cll.py:374` — The pivot map treats UNPIVOT nodes as PIVOT, fabricating passthrough dependencies on base columns that do not exist.
   Evidence: sqlglot stores UNPIVOT in the same `args["pivots"]` slot as an `exp.Pivot` with `unpivot=True`, and the loop never checks it. For `select id, month, rev from t1 unpivot (rev for month in (jan, feb)) as u` with schema `t1(id, jan, feb)`, this branch produces `MONTH: passthrough ← (T1, MONTH)` and `REV: passthrough ← (T1, REV)` — columns that do not exist on `t1`. On `main` the same query yields phantom-`U` deps that the dbt adapter drops (missing lineage); this PR turns missing lineage into *wrong* lineage.
   Pass A.

   Fix — skip unpivot nodes so they stay on the pre-existing phantom path, and add a regression test:
   ```python
   for pivot in base_source.args.get("pivots") or []:
       if pivot.args.get("unpivot"):
           # UNPIVOT inverts producer/consumer semantics; mapping it like a
           # PIVOT fabricates passthrough deps on nonexistent base columns.
           continue
       alias = pivot.alias
   ```

## Issues

1. `recce/util/cll.py:27` — This PR changes `cll()` output for unchanged inputs, but the opt-in SQLite cache has no algorithm-version invalidation, so `ENABLE_CLL_CACHE=1` users keep serving pre-fix phantom pivot lineage indefinitely.
   Evidence: the cache key is `make_node_key(node_id, content_key)` where `content_key` = adapter_type + checksum + parent_checksums + column_names (`recce/adapter/dbt_adapter/__init__.py:1101`) — no code version anywhere. `_CACHE_SCHEMA_VERSION` is written to `cache_meta` in `_init_db` but never read back (grep: zero readers), and eviction is access-based (`last_accessed`), so an actively used entry never expires.
   Pass F.

   Fix — make the version participate in the key so old entries miss naturally:
   ```python
   @staticmethod
   def make_node_key(node_id: str, content_key: str) -> str:
       h = hashlib.sha256()
       h.update(str(_CACHE_SCHEMA_VERSION).encode("utf-8"))
       h.update(b"\x00")
       h.update(node_id.encode("utf-8"))
       h.update(content_key.encode("utf-8"))
       return h.hexdigest()
   ```
   and bump `_CACHE_SCHEMA_VERSION` to 3 in this PR (and on any future semantic CLL change).

## Notes

1. PR body ("How" section) claims resolution "works whether the pivoted relation is a raw table or a CTE/subquery." For a CTE base, sqlglot registers the pivot alias as a real source (`P → exp.Table(C)`), so `_resolve_source_column` succeeds via the plain-Table branch and the new fallback never fires — column deps land on the CTE name `C` and are dropped downstream. Only the model-level (m2c) extension takes effect there. This matches the stated Known boundary, but the "How" claim overstates.

Verification: `tests/util/test_cll.py` 49/49 and full backend suite 1579 passed / 5 skipped on sqlglot 30.4.2; CLL suite also green on the pin floor sqlglot 27.29.0. flake8 / black / isort clean.

recce/util/cll.py
```diff
@@ -346,6 +358,72 @@ def source_column_dependency(ref_column: exp.Column) -> Optional[CllColumn]:
...
+    # through to the phantom-alias fallback below, losing both column- and
+    # model-level lineage. Build a map keyed by the pivot alias so those
+    # projections resolve back to the real pre-pivot source columns. Keyed off
+    # the pivot node's *structure*, never the alias string (version-dependent).
+    pivot_map: Dict[str, dict] = {}
+    for base_source in scope.sources.values():
+        if not isinstance(base_source, exp.Table):
+            continue
+        for pivot in base_source.args.get("pivots") or []:
+            alias = pivot.alias
```
gcko, line 391: BLOCKER: UNPIVOT nodes live in the same `args["pivots"]` slot (`exp.Pivot` with `unpivot=True`); mapping them like a PIVOT fabricates passthrough deps on base columns that don't exist (e.g. `MONTH ← T1.MONTH` for `t1(id, jan, feb)`). Skip them so they stay on the pre-existing phantom path.

```suggestion
        for pivot in base_source.args.get("pivots") or []:
            if pivot.args.get("unpivot"):
                # UNPIVOT inverts producer/consumer semantics; mapping it like
                # a PIVOT fabricates passthrough deps on nonexistent base
                # columns. Leave it on the pre-existing phantom path.
                continue
            alias = pivot.alias
```

gcko (changes requested):
Code Review Loop: Critical issues found (1 BLOCKER, 1 ISSUE). See review comment: https://github.com/DataRecce/recce/pull/1467#issuecomment-5162332570

gcko:
✅ **Code Review Loop — fixes pushed** (`29f3b7b5`)

- **Blocker (UNPIVOT fabricated lineage)** — fixed in `29f3b7b5`: UNPIVOT nodes (`exp.Pivot` with `unpivot=True`) are now skipped in the pivot map and stay on the safe phantom path; regression test `test_unpivot_not_treated_as_pivot` added. Verified: `month`/`rev` deps back to alias-`u` (dropped downstream), m2c empty — identical to main; PIVOT fix itself unaffected.
- **Issue (stale CLL cache after semantic change)** — fixed in `29f3b7b5`: `_CACHE_SCHEMA_VERSION` now participates in `make_node_key` and was bumped 2 → 3, so all pre-fix cached lineage misses naturally; test `test_schema_version_bump_is_cache_miss` added.
- **Note (PR body overstated CTE/subquery coverage)** — PR description corrected in place.

Verification at `29f3b7b5`: full backend suite 1581 passed / 5 skipped; CLL + cache suites also green on sqlglot pin floor 27.29.0; flake8/black/isort clean.

recce/util/cll.py
gcko: Fixed in 29f3b7b5. Confirmed the fabrication first: on 48b521ea, `select id, month, rev from t1 unpivot (rev for month in (jan, feb)) as u` with schema `t1(id, jan, feb)` produced `month: passthrough <- (t1, month)` and `rev: passthrough <- (t1, rev)` — columns that don't exist on `t1` — while main produced phantom `(u, *)` deps that get dropped downstream.

Applied the suggested guard at `recce/util/cll.py:374` (`if pivot.args.get("unpivot"): continue`). Post-fix output is byte-identical to main's safe phantom path: all three projections `passthrough <- (u, <col>)`, m2c empty. Regression test `test_unpivot_not_treated_as_pivot` asserts both halves — no `(t1, month)`/`(t1, rev)` deps anywhere, and the exact phantom-`u`/empty-m2c behavior main had.

Same commit also folds `_CACHE_SCHEMA_VERSION` into `CllCache.make_node_key` and bumps it to 3, so `ENABLE_CLL_CACHE=1` users miss on all pre-fix entries (per the review's Issue 1), with `test_schema_version_bump_is_cache_miss` covering the miss-on-bump property.

Verified: full backend suite 1581 passed / 5 skipped (sqlglot 30.x), CLL suites 99/99 on the pin floor sqlglot 27.29.0, flake8/black/isort clean.

gcko (approved):
Code Review Loop: blocking findings fixed and verified at 29f3b7b5.

PR #1468, opened 2026-07-16
DRC-3813: impact outranks additive in CLL graph-badge precedence

danyelf:
_Note: this is Claude being really enthusiastic about the issue MERELY being rank. From a design POV, the real issue is that a local change (e.g. a column-level additive) is both important to see -- but also should not hide the fact that it's impacted from a parent._

## Background (for a reader with no context)

In Recce's new column-level-lineage (CLL) experience, each node in the lineage graph shows a small **badge** summarizing how that model was affected by a change. The relevant two here:

- 🟢 **ADD** ("additive") — the model added a column; existing rows/columns are unchanged. Benign.
- 🟠 **COLUMN** ("column-impacted") — a column this model *depends on* changed upstream, so this model is impacted and worth reviewing.

Which badge to show is decided by `classifyGraphBadge` in `js/packages/ui/src/components/lineage/wholeModelTreatment.ts`.

## The problem

`classifyGraphBadge` checked the node's **own** change category *before* checking whether it was **impacted** by an upstream change:

```ts
if (changeCategory === "non_breaking")  return "additive";      // ← checked first
if (changeCategory === "partial_breaking") return "column-changed";
if (isImpacted) return "column-impacted";                        // ← too late
```

So a node that is genuinely impacted upstream, but whose own SQL only *added* something (category `non_breaking`), got the benign green **ADD** badge — hiding the more important fact that it's impacted. The backend had already correctly flagged the node as impacted; this was purely a **presentation** bug in the badge ordering. (Only `non_breaking` was affected — see the precedence table below.)

### Concrete repro

- **PARENT** model redefines a column: `c1` changes from `id` to `id || lastname`.
- **CHILD** model: `select id, name from parent where length(parent.c1) < 10` — it *filters* on `c1` but never selects it, and its own SQL is otherwise unchanged.

CHILD is impacted (the column it filters on changed), but because its own change is merely additive/non-breaking:

| | Badge before | Badge after |
|---|---|---|
| CHILD node | 🟢 **ADD** (green, "additive") — misleading | 🟠 **COLUMN** (amber, "column-impacted") — correct |

Here is the same CHILD node rendered with identical inputs — the fix reverted (left) vs applied (right):

Before/after: green ADD badge vs amber COLUMN badge on the same node

## The fix

Move the `isImpacted` check up by one line so impact outranks additive, while a genuine own-column change (`partial_breaking`) still outranks impact:

```ts
if (changeCategory === "partial_breaking") return "column-changed";
if (isImpacted) return "column-impacted";
if (changeCategory === "non_breaking")  return "additive";
```

Net precedence: **column-changed (own change) > column-impacted (impacted) > additive (own additive)**. This preserves the file's documented "changed-wins" doctrine — a node's own *change* still outranks being *impacted* — but a harmless additive change no longer masks the impact signal.

Precedence across all four change categories combined with `isImpacted: true` (only `non_breaking` was wrong):

| change_category | before | after | was it a bug? |
|---|---|---|---|
| `breaking` | (short-circuited earlier) | same | no |
| `non_breaking` | additive (green) | **column-impacted (amber)** | **yes — this bug** |
| `partial_breaking` | column-changed | column-changed | no — intentional |
| `unknown` | column-impacted | column-impacted | no |

## Verification

- Strict TDD: 3 combined-case tests (classifier light+dark in `wholeModelTreatment.test.ts` via the exported `pickGraphBadge`, plus a `<LineageNode>` render assertion in `LineageNode.test.tsx`) committed **red first**, green after the reorder. Reverting only the one-line reorder reddens exactly those 3 and nothing else.
- The four existing regression cases (additive-only, column-changed, column-impacted-only, whole-model short-circuit) stay green.
- Full frontend suite **4045 passed / 5 skipped**; Biome + `type:check` clean on all touched files.
- Reproducible **Storybook story** `LineageNode_ImpactedOutranksAdditive` renders the CHILD repro through the real `LineageNode`, driving the before/after image above.

Frontend-only; backend `node.impacted` was already correct and is untouched. Distinct from the DRC-2891/380x lineage-extraction work (that's a different layer). Opening as **draft** — validated in the workflow; pending final review.


gcko:
# Code Review: PR #1468

**SHA** `62481b8c` · **Verdict** NO-GO

## Blockers

1. `js/packages/ui/src/components/lineage/wholeModelTreatment.ts:170` — The `wire-enum-lint` CI gate fails on this HEAD: 4 added lines under `js/packages/ui/src/` carry the legacy `"non_breaking"` wire-enum literal without the `// wire-enum-ok` marker (DRC-3553 dual-vocabulary guard).
   Evidence: `bash scripts/check_wire_enum_literals.sh origin/main` → "Found 4 offending added line(s)": the moved precedence line (`if (inputs.changeCategory === "non_breaking") return "additive";` — a moved line counts as added), the quoted literals in the two new comments (`wholeModelTreatment.ts:164`, `wholeModelTreatment.test.ts:129`), and the new test props line `changeCategory: "non_breaking",` in `LineageNode.test.tsx:794`. The sibling test line in `wholeModelTreatment.test.ts` that carries `// wire-enum-ok` passes the same check.
   Pass F.

   Fix — append the marker to the two code lines; unquote the literal inside comments (the checker matches quoted forms only):
   ```ts
   if (inputs.changeCategory === "non_breaking") return "additive"; // wire-enum-ok
   ```
   ```ts
   changeCategory: "non_breaking", // wire-enum-ok
   ```
   and in the two comments write non_breaking without surrounding quotes.

Everything else verified clean: the reorder produces the documented precedence (partial_breaking > impacted > non_breaking; `breaking`/`unknown` rows unchanged — confirmed against the classifier source, not just the PR table); `LineageNode` is the sole `pickGraphBadge` consumer and `NodeView` tooltips route through the same classifier so both surfaces stay in sync; no duplicate precedence logic elsewhere; targeted suites 75/75, `tsc --noEmit` clean, Biome clean, story imports respect the `@datarecce/ui` package boundary; security pass clean (pure classifier, no user-input sinks). CI `test-and-build` and `lint` are green on this HEAD — only `wire-enum-lint` blocks.

js/packages/ui/src/components/lineage/wholeModelTreatment.ts
```diff
@@ -160,9 +160,14 @@ function classifyGraphBadge(inputs: TreatmentInputs): GraphBadgeKind | null {
...
   if (inputs.isWholeModelChanged || inputs.isWholeModelImpacted) return null;
-  if (inputs.changeCategory === "non_breaking") return "additive";
+  // Precedence: own column change (partial_breaking) > impact > own additive
+  // change (non_breaking). isImpacted must outrank "non_breaking" so a node
+  // that is impacted upstream but only additively changed itself surfaces the
+  // actionable column-impacted badge rather than the benign additive one
+  // (DRC-3813). partial_breaking still wins over impact per changed-wins (Q11).
   if (inputs.changeCategory === "partial_breaking") return "column-changed";
   if (inputs.isImpacted) return "column-impacted";
+  if (inputs.changeCategory === "non_breaking") return "additive";
```
gcko: BLOCKER: `wire-enum-lint` flags this moved line (a move counts as an added line) and the quoted literal in the comment above. Append the `// wire-enum-ok` marker and unquote non_breaking in the comment.

```suggestion
  // Precedence: own column change (partial_breaking) > impact > own additive
  // change (non_breaking). isImpacted must outrank non_breaking so a node
  // that is impacted upstream but only additively changed itself surfaces the
  // actionable column-impacted badge rather than the benign additive one
  // (DRC-3813). partial_breaking still wins over impact per changed-wins (Q11).
  if (inputs.changeCategory === "partial_breaking") return "column-changed";
  if (inputs.isImpacted) return "column-impacted";
  if (inputs.changeCategory === "non_breaking") return "additive"; // wire-enum-ok
```

gcko (changes requested):
Code Review Loop: Critical issue found (wire-enum-lint CI gate). See review comment: https://github.com/DataRecce/recce/pull/1468#issuecomment-5162577281

gcko:
✅ **Code Review Loop — fix pushed** (`456d08f0`)

- **Blocker (wire-enum-lint CI gate)** — fixed in `456d08f0`: `// wire-enum-ok` markers on the moved precedence line and the new test-props line; the two new comments now reference non_breaking unquoted. No logic changed. Verified locally: `scripts/check_wire_enum_literals.sh origin/main` → OK (was 4 offenders); targeted suites 75/75; Biome + `tsc --noEmit` clean; pre-push hook ran the related-file suite (1775 passed / 5 skipped).

Review state updated to approve below. Note: the PR is still marked **draft** — flipping it to ready is the author's call.

js/packages/ui/src/components/lineage/wholeModelTreatment.ts
gcko: Fixed in 456d08f0 — applied as suggested:

- `wholeModelTreatment.ts:170` — appended `// wire-enum-ok` to the moved `non_breaking` precedence line.
- `wholeModelTreatment.ts:164` — unquoted `non_breaking` in the comment (the checker matches quoted forms only).
- `wholeModelTreatment.test.ts:130` — unquoted `non_breaking` in the describe-block comment.
- `LineageNode.test.tsx:795` — appended `// wire-enum-ok` to the new `changeCategory: "non_breaking"` test props line, matching the existing convention at `wholeModelTreatment.test.ts:80,142`.

Verification on 456d08f0: `bash scripts/check_wire_enum_literals.sh origin/main` → "OK: no new legacy change-category wire-enum literals added under js/packages/ui/src/."; Biome clean on the 3 files; targeted suites 75/75; `tsc --noEmit` clean. Comment/marker-only changes — no logic touched.

gcko (approved):
Code Review Loop: blocking finding fixed and verified at 456d08f0.

PR #1480, opened 2026-07-28
test: replace tautological tests with behavioral contracts

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`test`, `refactor`, `fix`

Replaces tautological and implementation-mirroring tests with scoped behavioral
contracts derived from the repository documentation and recent production
changes.

The work is split by functional area:

- runtime/task cancellation and API behavior
- state persistence and loaders
- dbt smoke behavior
- frontend rendering and UI adapter contracts
- adapter and registry behavior
- column-level-lineage cache patching, impact rendering, and async ownership

Strengthening the tests kept surfacing real defects, so the PR also carries the
fixes for them.

**Column-level lineage ownership** (original scope): monotonic
request/layout/interaction ownership, structural-sharing-aware cache patch
detection, and unmount invalidation, so superseded CLL work cannot update the
cache or the UI.

**Defects found while closing review gaps** (round 2):

- `POST /runs/{run_id}/cancel` never cancelled anything. `submit_run` registered
  the task in `running_tasks` under a `UUID` key while every reader looks it up
  by `str`, so the lookup always missed, the handler swallowed the resulting
  `RecceException`, and the endpoint reported the no-op as `acknowledged`. Run
  status was never flipped and `task.cancel()` was never called. Every existing
  cancel test seeded the `str` key by hand, so none of them could see it.
- `recce summary` raised `AttributeError` on every SQLMesh project.
  `generate_summary_metadata` dereferenced `manifest_metadata.generated_at`
  unguarded, and the SQLMesh adapter returns `{}` for both metadata slots. It
  now renders `N/A`.
- The lineage panel could render completely empty on first load. Making the
  layout effect abortable (this branch) means `refreshLayout` is routinely the
  only layout producer on first load — e.g. when a run result is open on a model
  the current view does not show. If its node selection then failed,
  `refreshLayout` returned without laying out, `nodes` stayed at its initial
  identity, and the render guard drew an empty fragment: no canvas, no top bar,
  no retry. It now releases the layout generation and requests another run.
- `--impact-at-startup` could be silently dropped for a whole session. The
  one-shot arming sits after the node-selection await, so a superseded first
  layout discarded the decision and the effect never ran again. Superseded runs
  now request one more layout run while that decision is still pending.

Tests assert externally observable results and production entry points. Several
weak or redundant tests were removed rather than replaced solely to preserve
test count. Every guard claimed below was verified by mutation: the named test
fails when the guard is removed.

Current verification:

- frontend suite: 3,928 passed, 5 skipped (186 files)
- frontend lint: 662 files clean; TypeScript clean
- production frontend build: passed (5 static routes)
- Python suite: 1,552 passed, 5 skipped
- Python formatting, isort and Flake8: passed; pre-commit hooks passed
- mutation checks pass for: the change-analysis transitions at both call sites
  and the context readout, the layout-effect CLL disarm, both CLL error toasts,
  the auto-trigger rollback, the zoom floors, the `running_tasks` key
  normalization, the `generate_summary_metadata` guard, the `refreshLayout`
  layout-ownership release, the impact-at-startup retry, the
  `selectNode`/`selectParentNodes`/`selectChildNodes` supersede and unknown-node
  guards, the CLL history-push ordering, the `ChangeSummary` counter wiring and
  icons, and the `SchemaSummary` prop wiring

NONE

Please review the tests as behavioral contracts, not as a request to preserve
the prior assertion count. The most concurrency-sensitive code is in
`LineageViewOss`, `cllCachePatchLifecycle`, and their real component/React Query
harnesses.

Two things worth a second opinion:

1. `js/src/components/lineage/__tests__/LineageView.component.test.tsx` gained a
   `mockUseRealInitialNodes` opt-in. Production's empty-render guard is
   `nodes == initialNodes` (referential), so a stubbed node array made that path
   invisible to every test in the file — which is why the blank-panel bug above
   survived the first review round.
2. Two things I deliberately did **not** change, both pre-existing on `main`:
   a CLL `HttpError` aborts the layout update rather than laying out without CLL
   (the non-HTTP path already falls through and lays out), and a non-`HttpError`
   node-selection failure produces no toast at all. Both are product-behavior
   calls rather than regressions from this branch — say the word and I will fix
   them here. `running_tasks` also still grows unboundedly for the life of the
   server; evicting on completion interacts with the cancel/terminal-status race,
   so I left it as a follow-up.

```release-note
Fixed: run cancellation had no effect — POST /runs/{run_id}/cancel could never find the running task, so the run kept executing and its status was never set to cancelled.
Fixed: `recce summary` failed with an AttributeError on SQLMesh projects.
Fixed: the lineage view could render empty with no canvas or top bar when the first layout was superseded by a view change that then failed to select nodes.
Fixed: `--impact-at-startup` could be silently skipped for a session when a run result was open on a model outside the current view.
```


kentwelcome:
# Code Review: PR #1480

**SHA** `31295684` · **Verdict** NO-GO

Quality gates verified locally: Python 1,548 passed / 5 skipped (py3.12), frontend 3,899 passed / 5 skipped, Biome lint clean, `tsc` clean. The production changes (`LineageViewOss`, `cllCachePatchLifecycle`, `changeAnalysisState`, `patchLineageDiffFromCll`, `run_func.py`, `track.py`) produced no findings — mutation testing confirmed the new lifecycle guards (epoch supersession, pending-reuse, unmount invalidation, structural sharing) are each killed by a named test. Every issue below is a gap in the test contracts this PR exists to establish.

## Blockers

None.

## Issues

1. `js/packages/ui/src/components/lineage/__tests__/CllChangeAnalysisPropagation.test.ts:110-157` — The suite drives the pure helpers but re-creates `LineageViewOss`'s call sequence in test-local helpers, so the component's *consumption* of the transitions is unguarded, contradicting the file header's claim.
   Evidence: deleting the `nextChangeAnalysisMode` block (`LineageViewOss.tsx:796-803`) and hard-wiring the context's `isNodeShowingChangeAnalysis` (`LineageViewOss.tsx:1381-1387`) to `() => false` leaves all 3,899 frontend tests green; same for the mode-clear at `LineageViewOss.tsx:1010-1017`.
   Pass E.

2. `js/packages/ui/src/components/lineage/LineageViewOss.tsx:521-528` and `:924-931` — The new "no lineage graph ⇒ disarm the CLL lifecycle" branches have zero coverage.
   Evidence: deleting both `resolveCllForLayout({cllInput: undefined,…})` and `refreshCll({cllInput: undefined,…})` calls leaves the full suite green; `CllCachePatchLifecycle.test.ts:430-479` proves the lifecycle disarms when called, but nothing proves the component reaches it.
   Pass E.

3. `js/packages/ui/src/components/lineage/LineageViewOss.tsx:614-641` and `:986-1004` — The CLL failure path rewritten by this PR (auto-trigger rollback + both error toasts) is untested anywhere in the repo.
   Evidence: replacing `if (autoTriggered)` with `if (false)` and removing both toasts leaves the suite green; `cllCachePatchLifecycle.ts:64-68` promises "Rejections propagate so the caller can keep its own error handling" — the caller's half of that contract is unverified.
   Pass E.

4. `js/src/components/lineage/__tests__/GraphNode.test.tsx:795` — The `columnHeight` assertion tests the file's own mock, and the mocked value is wrong.
   Evidence: line 138 mocks `COLUMN_HEIGHT: 28`, which is what production reads under the vitest alias; the real constant is `24` (`js/packages/ui/src/contexts/lineage/utils.ts:17`). The assertion passes with any production value and breaks if the mock is corrected.
   Pass E / C.

5. `js/src/components/summary/__tests__/SchemaSummary.test.tsx:11-12` — Two real prop-wiring assertions were deleted with no replacement: `NodeTag`/`RowCountDiffTag` are now bare `<span />` mocks nothing asserts on.
   Evidence: the `resourceType`/`materialized` wiring and the `node.data.resourceType === "model"` guard on `RowCountDiffTag` (`SchemaSummary.tsx:44-50`) can be deleted and all three tests still pass; the pre-PR suite asserted both.
   Pass E.

6. `tests/test_summary.py:24-40` — The hand-built `SimpleNamespace` fixture commits to a shape production does not guarantee, and it replaced the only test coupling adapter `get_lineage()` output to `generate_summary_metadata()`.
   Evidence: `sqlmesh_adapter.py:65` returns `manifest_metadata = {}` and `dbt_adapter/__init__.py:791` can return `None`, while `recce/summary.py:425` dereferences `base_manifest.generated_at` unguarded — both real shapes raise `AttributeError` when passed directly. Adding `None`/`{}` cases to the parametrization restores the lost coupling.
   Pass E / C.

## Notes

1. `tests/apis/test_run_func.py:264` — `assert warning in caplog.text` depends on the `"uvicorn"` logger propagating to root; uvicorn's own `LOGGING_CONFIG` sets `propagate: false`, so any future test-process logging config kills the sole regression guard on this PR's `run_func.py` fix. `assertLogs("uvicorn")` (as this PR already uses at `tests/models/test_check.py:570`) is immune.
2. `js/src/components/lineage/__tests__/LineageView.component.test.tsx:2141` — "does not send stale CLL to layout after a newer disable completes" passes with `|| !resolution.isCurrent()` removed from `LineageViewOss.tsx:614`/`988`; only the layout-generation half of the guard is pinned (plausibly redundant defence, but then the test name overclaims).
3. `js/src/components/lineage/__tests__/LineageView.component.test.tsx:1585-2040` — `(useMutation as Mock).mockReturnValue(...)` implementations leak across tests; `vi.clearAllMocks()` at line 780 clears call records, not implementations, so later tests silently depend on declaration order.
4. Deleted `zoomConstants.test.ts` was mostly tautological, but one genuine cross-constant invariant went with it: `EXPLORE_MIN_ZOOM < LEGIBLE_MIN_ZOOM`; no remaining test covers the `minZoom` wiring at `LineageViewOss.tsx:1611`.
5. `tests/models/test_check.py:575` — `mock_get_cloud_client.assert_not_called()` is vacuous: `CheckDAO.clear()` calls `_get_cloud_client()` on neither branch. The `assertLogs` and identity assertions carry the test.
6. `js/src/components/lineage/__tests__/GraphNode.test.tsx:326` and `GraphColumnNode.test.tsx:214` — both titles claim the "< 30%" zoom threshold but `useStore` is a bare stub, so the `transform[2] > 0.3` selector never executes; the threshold could change unnoticed.

kentwelcome (changes requested):
Claude Code Review: NO-GO — 6 issues (test-contract gaps), 0 blockers. See review comment.

gcko:
@kentwelcome All 6 issues and all 6 notes addressed in 3ac6242c. Every guard below was verified by mutation — the named test fails when the guard is removed, and I re-ran your own mutations to confirm they now get caught.

## Issues

**1. `CllChangeAnalysisPropagation` re-creates the call sequence in test-local helpers.** Correct, and the file header overclaimed. Added a `change analysis mode` suite in `LineageView.component.test.tsx` that drives the real component and reads the treatment off the same context `GraphNode` reads it from. Your three mutations now fail:

- hard-wiring `isNodeShowingChangeAnalysis` to `() => false` → 3 failures (global radius, node-scoped radius, widening on a column click)
- deleting the `nextChangeAnalysisMode` block at `:796-803` → 1 failure. It needed a *synchronous* assertion: `refreshLayout` clears the mode too, one microtask later, so anything that awaits passes either way. The contract is "turning CLL off clears the impact treatment before the layout round trip", i.e. no frame of stale impact styling.
- deleting the mode-clear at `:1010-1017` → 1 failure (a view-option change that drops CLL).

The header now says what the file actually covers and points at the component suite.

**2. No-lineage-graph disarm branches.** Fixed for `resolveCllForLayout`: "drops the armed patch when the lineage graph goes away, so its return re-fetches" fails when the disarm is deleted (the pending result gets replayed instead of re-fetched). Its companion pins the reuse path, so the re-fetch assertion isn't vacuous.

On `refreshLayout`'s half at `:924-931` I owe you a caveat: it is now *executed* by a test ("does not fetch CLL or lay out when a run wants a view change before lineage loads" — I confirmed the line runs), but its epoch bump is not independently observable. The layout effect owns every `lineageGraph` transition and disarms there first, so deleting the `refreshCll` call leaves the full suite green and I could not construct a state where it matters. I left it in with a comment saying it is defence in depth rather than pretending a test pins it. Happy to delete it instead if you'd rather not carry unobservable code.

**3. CLL failure path untested.** Fixed — a `CLL failure handling` suite. Removing either toast fails it, `if (autoTriggered)` → `if (false)` fails the rollback test, and there's a case pinning that a non-HTTP failure still lays out (with `cll: undefined`) and does not toast.

**4. `GraphNode.test.tsx:795` tests its own mock, and the mock is wrong.** Fixed, though not by correcting 28 → 24 — that would still pass for any production value. The mock is now `1234`, an obviously-unreal fixture, so the assertion proves `GraphNode` forwards the shared constant instead of hardcoding a number. The real constant's contract (the `COLUMN_HEIGHT` doc comment's "must match `COLUMN_NODE_HEIGHT`", unenforced anywhere) now has its own test in `packages/ui/src/contexts/lineage/__tests__/columnHeight.test.ts`.

**5. `SchemaSummary` prop wiring deleted.** Fixed. The mocks record props now: one test asserts `NodeTag` gets the node's own `resourceType`/`materialized`, another that `RowCountDiffTag` renders for the model and not for a seed. Deleting the wiring or the `resourceType === "model"` guard fails one test each.

**6. `test_summary.py` `SimpleNamespace` fixture commits to a shape production doesn't guarantee.** Correct, and it was hiding a live bug. Restored the adapter coupling (`test_generate_summary_metadata_accepts_adapter_lineage` runs real `DbtAdapter.get_lineage()` output through it) and added the missing-`generated_at` cases.

One correction to the detail: the dbt `None` path you cited at `dbt_adapter/__init__.py:791` is not reachable through `get_lineage`, because `manifest.to_dict()` on line 794 raises first — that's why the old test called `get_lineage()` for both sides. The reachable one is **SQLMesh**: `sqlmesh_adapter.py:65` returns `{}` for both metadata slots, `get_lineage_diff` feeds that straight into `generate_summary_metadata`, and `{}.generated_at` raised `AttributeError` — so **`recce summary` was broken on every SQLMesh project**. `summary.py` now renders `N/A`; reverting the guard fails the two new parametrized cases.

## Notes

1. **`caplog` vs uvicorn's `propagate: false`.** Agreed and switched to `assertLogs("uvicorn")`, as you suggested. Reverting the `run_func.py` fix (`task.params.__class__` → `type(task.params)`) still fails the `unknown-type` case, so the guard survived the change.
2. **`isCurrent()` half of the guard.** You're right that it's redundant defence: every entry point bumps the layout generation and the lifecycle epoch together, so I could not separate them either. Rather than rename the test I added a comment naming which half it pins and pointing at the `CllCachePatchLifecycle` tests that pin the epoch contract directly.
3. **`useMutation` implementation leakage.** Fixed in the root `beforeEach` (`clearAllMocks` clears calls, not implementations). This mattered more than it looked — see the extra findings below.
4. **Lost `EXPLORE_MIN_ZOOM < LEGIBLE_MIN_ZOOM` invariant.** Replaced with a behavioral version instead of restoring the constant comparison: one test reads `minZoom` off the rendered canvas and `minZoom` off the `fitView` call and asserts the canvas floor is strictly lower. That pins the wiring at `:1611` *and* the cross-constant relation, with no hardcoded numbers.
5. **Vacuous `assert_not_called`.** Removed. The patch stays as a network guard with a comment saying it is a guard and not an assertion, since `clear()` reaches `_get_cloud_client` on neither branch.
6. **"< 30%" titles with a bare `useStore` stub.** Fixed in both files: the mock now applies the real selector to a fake `transform`, and each has a pair of cases at `0.29` / `0.31`. The threshold is production's now.

---

## Extra findings from an adversarial pass

Reviewing your issues turned up things worth flagging separately. Three are production defects:

**`POST /runs/{run_id}/cancel` never cancelled anything.** `submit_run` registered the task as `running_tasks[run.run_id]` — a `UUID` — while `cancel_run_handler` calls `_mark_run_cancelled(str(run_id))` and the lookup uses that `str`. The lookup always missed, the handler swallowed the resulting `RecceException`, and logged the no-op as `outcome="acknowledged"`: run status never flipped, `task.cancel()` never called. Every cancel test in the repo (including the ones this PR adds) seeds `running_tasks[str(run.run_id)]` by hand, so none of them could see it. Normalized through `_task_key`, and the new `test_submitted_run_is_cancellable_through_the_endpoint_key` is the only test that lets `submit_run` write the key itself — reverting the fix fails it.

**The lineage panel could render completely empty on first load.** This one comes from this branch. Making the layout effect abortable means `refreshLayout` is routinely the *only* layout producer on first load — the exact interleaving the PR's own "focuses a hidden run model after the all-models layout resolves" test pins. `refreshLayout` claims the layout generation as its first statement, so when its node selection then fails it returned with no layout, `nodes` stayed referentially `initialNodes`, and the guard at `:1538` drew `<></>`: no canvas, no top bar, nothing to retry from. On `main` the same failure was a toast over a rendered graph. `refreshLayout` now releases the generation and requests another layout run on that path.

Worth knowing why this survived round one: the component harness stubs `useNodesState` with a non-empty array, so `nodes == initialNodes` is never true in tests. I added a `mockUseRealInitialNodes` opt-in so at least one test runs with production's real initial identity.

**`--impact-at-startup` could be silently dropped for a whole session.** The one-shot arming sits after the node-selection await, so a superseded first layout discarded the decision and the effect never ran again (deps are `[lineageGraph, impact_at_startup]`). Superseded runs now request one more layout run while that decision is still pending. Guarded on the flag, so it is inert by default.

Test-contract gaps in the same spirit as yours:

- `selectNode` / `selectParentNodes` / `selectChildNodes` had **zero** coverage for their new `supersedeCllInteraction()` calls and unknown-node bails. Now driven through the real handlers; removing any of the five guards fails a test. Isolating the supersede calls needed the already-selecting path — otherwise `handleViewOptionsChanged` supersedes on the way past and the mutation survives.
- Both "excludes the stale show from history" assertions were vacuous: after `previous-cll` the restore is async, so the trailing `expect` ran before anything could change. Flushed and made positive; moving `cllHistory.push` above the generation guard now fails both.
- `ChangeSummary`: `removes` and `col_changed` held the same value in all four fixtures, so swapping those two `value=` props passed the whole suite. Separated them, and `expectCount` now asserts one icon per counter — which restores what the deleted 6-icon assertion covered (`summary/utils.ts::getIconForChangeStatus`'s `col_*` branches had no coverage at all).
- `cllCachePatchLifecycle`: `satisfies` on an `it.each` array constrains element types, not coverage, so a seventh `CllInput` field would have compiled while `normalizeCllApiInput` dropped it and `isSameCllApiInput` ignored it — one false cache reuse. The field list is now derived from `satisfies Record<keyof CllInput, true>` and the test matrix from an exhaustive `Record`, so a new field fails to compile until all three sites cover it.
- `smoke_test.sh`: `grep -q customers` over the whole summary was satisfied by the mermaid Lineage Graph block, which names `customers` from the manifests alone — the entire check-summary rendering could disappear and CI stayed green, and it is the only exercise of `generate_check_summary`/`generate_check_content` anywhere. Now scoped per section, with the check counts and mismatched nodes pinned. Also: `jq '.result | keys | first'` sorts alphabetically so it could not detect selector over-selection (now the whole key set), and the review-mode server check was satisfied by the previous instance's still-open socket and never asserted `review_mode` (now `wait`s for the port and asserts the field).

Three things I deliberately left, all pre-existing on `main` — say the word and I'll do them here:

- A CLL `HttpError` aborts the layout update instead of laying out without CLL. The non-HTTP path already falls through and lays out, so the asymmetry looks unintentional, but fixing it changes what the user sees on a CLL failure.
- A non-`HttpError` node-selection failure produces no toast (the `return` sits outside the `instanceof HttpError` block).
- `running_tasks` is never evicted, so it grows for the life of the server. Evicting on completion interacts with the cancel/terminal-status race, so it wants its own change.

Verification: frontend 3,928 passed / 5 skipped (186 files), Biome and `tsc` clean, production build passed; Python 1,552 passed / 5 skipped, Black/isort/Flake8 and pre-commit clean. PR body and `release-note` updated — the four user-visible fixes above should not have shipped as `NONE`.

kentwelcome:
@gcko Verified locally at `3ac6242c`:

- Frontend: 3,928 passed / 5 skipped across 186 files; Biome and `tsc` clean. CI green across the Python 3.11–3.13 and dbt matrices, including the hardened `smoke_test.sh`.
- Re-ran three of the mutations myself: reverting the `_task_key` writer fails exactly `test_submitted_run_is_cancellable_through_the_endpoint_key` and nothing else; reverting the `_format_generated_at` guard fails 4 tests including the adapter-coupling one; hard-wiring `isNodeShowingChangeAnalysis` to `() => false` fails exactly the 3 change-analysis cases. All match your claims.
- My local Python run showed unrelated `MicrobatchConcurrency` errors — reproduced identically on `main`, so it's a local dbt package mismatch, not this branch.

Decisions on the open items:

1. **Keep the `refreshLayout` no-graph disarm** (`:924-931`) as documented defence-in-depth. The comment saying no test pins it is the right trade — deleting working defence to satisfy coverage would be backwards.
2. **Leave the three pre-existing `main` issues out of this PR.** File them as follow-ups: `running_tasks` eviction wants its own change given the cancel/terminal-status race you noted; the two toast/layout asymmetries on CLL and node-selection failures can go together in one small PR since they both change user-visible failure behavior.

Approving.

kentwelcome (approved):
Verified locally at 3ac6242c — full frontend suite, CI matrices, and mutation spot-checks all confirm the review-response commit. Details in https://github.com/DataRecce/recce/pull/1480#issuecomment-5105035324

PR #1481, opened 2026-07-29
fix(lineage): restore per-environment schema in merged lineage (DRC-3975)

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix`

The Environment Information panel rendered the same schema string in both the Base and Current columns, for every project, regardless of the real environments. A customer read this as "Recce is comparing my PR against itself".

Two changes on 2026-04-20 combined to cause it:

1. `build_merged_lineage` (DRC-3258) merges base + current into a single node with `source = current_node if current_node is not None else base_node`, so the base node's schema never reached the wire format.
2. `extractSchemas` (DRC-3260) then filled **both** the base and current sets from that one `node.data.schema`. Before DRC-3260 it read `node.data.data.base?.schema` and `node.data.data.current?.schema` separately.

First released in recce v1.45.0, so it has been live since late April.

**Backend.** `MergedNode` gains `base_schema`, populated only when the node exists in both environments *and* the base schema actually differs from `schema`. Absent means "same as `schema`", not "unknown" — that keeps the payload cost at zero for the common case where both environments share a schema, which matters because the reporting customer ships 4,616 nodes per lineage response.

Only the base side needs rescuing: `source` already prefers `current_node`, so `schema` *is* the current schema wherever the node exists in current. A second `current_schema` field would be dead weight.

**Frontend.** `base_schema` on `MergedNodeData`, `baseSchema` on `LineageGraphNode.data`, resolved once as `merged.base_schema ?? merged.schema` in `buildLineageGraph` so consumers don't each re-derive the fallback. `extractSchemas` reads `baseSchema` for the base set and `schema` for the current set.

That `??` fallback is deliberate, not defensive padding: the cloud web app talks to whatever recce version the instance image happens to carry, so a newer UI against an older backend keeps working instead of rendering an empty Base column. This exact version skew is how the customer landed on the bug.

Closes DRC-3975

*Verified against production artifacts, not just fixtures.* For the reporting session the two environments are genuinely different — base `METRODATA_PROD` (schemas `rpt`/`mart`/`stg`/`int`/`core`, 779 catalog nodes) versus current `METRODATA_STG.PARADIME_TURBO_CI_PR_1012_BASE`. The platform was resolving them correctly all along; only the display was wrong. Worth knowing while reviewing, because it means this is a presentation fix and carries no risk to which environments Recce actually queries.

This also restores `schemas_match` in `EnvInfo.tsx`, which has been unconditionally `true` since v1.45.0 because both `schema_count`s and the equality check derived from the same collapsed sets. We had no telemetry signal for three months — worth a glance at that dashboard once this ships.

Tests were written first and confirmed failing for the right reason (`'MergedNode' object has no attribute 'base_schema'`; `expected undefined to be 'public'`) before any implementation. Coverage includes the diverging case, the identical case (asserting the field stays off the wire), added nodes, removed nodes, and the older-backend fallback.

Checks: `make format` clean, `make flake8` clean, `1556 passed, 5 skipped`; `pnpm lint` clean, `pnpm type:check` clean, `186 files / 3932 tests passed`.

Not covered here: I verified at the unit level, not by opening the panel against a live instance. The ideal end-to-end check is `recce server` against the two real manifests, which needs a frontend build plus a Snowflake connection for anything past lineage.

```release-note
The Environment Information panel now shows the correct schema for the base environment. It previously repeated the current environment's schema in the Base column whenever the two environments used different schemas.
```

### DIFF
--- a/js/packages/ui/src/api/info.ts
+++ b/js/packages/ui/src/api/info.ts
@@ -75,6 +75,12 @@ export interface MergedNodeData {
   resource_type: string;
   package_name: string;
   schema?: string;
+  /**
+   * The base environment's schema, sent only when it differs from `schema`
+   * (which reports the current environment wherever the node exists there).
+   * Absent means "same as `schema`", not "unknown" (DRC-3975).
+   */
+  base_schema?: string;
   materialized?: string;
   tags?: string[];
   source_name?: string;
--- a/js/packages/ui/src/contexts/lineage/__tests__/utils.test.ts
+++ b/js/packages/ui/src/contexts/lineage/__tests__/utils.test.ts
@@ -56,6 +56,49 @@ describe("buildLineageGraph", () => {
     expect(node.data.materialized).toBe("table");
   });
 
+  test("resolves baseSchema from base_schema when the environments diverge", () => {
+    const lineage: MergedLineageResponse = {
+      nodes: {
+        "model.proj.orders": {
+          name: "orders",
+          resource_type: "model",
+          package_name: "proj",
+          schema: "ci_pr_1012",
+          base_schema: "prod_rpt",
+        },
+      },
+      edges: [],
+      metadata: { base: {}, current: {} },
+    };
+
+    const { nodes } = buildLineageGraph(lineage);
+
+    expect(nodes["model.proj.orders"].data.schema).toBe("ci_pr_1012");
+    expect(nodes["model.proj.orders"].data.baseSchema).toBe("prod_rpt");
+  });
+
+  test("falls back to schema for baseSchema when base_schema is absent", () => {
+    // Absent means "both environments share this schema" — and an older
+    // instance backend never sends the field at all, so the fallback is what
+    // keeps a newer UI correct against it.
+    const lineage: MergedLineageResponse = {
+      nodes: {
+        "model.proj.orders": {
+          name: "orders",
+          resource_type: "model",
+          package_name: "proj",
+          schema: "public",
+        },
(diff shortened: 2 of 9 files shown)


kentwelcome:
# Code Review: PR #1481

**SHA** `c1e8a6b` · **Verdict** GO

## Blockers
None.

## Issues
None.

## Notes
1. `js/packages/ui/src/utils/envUtils.ts:40` — The `?? node.data.schema` fallback re-derives what `buildLineageGraph` (`js/packages/ui/src/contexts/lineage/utils.ts:127`) already resolved, so the fallback now lives in two places while the PR description says it is resolved once. Harmless today — it only fires for graphs not built by `buildLineageGraph` (e.g. test mocks) — but it is a second source of truth for the same rule. Pass F.

## Limits
- Local full Python suite could not be collected (venv is Python 3.14; dbt/mashumaro fail at import — pre-existing, unrelated to this diff). Verified instead via `tests/test_merged_lineage.py` locally (35 passed) and the PR's green CI across Python 3.11–3.13 and the dbt matrix at this SHA.

---
Verification run: frontend `pnpm type:check` clean, `pnpm lint` clean, `pnpm test` 3932 passed / 5 skipped (186 files); security pass (data-flow trace from dbt manifest → wire → React render sinks) found no XSS/injection/PII exposure — both render sites are escaped React text nodes and telemetry sends only counts, never schema names.

kentwelcome (approved):

PR #1484, opened 2026-07-29
fix(mcp): do not report an unbuilt model's columns as removed

iamcxa:
## What was wrong

`schema_diff` compares each node's base and current column sets. Nodes come from the manifest, columns come from the catalog. A model that a selective build did not rebuild is therefore present on both sides with no column data on the current side, and every base column falls into `base - current` and is reported `removed`.

A customer merge request was told that 39 columns had been dropped from an unmodified model and that this was a possible schema regression. Nothing had been dropped. The same session's node-level metadata recorded that model as unchanged, so two of our own outputs contradicted each other and the customer was shown the wrong one. The claim was prominent, carried a "investigate before merge" action, and reproduced across two runs — the second summary cited the repetition as corroboration, which a deterministic bug will always supply.

## The change

A node is skipped when either side has no column data, and an empty column map counts as absence — a relation with no columns cannot exist, so an empty map means the catalog never described the model rather than that it lost everything.

Skipping happens before the row cap, so unverifiable rows can neither displace real changes nor inflate `more`.

Suppression on its own would trade a false positive for a false negative: both runtime prompts read an empty `data` as an affirmative "zero column changes". So the response carries a `schema_coverage` sibling — `status`, `unchecked_nodes`, `unchecked_node_count`, `more` — and one contract: **`data == []` means "no column changes" only when `status` is `complete`.** It sits beside the frame rather than inside `DataFrame`, which backs every tool's payload and has no business carrying a notion that means something only here.

## Verification

Driven with lineage nodes rebuilt from the incident session's own base and current manifests and catalogs, using the adapter's rule (node from the manifest, columns only when the catalog has that node), and with the tool arguments taken verbatim from that run's trace:

| | rows for the affected model | `schema_coverage` |
|---|---|---|
| before | **39, all `removed`** | absent |
| after | **0** | `partial`, model named in `unchecked_nodes` |

247 tests pass across `test_mcp_server.py`, `test_mcp_e2e.py` and `test_mcp_cloud_backend.py`. The new test takes its node shape from what the dbt adapter actually emits rather than a hand-authored one — an earlier attempt at this fix elsewhere passed its tests against a node shape the real producer cannot emit, and shipped nothing.

## Scope and known limits

- All in-server call sites route through this function, so the direct tool, `run_check`, and the check/report paths are covered together.
- `CloudBackend._tool_schema_diff` is a different implementation reading the instance lineage payload, and is deliberately untouched. On that path an unbuilt model contributes no rows at all, so it produces a silent omission rather than a false alarm. Tracked separately.
- **A genuine column drop on a model the run did not build is not reported as a drop.** It cannot be: the change has not been applied to the warehouse, so there is nothing to observe. The node still carries its node-level change status, and the model is named in `unchecked_nodes`, so the gap is stated rather than hidden.
- The `sqlmesh` adapter always attaches columns from the model definition, so it has no catalog-absence case and its behaviour is unchanged.
- **The consuming prompts need the matching half.** Until they are taught what `partial` means, an empty `data` still reads to them as a verified zero. That half lives in the consumer repo and should land in the same image build as this.

### DIFF
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -382,12 +382,18 @@ class CloudBackend:
             for column, change_status in column_changes.items():
                 changes.append((node_id, column, change_status))
         limit = 100
-        return DataFrame.from_data(
+        result = DataFrame.from_data(
             columns={"node_id": "text", "column": "text", "change_status": "text"},
             data=changes[:limit],
             limit=limit,
             more=len(changes) > limit,
         ).model_dump(mode="json")
+        # The cloud session hands us the column changes it already computed but
+        # says nothing about which nodes it could not describe, so coverage is
+        # unassessed rather than complete. Emitted under the same key as the
+        # local path so the one shared tool description holds for both backends.
+        result["schema_coverage"] = _schema_coverage(None)
+        return result
 
     async def _tool_impact_analysis(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
         info = await self._request("GET", "info")
@@ -436,6 +442,10 @@ class CloudBackend:
             "max_affected_row_count": 0,
             "confirmed_impacted_models": impacted_models,
             "confirmed_not_impacted_models": not_impacted_models,
+            # Metadata-only triage: the session reports the column changes it
+            # found but not what it failed to describe, so coverage is
+            # unassessed. Same key as the local path (see _schema_coverage).
+            "schema_coverage": _schema_coverage(None),
             "errors": [],
         }
 
@@ -832,12 +842,16 @@ class RecceMCPServer:
                     name="schema_diff",
                     description="Get the schema diff (column changes) between base and current environments. "
                     "Shows added, removed, and type-changed columns in compact dataframe format. "
-                    "Also returns schema_coverage. An empty `data` means 'no column changes' ONLY when "
-                    "schema_coverage.status is 'complete'. When it is 'partial', the models named in "
-                    "schema_coverage.unchecked_nodes have no column metadata on at least one side, so "
-                    "they were never compared — that is not evidence they are unchanged, and it is not "
-                    "evidence anything was removed. The commonest cause is a build that only builds what "
-                    "changed, but this tool does not observe the cause; do not report one as fact.",
+                    "Also returns schema_coverage, which says how much of the comparison could run. "
+                    "An empty `data` means 'no column changes' ONLY when schema_coverage.status is "
+                    "'complete'. When it is 'partial', the nodes listed in "
+                    "schema_coverage.unchecked_nodes — dbt node ids, e.g. 'model.<package>.<name>' — "
+                    "have no column metadata on at least one side, so they were never compared: that "
+                    "is not evidence they are unchanged, and it is not evidence anything was removed. "
+                    "The commonest cause is a build that only builds what changed, but this tool does "
+                    "not observe the cause; do not report one as fact. When it is 'unknown', coverage "
+                    "was not assessed at all (cloud-backed sessions), so an empty `data` is "
+                    "uncorroborated rather than verified.",
                     inputSchema={
                         "type": "object",
                         "properties": {
@@ -1450,6 +1464,18 @@ class RecceMCPServer:
 
                             Models with data_impact: 'potential' have unknown data impact — follow
                             the model's next_action field to investigate with profile_diff/query_diff.
+
+                            Schema results come with a coverage contract. A model's
+                            schema_changes: [] means it WAS compared and nothing changed;
+                            schema_changes: null means it was never compared, because one side has no
+                            column metadata — that is not evidence its schema is unchanged, and not
+                            evidence columns were removed. Top-level schema_coverage summarises this:
+                            'complete' = every impacted model was compared, 'partial' = the nodes in
+                            schema_coverage.unchecked_nodes were not, 'unknown' = the comparison did
+                            not run at all (see errors, or a cloud-backed session), so no
+                            schema_changes value on any model is corroborated. unchecked_nodes holds
+                            dbt node ids ('model.<package>.<name>') while confirmed_impacted_models
+                            entries key on the bare model name.
                         """
                         ).strip(),
                         inputSchema={
@@ -2323,6 +2349,8 @@ class RecceMCPServer:
                 "'potential' = no value_diff available (views, no PK, or skipped) "
                 "— follow the model's next_action to investigate. "
                 "Only models with next_action != null need further tool calls. "
+                "schema_changes: null means that model was never compared (see schema_coverage) — "
+                "not that its schema is unchanged; [] means compared and unchanged. "
                 "Note: incremental model value_diff may reflect "
                 "build window artifacts if not fully refreshed."
             ),
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -575,6 +575,60 @@ def _columns_comparable(base_node: dict, current_node: dict) -> bool:
     return bool(base_node.get("columns")) and bool(current_node.get("columns"))
 
 
+# Resource types dbt never describes in catalog.json, because the catalog
+# describes warehouse relations and these are not relations. See
+# `recce/adapter/dbt_adapter/__init__.py`, which builds them with no `columns`
+# key at all.
+_NON_RELATION_RESOURCE_TYPES = frozenset({"exposure", "metric", "semantic_model", "saved_query"})
+
+
+def _can_carry_catalog_columns(node: dict) -> bool:
+    """Whether the catalog could ever have described this node's columns.
+
+    Missing columns on a node dbt never catalogues are the normal state, not a
+    gap in what we checked. Exposures, metrics and semantic models are not
+    relations, and an ephemeral model is inlined as a CTE rather than
+    materialized, so none of them ever get a catalog entry. Counting them as
+    unchecked would pin `status` to "partial" on any project that owns one — a
+    genuinely clean schema_diff could then never be reported as verified — and
+    would spend the capped `unchecked_nodes` slots on nodes that were never
+    comparable to begin with.
+
+    A denylist, not an allowlist: a resource type we have not met, or a node
+    carrying no `resource_type` at all, still counts as a coverage gap. The
+    trade-off is deliberate. Skipping every node with no columns on *either*
+    side would also have covered these cases, but it would silently drop a model
+    that neither environment built — precisely the absence `unchecked_nodes`
+    exists to name — so we over-report a gap rather than hide one.
+    """
+    if node.get("resource_type") in _NON_RELATION_RESOURCE_TYPES:
+        return False
+    return (node.get("config") or {}).get("materialized") != "ephemeral"
+
+
+def _schema_coverage(unchecked_node_ids: Optional[List[str]]) -> Dict[str, Any]:
+    """Build the schema_coverage block shared by schema_diff and impact_analysis.
+
+    `None` means the comparison never ran — it raised, or the backend does not
+    report coverage. That is distinct from an empty list, which means the
+    comparison ran and found nothing it could not check.
+    """
+    if unchecked_node_ids is None:
+        return {
+            "status": "unknown",
+            "unchecked_nodes": [],
+            "unchecked_node_count": 0,
+            "more": False,
+        }
+    ordered = sorted(unchecked_node_ids)
+    return {
+        "status": "partial" if ordered else "complete",
+        "unchecked_nodes": ordered[:_UNCHECKED_NODE_LIMIT],
+        "unchecked_node_count": len(ordered),
+        "more": len(ordered) > _UNCHECKED_NODE_LIMIT,
+    }
+
+
 class RecceMCPServer:
     """MCP Server for Recce data validation tools"""
 
@@ -1744,9 +1798,13 @@ class RecceMCPServer:
             current_columns = current_node.get("columns") or {}
 
             # Skipped before the row cap so unverifiable rows can neither
-            # displace real changes nor inflate `more`.
+            # displace real changes nor inflate `more`. A node the catalog never
+            # describes is skipped without being named: it is not a gap in what
+            # we checked, and naming it would make `status` permanently
+            # "partial" on any project that owns one.
             if not _columns_comparable(base_node, current_node):
-                unchecked_nodes.append(node_id)
+                if _can_carry_catalog_columns(current_node):
+                    unchecked_nodes.append(node_id)
                 continue
 
             # Get column names in base and current
@@ -1791,13 +1849,7 @@ class RecceMCPServer:
         # that ignore it still see the unchanged shape, minus rows we cannot
         # stand behind. `data == []` means "no column changes" only when
         # `status` is "complete".
-        unchecked_nodes.sort()
-        result["schema_coverage"] = {
-            "status": "partial" if unchecked_nodes else "complete",
-            "unchecked_nodes": unchecked_nodes[:_UNCHECKED_NODE_LIMIT],
-            "unchecked_node_count": len(unchecked_nodes),
-            "more": len(unchecked_nodes) > _UNCHECKED_NODE_LIMIT,
-        }
+        result["schema_coverage"] = _schema_coverage(unchecked_nodes)
         return result
 
     async def _tool_row_count_diff(self, arguments: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


iamcxa:
Second commit, from an adversarial review of the first.

The reviewer found the same comparison duplicated inside `impact_analysis`, which read absent current columns as an empty set and emitted every base column as `removed` — the exact customer-visible claim, reachable through a second advertised tool. Fixing `schema_diff` alone would have left the incident live.

Both call sites now share one predicate. In `impact_analysis` an uncomparable model gets `schema_changes: null` rather than `[]`, since an empty list reads as "checked, nothing changed" — the one claim that cannot be made here. Coverage is reported alongside the models and starts as `unknown`, so a failure in that step cannot leave the response asserting complete coverage.

The tool description now carries the contract, because the description is what the consuming agent actually reads.

The regression guard is an end-to-end test on a real DuckDB context: a downstream model left out of the current catalog, driven through both tools. With the guard removed from `impact_analysis` it reports both of that model's columns as removed, so the test is load-bearing rather than decorative.

One existing test changed. It forced a schema-diff failure by setting `columns` to `None`, which the guard now treats as a legitimate "not built" rather than an error. Its intent — that a schema-diff failure must not block value-diff — is unchanged; it now forces the failure with a non-mapping value.

iamcxa:
Second reviewer landed after the first and found something both I and the other reviewer missed, so it is worth stating plainly.

Both the tool description and the prompts asserted that unchecked models "were not built by this run". The code observes only that column metadata is missing on one side. A build that skipped them is the commonest reason, not the observed one — and turning an absence into a confident statement about what happened is precisely the error this change exists to remove. I made it again, one level up, in the fix for it.

Both repos now describe the absence and say the cause is not observed. The reviewer also warned that the blunt instruction "never tell the reader to fix their pipeline" could suppress a genuine failure — a modified model that should have been built and was not. That is right, so it became a concrete forbidden example with its reason attached rather than a flat prohibition, and naming a cause is allowed when a tool result independently shows one.

Two residuals it raised that are NOT addressed here, recorded so they are not lost:

- Schema facts travel as ungrounded prose. If the model ignores these rules and some unrelated numeric claim grounds, the server can still publish a fabricated structural statement. Grounding is per-summary, not per-sentence. That is a structural limit of the evidence ledger, not something prompt text can close.
- There is no prompt regression test and no Haiku eval for these rules, so their behavioural sufficiency is unverified. On paper the weaker model gets the same concrete examples as the orchestrator; in practice nobody has measured it.

Q-A was answered the same way by both reviewers and by our own code: this is our defect. `recce_cloud/README.md` asks for `dbt docs generate` and never requires a complete catalog or any minimum coverage, and the incident catalog proves generation ran. No documented customer contract was broken.

gcko:
# Code Review: PR #1484

**SHA** `f394014d` · **Verdict** NO-GO

## Blockers

1. `recce/mcp_server.py:2006` — In `_tool_impact_analysis` the guard fires on **added and removed** models, discarding their real column changes and mislabelling them "unchecked". `_tool_schema_diff` pre-filters to `set(current) & set(base)` at line 1724, so added/removed nodes never reach its guard; the `impacted_models` loop has no such filter, and the copied guard's "Same comparison as the schema_diff tool" comment is therefore false.
   Evidence: driving `_tool_impact_analysis({"skip_value_diff": True})` with a model present only in `current` (`change_status: "added"`):
   | | `schema_changes` | `next_action` |
   |---|---|---|
   | `origin/main` | `[{id, added}, {new_col, added}]` | `columns=['id','new_col']`, reason `schema changed, value_diff unavailable` |
   | this branch | `None` | `columns=None`, reason `modified model, value_diff unavailable (view or no PK)` |

   and with a model present only in `base` (`change_status: "removed"`): `[{id, removed}, {gone, removed}]` → `None`. Both are named in `schema_coverage.unchecked_nodes`. A new model's columns *are* all added and a deleted model's columns *are* all removed — those are observations, not absences, and the `change_status` needed to tell them apart is already on the entry at line 1929.
   Pass A / F.

2. `recce/mcp_server.py:1748` — `nodes_to_compare` (line 1724) is every resource type in both lineages, and several kinds **never** carry catalog columns, so they permanently land in `unchecked_nodes` and pin `status` to `"partial"`.
   Evidence: `recce/adapter/dbt_adapter/__init__.py:875-900` builds exposures, metrics and semantic models with no `columns` key at all; line 841 assigns `columns` only when `unique_id in catalog.nodes`, which excludes ephemeral models. Probe — one clean model plus one exposure, nothing changed on either side:
   ```
   data: []
   schema_coverage: {'status': 'partial', 'unchecked_nodes': ['exposure.project.dash'],
                     'unchecked_node_count': 1, 'more': False}
   ```
   Two consequences. The `"complete"` branch of the new contract never fires on a project with a single exposure or ephemeral model, so a genuinely clean schema_diff can never be reported as verified — the false positive becomes a permanent false "unverified". And the 50-slot `unchecked_nodes` list fills with non-relations that displace the unbuilt models the field exists to name. Neither is caught by the new tests, whose fixtures contain only models (`tests/test_mcp_e2e.py:801` asserts `complete`).
   Pass A / F.

## Issues

3. `recce/mcp_server.py:1380` — The `impact_analysis` description is unchanged, so the new top-level `schema_coverage` (line 2278) and the `schema_changes: null` vs `[]` distinction (line 2007) are invisible to the agent. `AGENTS.md`/`CLAUDE.md`: "MCP tool description = LLM agent contract. Description MUST match actual response format." The `schema_diff` description got the contract written into it at lines 780-786; the identical semantics on this tool got nothing, and `_guidance` at line 2263 does not mention either field.
   Evidence: `grep -n "schema_coverage" recce/mcp_server.py` → only 1795, 2028, 2278; no hit inside the `impact_analysis` `Tool(...)` block at 1379-1426.
   Pass C.

4. `recce/mcp_server.py:780` — The description asserts unconditionally "Also returns schema_coverage", but `CloudBackend._tool_schema_diff` (line 373-390) returns a bare `DataFrame` dump with no such key. `list_tools` (line 711) is one shared registry and dispatch routes to `self.backend.call_tool` at line 1496, so a cloud-mode agent reads a contract the payload cannot satisfy — and the stated rule (`data == []` means no changes only when `status` is `complete`) has no branch for an absent field, so an empty `data` becomes unreadable rather than merely uncorroborated.
   Evidence: line 390 `).model_dump(mode="json")` is the whole cloud return; the PR body confirms the cloud path is deliberately untouched, but the description covering it was still widened.
   Pass C / F.

5. `tests/test_mcp_e2e.py:853-856` — The `impact_analysis` half of the new test asserts nothing that can fail. The `for`/`if model["name"] == "not_rebuilt"` loop passes trivially if the model is absent from `confirmed_impacted_models`, and `status in ("partial", "unknown")` accepts `"unknown"`, which is reachable only when the `schema_diff` step raised (line 2035) — i.e. the test passes when the code under test crashes.
   Evidence: same fixture, instrumented — `COVERAGE: {'status': 'partial', ...}`, `ERRORS: []`. The value is deterministic, so the disjunction buys nothing but tolerance for failure. Both blockers above also survive this file untouched.
   Pass E.

## Notes

6. `recce/mcp_server.py:1799` — `_UNCHECKED_NODE_LIMIT` truncation and `more: True` have no test on either call site; the >50 path ships unexercised.

7. `recce/mcp_server.py:2011` — The column-set comparison is duplicated between `_tool_schema_diff` (1752-1769) and `_tool_impact_analysis` (2011-2025). Blocker 1 is that duplication's cost: the guard was extracted but the intersection filter it depends on was not.

8. `recce/mcp_server.py:1797` — `unchecked_nodes` carries node IDs while the description (line 782) calls them "the models named in" and `confirmed_impacted_models` entries key on `name`, leaving the consumer to map between the two.

## Verification

`make test` → 1558 passed, 5 skipped. `make check` / `make flake8` clean. Both blockers were reproduced against `f394014d` and diffed against `origin/main` (`f9f7c7bc`) in a separate worktree.

gcko (dismissed):

recce/mcp_server.py
```diff
@@ -1928,17 +1979,37 @@ async def _tool_impact_analysis(self, arguments: Dict[str, Any]) -> Dict[str, An
...
+                base_node = base_nodes.get(node_id, {})
+                current_node = current_nodes.get(node_id, {})
+
+                # Same comparison as the schema_diff tool, so it needs the same
+                # guard. `None` rather than `[]`: an empty list reads as "checked,
+                # nothing changed", which is the claim we cannot make here.
+                if not _columns_comparable(base_node, current_node):
+                    model["schema_changes"] = None
+                    schema_unchecked.append(node_id)
+                    continue
```
gcko: Blocker 1 — added and removed models reach this guard and lose their real column changes; `_tool_schema_diff` never sees them because it pre-filters to `set(current_nodes) & set(base_nodes)` at line 1724.

```suggestion
                # Same absence guard as the schema_diff tool — but that tool
                # compares only nodes present in BOTH lineages (line 1724).
                # Here added and removed models also arrive, and a one-sided
                # node is an observation, not an absence: a new model's columns
                # really are all added, a deleted one's really are all removed.
                # `None` rather than `[]`: an empty list reads as "checked,
                # nothing changed", which is the claim we cannot make here.
                exists_in_both = bool(base_node) and bool(current_node)
                if exists_in_both and not _columns_comparable(base_node, current_node):
                    model["schema_changes"] = None
                    schema_unchecked.append(node_id)
                    continue
```

tests/test_mcp_e2e.py
```diff
@@ -796,6 +798,62 @@ async def test_schema_diff_detects_added_column(self, mcp_e2e):
...
+        assert coverage["status"] == "partial"
+        assert "model.recce_test.not_rebuilt" in coverage["unchecked_nodes"]
+
+        # The same comparison is duplicated inside impact_analysis, so the same
+        # model must not come back as a wholesale drop through that tool either.
+        impact = await server._tool_impact_analysis({})
+        for model in impact["confirmed_impacted_models"]:
+            if model["name"] == "not_rebuilt":
+                assert model["schema_changes"] is None, "impact_analysis compared a model the run did not build"
+        assert impact["schema_coverage"]["status"] in ("partial", "unknown")
```
gcko: Issue 5 — the loop asserts nothing if `not_rebuilt` is absent, and `"unknown"` is reachable only when the schema step raised at line 2035, so the test passes when the code under test crashes. Instrumenting this same fixture gives `status: 'partial'`, `errors: []` deterministically.

```suggestion
        impact = await server._tool_impact_analysis({})
        by_name = {model["name"]: model for model in impact["confirmed_impacted_models"]}
        assert "not_rebuilt" in by_name, "not_rebuilt dropped out of the blast radius"
        assert by_name["not_rebuilt"]["schema_changes"] is None, "impact_analysis compared a model the run did not build"
        assert impact["schema_coverage"]["status"] == "partial"
```

recce/mcp_server.py
gcko: Confirmed and fixed in 6ed06639. Your reading is exactly right: the guard was copied but the `set(current) & set(base)` filter it depends on was not, so `impacted_models` hands it one-sided nodes that are indistinguishable from an unbuilt model.

I took the shape of your suggestion, with the catalog check folded in so the ephemeral/non-relation case from blocker 2 falls through here too:

```python
exists_in_both = bool(base_node) and bool(current_node)
if (
    exists_in_both
    and not _columns_comparable(base_node, current_node)
    and _can_carry_catalog_columns(current_node)
):
    model["schema_changes"] = None
    schema_unchecked.append(node_id)
    continue
```

Falling through is safe for a one-sided node: `base_node` is `{}` for an added model, so `base_cols` is empty and every current column reports `added`; the mirror holds for `removed`. I also moved the column maps into locals read as `base_node.get("columns") or {}` — a one-sided node can now reach that code, and a `None` map would otherwise raise into the `except` and reset coverage to `"unknown"`.

New test, red before the commit and green after — `tests/test_mcp_server.py::TestImpactAnalysisBehavior::test_one_sided_models_keep_their_column_changes`: one model only in `current` (`added`), one only in `base` (`removed`), asserting the full column sets come back with the right `change_status`, that neither is `None`, and that `schema_coverage.status` is `"complete"` rather than counting them as a gap.

tests/test_mcp_e2e.py
gcko: Agreed — the loop was vacuous and the disjunction tolerated a crash. Applied in c2f1ce7d, essentially your suggestion plus an explicit `errors` assertion so a raised schema step fails loudly instead of being absorbed by the status check:

```python
impact = await server._tool_impact_analysis({})
by_name = {model["name"]: model for model in impact["confirmed_impacted_models"]}
assert "not_rebuilt" in by_name, "not_rebuilt dropped out of the blast radius"
assert by_name["not_rebuilt"]["schema_changes"] is None, "impact_analysis compared a model the run did not build"
assert impact["errors"] == []
assert impact["schema_coverage"]["status"] == "partial"
assert "model.recce_test.not_rebuilt" in impact["schema_coverage"]["unchecked_nodes"]
```

gcko:
Thanks — both blockers were real, and both reproduce as described. Fixed across three commits; nothing was dismissed.

| # | Finding | Verdict | Action | Commit |
|---|---------|---------|--------|--------|
| 1 | added/removed models lose real column changes | Valid | `exists_in_both` guard + fall-through | `6ed06639` |
| 2 | non-relation / ephemeral nodes pin `status` to `partial` | Valid | `_can_carry_catalog_columns` denylist | `6ed06639` |
| 3 | `impact_analysis` description omits the contract | Valid | description + `_guidance` rewritten | `2dc4ac75` |
| 4 | cloud payload cannot satisfy the widened description | Valid | cloud now emits `status: "unknown"` | `2dc4ac75` |
| 5 | test asserts nothing that can fail | Valid | look up by name, exact status, `errors == []` | `c2f1ce7d` |
| 6 | `_UNCHECKED_NODE_LIMIT` truncation untested | Valid | test added | `6ed06639` |
| 7 | column comparison duplicated across two tools | Partial | coverage block deduped; guards now intentionally differ | `6ed06639` |
| 8 | `unchecked_nodes` holds ids, models key on name | Valid | both descriptions state it | `2dc4ac75` |

### Blockers 1 and 2

Detail is in the replies on the two inline threads. Blocker 1 took the shape of your suggestion with the blocker-2 catalog check folded in.

Blocker 2 was a genuine design choice, so here is the reasoning. I filtered on whether the catalog could ever have described the node, as a **denylist**:

```python
_NON_RELATION_RESOURCE_TYPES = frozenset({"exposure", "metric", "semantic_model", "saved_query"})

def _can_carry_catalog_columns(node: dict) -> bool:
    if node.get("resource_type") in _NON_RELATION_RESOURCE_TYPES:
        return False
    return (node.get("config") or {}).get("materialized") != "ephemeral"
```

The alternative — skip any node with no columns on *either* side — is cheaper and covers ephemeral models for free, but it silently drops a model that neither environment built, which is precisely the absence `unchecked_nodes` exists to name. The trade-off I accepted instead is over-reporting: an unfamiliar `resource_type`, or a node carrying none at all, still counts as a gap. A denylist rather than an allowlist for the same reason — an allowlist would make every future resource type invisible by default, and a coverage field that quietly under-reports is worse than one that occasionally names a node it did not need to. The rationale is in the helper's docstring, not just here.

Ephemeral is covered explicitly rather than incidentally: `config.materialized == "ephemeral"`, since `_build_lineage` assigns `columns` only when `unique_id in catalog.nodes` and dbt never catalogues an inlined CTE.

Tests, each red before `6ed06639` and green after:

- `test_tool_schema_diff_coverage_ignores_nodes_the_catalog_never_describes` — one clean model plus an exposure, a metric, a semantic model and an ephemeral model, nothing changed. Was `partial` with `unchecked_nodes` non-empty; now `complete` with `unchecked_nodes == []`, so the `"complete"` branch can fire on a project that owns one.
- `test_ephemeral_model_is_not_a_coverage_gap` — same for the `impact_analysis` call site.
- `test_one_sided_models_keep_their_column_changes` — blocker 1.

### Issue 4 — cloud

I widened the payload rather than narrowing the description. Both cloud tools now emit `schema_coverage` with `status: "unknown"`, which is the truthful value: the session hands over the column changes it computed but never says what it could not describe. That is already what `"unknown"` means on the local `impact_analysis` path, so the enum did not grow, one shared description now holds for both backends, and the "unknown" branch is spelled out — an empty `data` there is uncorroborated rather than unreadable. Narrowing the description instead would have left a documented field with no defined behaviour in half the sessions.

### Note 7 — the duplication

Partially addressed, deliberately. The three copies of the coverage block are now one `_schema_coverage(unchecked_node_ids)` helper (`None` → `"unknown"`, `[]` → `"complete"`), which is where the truncation and sort bugs would have diverged. I did not merge the two column-comparison loops: after blocker 1 they are no longer the same comparison — `schema_diff` sees intersected nodes only and emits rows, `impact_analysis` sees the whole blast radius including one-sided nodes and mutates per-model entries. Collapsing them would mean threading that difference through a shared function, which reintroduces the coupling that caused blocker 1 rather than removing it. Happy to take it as a follow-up if you would rather see the loop unified.

### Verification

`make check` and `make flake8` clean. `python3 -m pytest tests` → **1562 passed, 5 skipped** (1558 before, plus the four new tests). Each blocker and issue fix was confirmed red before its commit and green after; note 6's truncation test passes both before and after by design — it covers previously unexercised behaviour rather than a defect.

gcko (approved):
Code Review Loop: all findings addressed on c2f1ce7d. Blockers re-verified against origin/main behaviour; 1562 passed, lint clean, CI green.

PR #1489, opened 2026-08-01
docs(agents): rightsize always-loaded agent context, and correct two claims found in review

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`docs` — agent-context files only. No product code, no Python, no TypeScript.

`CLAUDE.md` and `AGENTS.md` load into every session; a skill body loads only when the
skill is invoked. The always-resident pair had drifted into a mix of stale facts and
material that is trivially derivable from `ls`, `make help`, or a manifest.

Two commits, kept separate so the second is reviewable on its own:

**1. `docs(agents): rightsize always-loaded agent context`**

- `AGENTS.md` — removed the Tech Stack table (it claimed MUI 7 / Biome 2.4; actual
  overrides are `@mui/material: ^9.0.0` and `@biomejs/biome: 2.5.5`, so it was
  actively wrong), the Repository Layout and Essential Commands tables (derivable
  from `ls` and `make help`, both verified to cover what was removed), and three
  Common Pitfalls rows that restated commands already in the Development Workflow
  block. Added a "Non-obvious conventions" section holding the pnpm and
  Black/isort/flake8 facts relocated from `CLAUDE.md` plus two gotchas that have
  each cost real debugging time.
- `CLAUDE.md` — moved the MCP response-format contracts and the cloud-vs-local
  exception-type trap into `recce-mcp-dev/SKILL.md`, leaving a pointer; compressed
  the dependency-update and PR-review procedures to skill pointers plus the
  prerequisites and API endpoints that are not in those skills.
- `recce-mcp-dev` — received the moved blocks; corrected the "All 8 tools" coverage
  claim; dropped the "Python 3.9: `Union[X, Y]`" pitfall, obsolete under
  `requires-python = ">=3.10"`.
- `recce-mcp-e2e` — named the 8 tools the harness actually covers.

`AGENTS.md`'s **Critical Constraints block is unchanged, byte for byte** (md5
`0e23fd0fb248e634318603cbb8f3ed34` before and after). No prohibition was narrowed,
no covered category dropped, no absolute softened.

**2. `docs(agents): correct hook-system and MCP tool-count claims`**

An adversarial review of commit 1 found two factual errors and four smaller gaps.
Every claim below was verified against the code, not inferred.

- **`AGENTS.md` said the two pre-commit systems were mutually exclusive, and that
  husky mode bypassed black/isort/flake8. Both false.** `js/.husky/pre-commit` ends
  in a committed `# start templated` block that `exec`s
  `pre-commit hook-impl --config=.pre-commit-config.yaml`. Confirmed by running
  `bash js/.husky/pre-commit` under `core.hooksPath=js/.husky/_` and watching
  black/isort/flake8 execute, and by `git show HEAD:js/.husky/pre-commit` — it is
  committed, not a local artifact. Husky mode is a superset (biome guard +
  `pnpm lint:staged` + all Python hooks); `make install-dev` is a strict downgrade,
  and `.git/hooks/pre-commit` has no chain back. The old text told the reader to
  "re-run whichever installer you need", which points at the one command that
  actually drops a guardrail.
- **`list_tools` does not register 20 tools unconditionally — 20 is the maximum.**
  7 register always, `analyze_model` only when `self.backend is None`, and 12 only
  when `self.mode == RecceServerMode.server`. Preview and read-only mode expose 8;
  cloud mode never exceeds 19. The prescribed `grep -c 'Tool('` recipe counts code
  sites rather than any session's surface, reproducing the class of error it was
  added to prevent. Replaced with a per-mode table in both skills.
- The **third** "all 8 tools" instance in `recce-mcp-dev` survived commit 1's
  correction.
- `recce-mcp-e2e`'s `description:` still promised "full E2E verification" while its
  body said the opposite — and the description is the part that stays resident.
- The `CLAUDE.md` MCP pointer omitted the `recce/apis/*_api.py` trigger that the
  skill description carries. `grep -rln 'except RecceException' recce/apis/` returns
  `run_api.py`, `check_api.py`, `check_events_api.py`, all three of which can regress.
- The two new skill pointers were ambiguous across three scopes (project, user,
  plugin) — now plugin-qualified.
- Restored "Keep responses concise and action-oriented", the one dropped line that
  is not reconstructible from `ls`, `make help`, or a manifest.

None filed.

Verification run for this branch:

- `python -m pytest tests/ -q` → **1578 passed, 5 skipped**. No product code changed;
  this confirms the tree is clean.
- Tool inventory: `grep -c 'Tool(' recce/mcp_server.py` → 20, matching 20 unique
  `name="…"` values. The per-mode split was read off the `if` guards at
  `recce/mcp_server.py:929` (`self.backend is None`) and `:1071`
  (`self.mode == RecceServerMode.server`).
- E2E coverage: cross-checked each tool name against
  `.claude/skills/recce-mcp-e2e/test_mcp_e2e_template.py` — 8 present, 12 absent.
  The template itself is deliberately untouched so it stays runnable.
- `make help` (`Makefile:19`) was confirmed to print `make check`, `make mypy`,
  `make flake8` and `make deps-check-frontend`, i.e. everything the deleted Essential
  Commands table listed.
- Exception hierarchy: `RecceCloudException(Exception)` at
  `recce/util/recce_cloud.py:31` vs `RecceException(Exception)` at
  `recce/exceptions.py:1` — unrelated, as the skill states.

Two judgment calls worth a second opinion:

1. The `recce-dev:`-qualified skill names in `CLAUDE.md`. A bare
   `pr-review-response` currently resolves to a personal user-level skill; naming the
   plugin explicitly seemed right for a checked-in file, but revert if the team
   prefers the bare name.
2. Commit 1 removes `make check` from always-loaded context on the grounds that
   `make help` covers it. Verified true, but it is the one command that genuinely
   disappeared rather than being restated elsewhere.

Corresponding fixes to the local agent-memory files (the wrong hook claim, an
"every value drifted" overstatement that was really 8 of 9, and a cozempic durability
note that named a non-existent function) were applied outside the repo and are not
part of this diff.

```release-note
NONE
```

### DIFF
--- a/.claude/skills/recce-mcp-dev/SKILL.md
+++ b/.claude/skills/recce-mcp-dev/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: recce-mcp-dev
-description: Use when modifying recce/mcp_server.py, MCP tool handlers, error classification, or MCP-related tests. Also use when adding new MCP tools or changing tool response formats.
+description: Use when modifying recce/mcp_server.py, MCP tool handlers, error classification, or MCP-related tests. Also use when adding new MCP tools or changing tool response formats, or when adding a CheckDAO/RunDAO write in recce/apis/*_api.py (cloud-mode exceptions bypass `except RecceException`).
 ---
 
 # Recce MCP Server Development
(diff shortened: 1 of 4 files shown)


kentwelcome (approved):
LGTM

PR #1504, opened 2026-08-11
fix(mcp): support both mcp 1.x and 2.0 SDKs

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` — compatibility fix for the optional `mcp` dependency.

`recce mcp-server` fails to start against `mcp` 2.0:

```
[Error] Failed to start MCP server: 'Server' object has no attribute 'list_tools'
```

mcp 2.0 removed the low-level `Server` decorators (`@server.list_tools()`,
`@server.call_tool()`). Handlers are now passed to the constructor and use a
`(ctx, params) -> Result` signature. This PR supports both majors rather than
cutting over, so users on `mcp` 1.x keep working.

- `MCP_V2` flag selects the registration path at import time.
- `_setup_handlers` becomes `_make_handlers`, returning the same two handler
  bodies unchanged — still `List[Tool]` / `List[TextContent]`, still raising on
  error. On 1.x they are registered with the old decorators; on 2.0 the thin
  `_handle_list_tools` / `_handle_call_tool` adapters wrap them into
  `ListToolsResult` / `CallToolResult`.
- `_handle_call_tool` catches exceptions and returns `isError=True` explicitly.
  1.x did that inside the decorator; 2.0 turns a raised exception into a
  JSON-RPC protocol error instead, which an agent reads as a transport failure
  rather than a tool failure. Without this the error contract would silently
  change on 2.0.
- Tests move to `tests/mcp_compat.py` helpers instead of poking
  `server.server.request_handlers[...]` (that dict is gone in 2.0) and instead
  of reading `Tool.inputSchema` (renamed `input_schema` in 2.0).
- Dependency pin widened from `mcp~=1.23` to `mcp>=1.23,<3`.

Second commit adds the regression check that was missing. Nothing in CI ever
started `recce mcp-server`, which is why a startup failure shipped:

- `smoke_test.sh` gains an `mcp-server` mode. `recce mcp-server` speaks stdio,
  so there is no port to poll — the check feeds it `initialize` /
  `notifications/initialized` / `tools/list` and asserts the server names itself
  and advertises at least one tool.
- `SMOKE_SERVER` selects the surface and defaults to `server`, so every existing
  caller behaves exactly as before. `SMOKE_MCP_VERSION` pins which SDK version
  to install, because `mcp` is an optional extra that CI's `uv sync` does not
  carry.
- A new `mcp-smoke-test` job runs that mode against `mcp` 1.28.1 and 2.0.0.

None.

Verified on both majors, since CI only exercises one:

| | mcp 1.28.1 | mcp 2.0.0 |
|---|---|---|
| `tests/test_mcp_server.py` + `test_mcp_cloud_backend.py` | pass | pass |
| Full suite (`pytest tests/`) | 1581 passed, 5 skipped | n/a — see below |
| Real `ClientSession` over memory streams | 20 tools, call OK, `isError=True` on failure | identical |

The wire check ran an actual MCP client against the server on both versions to
confirm registration, a successful call, and the error path produce the same
result.

The new smoke-test function was exercised locally three ways: against a real
`recce mcp-server` on mcp 2.0.0 and on 1.28.1 (both reported 20 tools), and
against a stub that reproduces the original startup failure, which fails the
check with exit 1 as intended.

Two things worth knowing:

- `tests/test_mcp_e2e.py` could not run in my mcp-2.0 environment for an
  unrelated reason: its duckdb rejects dbt-duckdb's `python_scan_all_frames`
  setting during fixture setup. It passes fully on the mcp 1.28 environment
  (60 passed), which is the version CI installs.
- mcp 2.0 pulls in a new transitive dependency, `httpx2`. Worth a `pip check`
  in a clean environment before anyone pins forward to 2.0.

The 1.x branch in `_build_server` can be deleted once the floor moves to
`mcp>=2`.

```release-note
`recce mcp-server` now works with both mcp 1.x and mcp 2.0. Previously the
server failed to start on mcp 2.0 with "'Server' object has no attribute
'list_tools'".
```

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -14,6 +14,9 @@ on:
       - "js/**"
       - "recce/data/**"
 
+permissions:
+  contents: read
+
 jobs:
   smoke-test:
     if: github.actor != 'dependabot[bot]'
@@ -57,7 +60,7 @@ jobs:
       # `mcp` is an optional extra pinned to >=1.23,<3, and the two majors
       # register tool handlers differently. Both have to boot.
       matrix:
-        mcp-version: ["1.28.1", "2.0.0"]
+        mcp-version: ["1.29", "2.0"]
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -49,3 +49,33 @@ jobs:
         run: |
           source .venv/bin/activate
           ./integration_tests/dbt/smoke_test.sh
+
+  mcp-smoke-test:
+    if: github.actor != 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    strategy:
+      # `mcp` is an optional extra pinned to >=1.23,<3, and the two majors
+      # register tool handlers differently. Both have to boot.
+      matrix:
+        mcp-version: ["1.28.1", "2.0.0"]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+
+      - name: Install Recce and dbt
+        run: |
+          uv venv
+          uv sync --no-dev --python 3.12
+          uv pip install dbt-core dbt-duckdb
+
+      - name: Run smoke test - mcp server
+        env:
+          SMOKE_SERVER: mcp-server
+          SMOKE_MCP_VERSION: ${{ matrix.mcp-version }}
+        run: |
+          source .venv/bin/activate
+          ./integration_tests/dbt/smoke_test.sh
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -82,3 +82,13 @@ jobs:
         run: |
           source .venv/bin/activate
           ./integration_tests/dbt/smoke_test.sh
+
+      # The step above is the only place either major gets installed: the tox envs
+      # that run pytest resolve one mcp release, so on their own they leave the other
+      # major's compat path untested. The smoke test proves the server boots; these
+      # prove the handlers behave.
+      - name: Run MCP tests
+        run: |
+          source .venv/bin/activate
+          uv pip install pytest pytest-asyncio pandas duckdb
+          python -m pytest tests/test_mcp_server.py tests/test_mcp_cloud_backend.py tests/test_mcp_e2e.py -q
--- a/integration_tests/dbt/smoke_test.sh
+++ b/integration_tests/dbt/smoke_test.sh
@@ -9,7 +9,8 @@ pwd
 SMOKE_SERVER="${SMOKE_SERVER:-server}"
 # Only used when SMOKE_SERVER=mcp-server. Major.minor, e.g. "1.29" or "2.0":
 # the mcp SDK majors register tool handlers differently, so the version under
-# test has to be explicit, but `~=` still picks up the latest patch release.
+# test has to be explicit. `~=` keeps it a floor, not a pin — `~=1.29` is
+# `>=1.29,<2.0`, so later patch *and* minor releases are picked up.
 SMOKE_MCP_VERSION="${SMOKE_MCP_VERSION:-2.0}"
 
 case "$SMOKE_SERVER" in
@@ -159,18 +160,25 @@ function check_mcp_server_status() {
     # SSE lines are CRLF-terminated; a trailing CR makes the POST url malformed.
     endpoint=$(awk '/^data: \//{sub(/^data: /,""); sub(/\r$/,""); print; exit}' "$stream")
 
+    # ids 3 and 4 are the error contract, which is where the two SDK majors actually
+    # diverge: 1.x turns a raised exception and a schema violation into isError, 2.0
+    # would turn the former into a transport error and skip the latter entirely.
+    # `tools/list` alone cannot see either, so call a tool both ways.
     for request in \
         '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}' \
         '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
-        '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'; do
+        '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}' \
+        '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"get_server_info","arguments":{}}}' \
+        '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"lineage_diff","arguments":{"select":123}}}'; do
         if ! curl -sf -o /dev/null -X POST "$base$endpoint" -H 'Content-Type: application/json' -d "$request"; then
             echo "The MCP server rejected a request: $request"
             exit 1
         fi
     done
 
-    if ! timeout 20 bash -c "until grep -q '\"id\":2' '$stream'; do sleep 0.5; done"; then
-        echo "The MCP server did not answer tools/list."
+    # Requests are answered in order on one session, so id 4 arriving means all of them have.
+    if ! timeout 20 bash -c "until grep -q '\"id\":4' '$stream'; do sleep 0.5; done"; then
+        echo "The MCP server did not answer every request."
         exit 1
     fi
 
@@ -179,11 +187,29 @@ function check_mcp_server_status() {
     server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' <<< "$responses")
     tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' <<< "$responses")
     assert_string_value "$server_name" "recce"
+    # A count alone is a weak gate — server mode advertises 20 tools, so `>= 1` still
+    # passes with 19 of them silently unregistered. Name one that must be there.
+    if ! jq -e 'select(.id == 2) | [.result.tools[].name] | index("lineage_diff")' > /dev/null <<< "$responses"; then
+        echo "The MCP server did not advertise lineage_diff."
+        exit 1
+    fi
     if [ "${tool_count:-0}" -lt 1 ]; then
         echo "The MCP server started but advertised no tools."
         exit 1
     fi
     echo "MCP server is up and advertised $tool_count tools."
+
+    if ! jq -e 'select(.id == 3) | .result.isError != true' > /dev/null <<< "$responses"; then
+        echo "The MCP server failed a get_server_info tool call."
+        exit 1
+    fi
+    # 123 violates lineage_diff's `select: string` schema. mcp 1.x rejects this in the
+    # SDK; on 2.0 the adapter has to, or the argument reaches the tool coerced.
+    if ! jq -e 'select(.id == 4) | .result.isError == true and (.result.content[0].text | test("Input validation error"))' > /dev/null <<< "$responses"; then
+        echo "The MCP server did not reject a schema-invalid tool argument."
+        exit 1
+    fi
+    echo "MCP server tool calls behave correctly on success and on invalid input."
 
     echo "Stopping the MCP server..."
     kill $(jobs -p) 2>/dev/null || true
@@ -193,6 +219,10 @@ function check_mcp_server_status() {
 
 if [ "$SMOKE_SERVER" = "mcp-server" ]; then
     echo "Starting the MCP server..."
+    # Every `exit 1` inside the check fires with the server already backgrounded, and
+    # after the SSE reader starts, with that too. Both inherit this step's stdout, so
+    # without a trap a failed smoke test keeps the CI step open with nothing left to say.
+    trap 'kill $(jobs -p) 2>/dev/null || true' EXIT
     recce mcp-server --sse --port "$MCP_PORT" &
     check_mcp_server_status
 else
--- a/integration_tests/dbt/smoke_test.sh
+++ b/integration_tests/dbt/smoke_test.sh
@@ -129,46 +129,72 @@ function check_server_status() {
 }
 
 # Recce MCP Server
-# Stdio, not HTTP: there is no port to poll, so the server is started with the
-# handshake on stdin and its replies land in a file. The startup path is what
-# breaks (tool registration differs between mcp 1.x and 2.0), so the handshake
-# plus a non-empty tool list is the whole check.
-MCP_REQUESTS=$(mktemp)
-MCP_RESPONSES=$(mktemp)
-MCP_STDERR=$(mktemp)
-cat > "$MCP_REQUESTS" << 'EOF'
-{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}
-{"jsonrpc":"2.0","method":"notifications/initialized"}
-{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}
-EOF
+# The MCP server talks HTTP/SSE: responses arrive on the GET /sse stream, and
+# requests are POSTed to the session endpoint that stream hands out in its first
+# event. Liveness alone is not enough — tool registration differs between mcp
+# 1.x and 2.0, so the handshake plus a non-empty tool list is the real check.
+MCP_PORT=8765
 
-# Takes the pid of the backgrounded `recce mcp-server`. The server exits once it
-# reaches the end of the request file, so waiting on it is the readiness signal.
 function check_mcp_server_status() {
-    local mcp_pid="$1"
+    local base="http://localhost:$MCP_PORT"
+    local stream stream_pid endpoint responses server_name tool_count
+    stream=$(mktemp)
+
     echo "Waiting for the MCP server to respond..."
-    if ! wait "$mcp_pid"; then
-        echo "Failed to start the MCP server."
-        cat "$MCP_STDERR"
+    if ! timeout 60 bash -c "until curl -sf $base/health > /dev/null; do
+    echo \"MCP server not ready yet...\"
+    sleep 2
+    done"; then
+        echo "Failed to start the MCP server within the time limit."
         exit 1
     fi
 
-    local server_name tool_count
-    server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' "$MCP_RESPONSES")
-    tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' "$MCP_RESPONSES")
+    # The response stream has to be open before any request is sent.
+    curl -sN "$base/sse" > "$stream" &
+    stream_pid=$!
+    if ! timeout 20 bash -c "until grep -q '^data: /' '$stream'; do sleep 0.5; done"; then
+        echo "The MCP server did not hand out a session endpoint."
+        exit 1
+    fi
+    # SSE lines are CRLF-terminated; a trailing CR makes the POST url malformed.
+    endpoint=$(awk '/^data: \//{sub(/^data: /,""); sub(/\r$/,""); print; exit}' "$stream")
+
+    for request in \
+        '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}' \
+        '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
+        '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'; do
+        if ! curl -sf -o /dev/null -X POST "$base$endpoint" -H 'Content-Type: application/json' -d "$request"; then
+            echo "The MCP server rejected a request: $request"
+            exit 1
+        fi
+    done
+
+    if ! timeout 20 bash -c "until grep -q '\"id\":2' '$stream'; do sleep 0.5; done"; then
+        echo "The MCP server did not answer tools/list."
+        exit 1
+    fi
+
+    responses=$(awk '/^data: \{/{sub(/^data: /,""); print}' "$stream")
+    kill "$stream_pid" 2>/dev/null || true
+    server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' <<< "$responses")
+    tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' <<< "$responses")
     assert_string_value "$server_name" "recce"
     if [ "${tool_count:-0}" -lt 1 ]; then
         echo "The MCP server started but advertised no tools."
-        cat "$MCP_STDERR"
         exit 1
     fi
     echo "MCP server is up and advertised $tool_count tools."
+
+    echo "Stopping the MCP server..."
+    kill $(jobs -p) 2>/dev/null || true
+    wait || true
+    echo "MCP server stopped."
 }
 
 if [ "$SMOKE_SERVER" = "mcp-server" ]; then
     echo "Starting the MCP server..."
-    timeout 60 recce mcp-server < "$MCP_REQUESTS" > "$MCP_RESPONSES" 2> "$MCP_STDERR" &
-    check_mcp_server_status $!
+    recce mcp-server --sse --port "$MCP_PORT" &
+    check_mcp_server_status
 else
     echo "Starting the server..."
     recce server &
--- a/integration_tests/dbt/smoke_test.sh
+++ b/integration_tests/dbt/smoke_test.sh
@@ -7,19 +7,20 @@ pwd
 
 # Which server surface to smoke test: "server" (default) or "mcp-server".
 SMOKE_SERVER="${SMOKE_SERVER:-server}"
-# Only used when SMOKE_SERVER=mcp-server. The mcp SDK majors register tool
-# handlers differently, so the version under test has to be explicit.
-SMOKE_MCP_VERSION="${SMOKE_MCP_VERSION:-2.0.0}"
+# Only used when SMOKE_SERVER=mcp-server. Major.minor, e.g. "1.29" or "2.0":
+# the mcp SDK majors register tool handlers differently, so the version under
+# test has to be explicit, but `~=` still picks up the latest patch release.
+SMOKE_MCP_VERSION="${SMOKE_MCP_VERSION:-2.0}"
 
 case "$SMOKE_SERVER" in
     server) ;;
     mcp-server)
         # `mcp` is an optional extra, so CI's install does not carry it.
-        echo "Installing mcp==$SMOKE_MCP_VERSION"
+        echo "Installing mcp~=$SMOKE_MCP_VERSION"
         if command -v uv > /dev/null; then
-            uv pip install "mcp==$SMOKE_MCP_VERSION"
+            uv pip install "mcp~=$SMOKE_MCP_VERSION"
         else
-            python -m pip install "mcp==$SMOKE_MCP_VERSION"
+            python -m pip install "mcp~=$SMOKE_MCP_VERSION"
         fi
         ;;
     *)
@@ -128,37 +129,46 @@ function check_server_status() {
 }
 
 # Recce MCP Server
-# Stdio, not HTTP: there is no port to poll, so speak the protocol instead. The
-# startup path is what breaks (tool registration differs between mcp 1.x and
-# 2.0), so the handshake plus a non-empty tool list is the whole check.
-function check_mcp_server_status() {
-    echo "Starting the MCP server..."
-    local stderr_log output server_name tool_count
-    stderr_log=$(mktemp)
+# Stdio, not HTTP: there is no port to poll, so the server is started with the
+# handshake on stdin and its replies land in a file. The startup path is what
+# breaks (tool registration differs between mcp 1.x and 2.0), so the handshake
+# plus a non-empty tool list is the whole check.
+MCP_REQUESTS=$(mktemp)
+MCP_RESPONSES=$(mktemp)
+MCP_STDERR=$(mktemp)
+cat > "$MCP_REQUESTS" << 'EOF'
+{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}
+{"jsonrpc":"2.0","method":"notifications/initialized"}
+{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}
+EOF
 
-    if ! output=$( {
-        echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}'
-        echo '{"jsonrpc":"2.0","method":"notifications/initialized"}'
-        echo '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
-    } | timeout 60 recce mcp-server 2>"$stderr_log" ); then
-        echo "The MCP server failed to start:"
-        cat "$stderr_log"
+# Takes the pid of the backgrounded `recce mcp-server`. The server exits once it
+# reaches the end of the request file, so waiting on it is the readiness signal.
+function check_mcp_server_status() {
+    local mcp_pid="$1"
+    echo "Waiting for the MCP server to respond..."
+    if ! wait "$mcp_pid"; then
+        echo "Failed to start the MCP server."
+        cat "$MCP_STDERR"
         exit 1
     fi
 
-    server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' <<< "$output")
-    tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' <<< "$output")
+    local server_name tool_count
+    server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' "$MCP_RESPONSES")
+    tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' "$MCP_RESPONSES")
     assert_string_value "$server_name" "recce"
     if [ "${tool_count:-0}" -lt 1 ]; then
         echo "The MCP server started but advertised no tools."
-        cat "$stderr_log"
+        cat "$MCP_STDERR"
         exit 1
     fi
     echo "MCP server is up and advertised $tool_count tools."
 }
 
 if [ "$SMOKE_SERVER" = "mcp-server" ]; then
-    check_mcp_server_status
+    echo "Starting the MCP server..."
+    timeout 60 recce mcp-server < "$MCP_REQUESTS" > "$MCP_RESPONSES" 2> "$MCP_STDERR" &
+    check_mcp_server_status $!
 else
     echo "Starting the server..."
     recce server &
--- a/integration_tests/dbt/smoke_test.sh
+++ b/integration_tests/dbt/smoke_test.sh
@@ -199,7 +199,10 @@ function check_mcp_server_status() {
     fi
     echo "MCP server is up and advertised $tool_count tools."
 
-    if ! jq -e 'select(.id == 3) | .result.isError != true' > /dev/null <<< "$responses"; then
+    # `.result != null` first: on a JSON-RPC error response there is no `result` at all,
+    # and `null.isError != true` is true — which is exactly the escaped-exception
+    # regression this call is here to catch.
+    if ! jq -e 'select(.id == 3) | .result != null and .result.isError != true' > /dev/null <<< "$responses"; then
         echo "The MCP server failed a get_server_info tool call."
         exit 1
     fi
--- a/integration_tests/dbt/smoke_test.sh
+++ b/integration_tests/dbt/smoke_test.sh
@@ -5,6 +5,29 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$SCRIPT_DIR"
 pwd
 
+# Which server surface to smoke test: "server" (default) or "mcp-server".
+SMOKE_SERVER="${SMOKE_SERVER:-server}"
+# Only used when SMOKE_SERVER=mcp-server. The mcp SDK majors register tool
+# handlers differently, so the version under test has to be explicit.
+SMOKE_MCP_VERSION="${SMOKE_MCP_VERSION:-2.0.0}"
+
+case "$SMOKE_SERVER" in
+    server) ;;
+    mcp-server)
+        # `mcp` is an optional extra, so CI's install does not carry it.
+        echo "Installing mcp==$SMOKE_MCP_VERSION"
+        if command -v uv > /dev/null; then
+            uv pip install "mcp==$SMOKE_MCP_VERSION"
+        else
+            python -m pip install "mcp==$SMOKE_MCP_VERSION"
+        fi
+        ;;
+    *)
+        echo "Unknown SMOKE_SERVER '$SMOKE_SERVER' (expected 'server' or 'mcp-server')."
+        exit 1
+        ;;
+esac
+
 # Prepare env
 git restore models/customers.sql
 dbt --version
@@ -104,10 +127,44 @@ function check_server_status() {
     echo "Server stopped."
 }
 
-echo "Starting the server..."
-recce server &
-check_server_status false
+# Recce MCP Server
+# Stdio, not HTTP: there is no port to poll, so speak the protocol instead. The
+# startup path is what breaks (tool registration differs between mcp 1.x and
+# 2.0), so the handshake plus a non-empty tool list is the whole check.
+function check_mcp_server_status() {
+    echo "Starting the MCP server..."
+    local stderr_log output server_name tool_count
+    stderr_log=$(mktemp)
 
-echo "Starting the server (review mode)..."
-recce server --review recce_state.json &
-check_server_status true
+    if ! output=$( {
+        echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}'
+        echo '{"jsonrpc":"2.0","method":"notifications/initialized"}'
+        echo '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
+    } | timeout 60 recce mcp-server 2>"$stderr_log" ); then
+        echo "The MCP server failed to start:"
+        cat "$stderr_log"
+        exit 1
+    fi
+
+    server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' <<< "$output")
+    tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' <<< "$output")
+    assert_string_value "$server_name" "recce"
+    if [ "${tool_count:-0}" -lt 1 ]; then
+        echo "The MCP server started but advertised no tools."
+        cat "$stderr_log"
+        exit 1
+    fi
+    echo "MCP server is up and advertised $tool_count tools."
+}
+
+if [ "$SMOKE_SERVER" = "mcp-server" ]; then
+    check_mcp_server_status
+else
+    echo "Starting the server..."
+    recce server &
+    check_server_status false
+
+    echo "Starting the server (review mode)..."
+    recce server --review recce_state.json &
+    check_server_status true
+fi
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-mcp = ["mcp~=1.23"]
+mcp = ["mcp>=1.23,<3"]
 dev = [
     "pytest>=4.6",
     "pytest-asyncio>=0.23,<2.0",
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -726,15 +726,15 @@ class RecceMCPServer:
         return ListToolsResult(tools=await self._list_tools())
 
     async def _get_tool_input_schema(self, name: str) -> Optional[Dict[str, Any]]:
-        """Look up a tool's JSON schema, refreshing the cache on a miss.
+        """Look up a tool's JSON schema, building the cache on first use.
 
-        Mirrors the `_tool_cache` the mcp 1.x SDK keeps: `set_backend` can change the
-        advertised surface at runtime, so a miss means "re-read the list", not "unknown
-        tool". The cache is what keeps validation quiet — `_list_tools` logs its result
-        and writes an MCPLogger entry, so rebuilding it per call would forge a
-        `Returning N tools` line on every `tools/call`.
+        A miss must not mean "re-read the list": an unknown name misses every time, so
+        that would rebuild on every bad call and forge a `Returning N tools` line each
+        time — `_list_tools` logs its result and writes an MCPLogger entry. `set_backend`
+        is the only thing that moves the advertised surface after `__init__`, and it
+        clears this cache itself, so building once is enough.
         """
-        if name not in self._tool_schema_cache:
+        if not self._tool_schema_cache:
             self._tool_schema_cache = {tool.name: _tool_input_schema(tool) for tool in await self._list_tools()}
         return self._tool_schema_cache.get(name)
 
@@ -2563,6 +2563,10 @@ class RecceMCPServer:
             raise ValueError(f"Invalid mode '{mode}'. Use 'local' or 'cloud'.")
 
         async with self._backend_lock:
+            # Both branches below change what _list_tools advertises (backend, context,
+            # single_env), so the schemas validation reads have to be re-derived.
+            self._tool_schema_cache = {}
+
             if mode == "cloud":
                 session_id = arguments.get("session_id")
                 if not session_id:
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -19,7 +19,7 @@ from urllib.parse import quote
 import requests
 from mcp.server import Server
 from mcp.server.stdio import stdio_server
-from mcp.types import TextContent, Tool
+from mcp.types import CallToolResult, ListToolsResult, TextContent, Tool
 
 # DRC-3634: submit_run is the run-persistence entry point shared with the
 # recce server's run_router.  Local-mode ad-hoc diff tools import it to
@@ -42,6 +42,11 @@ from recce.tasks.valuediff import ValueDiffDetailTask
 from recce.util.recce_cloud import RECCE_CLOUD_API_HOST, RecceCloudException
 
 logger = logging.getLogger(__name__)
+
+# mcp 2.0 dropped the `@server.list_tools()` / `@server.call_tool()` decorators in
+# favour of constructor handlers with a `(ctx, params) -> Result` signature.
+# ponytail: one flag, two registration paths; drop the 1.x branch when the floor is mcp>=2.
+MCP_V2 = not hasattr(Server, "list_tools")
 
 try:
     from sentry_sdk import metrics as sentry_metrics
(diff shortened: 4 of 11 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -49,3 +49,33 @@ jobs:
...
           ./integration_tests/dbt/smoke_test.sh
+
+  mcp-smoke-test:
+    if: github.actor != 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    strategy:
+      # `mcp` is an optional extra pinned to >=1.23,<3, and the two majors
+      # register tool handlers differently. Both have to boot.
+      matrix:
+        mcp-version: ["1.28.1", "2.0.0"]
```
kentwelcome: I don't want to assign the full version number. Only the major and minor version. Ex: 1.29, 2.0, etc...

integration_tests/dbt/smoke_test.sh
```diff
@@ -5,6 +5,29 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$SCRIPT_DIR"
 pwd
 
+# Which server surface to smoke test: "server" (default) or "mcp-server".
+SMOKE_SERVER="${SMOKE_SERVER:-server}"
+# Only used when SMOKE_SERVER=mcp-server. The mcp SDK majors register tool
+# handlers differently, so the version under test has to be explicit.
+SMOKE_MCP_VERSION="${SMOKE_MCP_VERSION:-2.0.0}"
```
kentwelcome: The default value should be `2.0`

integration_tests/dbt/smoke_test.sh
```diff
@@ -5,6 +5,29 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+
+case "$SMOKE_SERVER" in
+    server) ;;
+    mcp-server)
+        # `mcp` is an optional extra, so CI's install does not carry it.
+        echo "Installing mcp==$SMOKE_MCP_VERSION"
+        if command -v uv > /dev/null; then
+            uv pip install "mcp==$SMOKE_MCP_VERSION"
```
kentwelcome: Don't use `==` to assign a full version number. We should use `~=` to assign a version. We should use the latest version when a minor version is released.

integration_tests/dbt/smoke_test.sh
```diff
@@ -5,6 +5,29 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+        else
+            python -m pip install "mcp==$SMOKE_MCP_VERSION"
```
kentwelcome: Same as the previous review comment.

integration_tests/dbt/smoke_test.sh
```diff
@@ -104,10 +127,44 @@ function check_server_status() {
...
+    if [ "${tool_count:-0}" -lt 1 ]; then
+        echo "The MCP server started but advertised no tools."
+        cat "$stderr_log"
+        exit 1
+    fi
+    echo "MCP server is up and advertised $tool_count tools."
+}
+
+if [ "$SMOKE_SERVER" = "mcp-server" ]; then
+    check_mcp_server_status
```
kentwelcome, line 230: No mcp-server be launched before check the mcp server status. 
Launch mcp server before check the status.
`recce mcp-server&`

integration_tests/dbt/smoke_test.sh
```diff
@@ -104,10 +127,44 @@ function check_server_status() {
...
+function check_mcp_server_status() {
+    echo "Starting the MCP server..."
+    local stderr_log output server_name tool_count
+    stderr_log=$(mktemp)
+
+    if ! output=$( {
+        echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}'
+        echo '{"jsonrpc":"2.0","method":"notifications/initialized"}'
+        echo '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
+    } | timeout 60 recce mcp-server 2>"$stderr_log" ); then
```
kentwelcome: Please follow the logic of check_server_status. The `recce mcp-server` should be executed outside this function

.github/workflows/integration-tests.yaml
kentwelcome: Done in 30d49a4a — the matrix is now \`mcp-version: ["1.29", "2.0"]\`, and the script installs with \`~=\` so the latest patch of that minor is used.

integration_tests/dbt/smoke_test.sh
kentwelcome: Done in 30d49a4a — `SMOKE_MCP_VERSION` now defaults to `2.0`.

integration_tests/dbt/smoke_test.sh
kentwelcome: Done in 30d49a4a — `uv pip install "mcp~=$SMOKE_MCP_VERSION"`. With the major.minor value that resolves to `>=2.0,<3`, so a new patch or minor release is picked up automatically.

integration_tests/dbt/smoke_test.sh
kentwelcome: Done in 30d49a4a — the `python -m pip` fallback uses `~=` as well.

integration_tests/dbt/smoke_test.sh
kentwelcome: Restructured in 30d49a4a to follow `check_server_status`:

```bash
echo "Starting the MCP server..."
timeout 60 recce mcp-server < "$MCP_REQUESTS" > "$MCP_RESPONSES" 2> "$MCP_STDERR" &
check_mcp_server_status $!
```

The server is stdio, so it needs its requests on stdin to have anything to do. The three handshake lines now live in `$MCP_REQUESTS`, written once above the function, which lets the server be backgrounded exactly like `recce server &`. The function only waits on the pid and inspects the responses — no launching inside it.

Verified against a real `recce mcp-server` on mcp 1.29 and on 2.0 (both report 20 tools), and against a stub reproducing the original startup error, which fails with exit 1 and prints the captured stderr.

integration_tests/dbt/smoke_test.sh
kentwelcome: Fixed in 30d49a4a — `recce mcp-server` is now started (backgrounded, with the handshake fed on stdin) before `check_mcp_server_status` runs, same shape as `recce server &` + `check_server_status`. See the reply on the other comment for the full snippet.

recce/mcp_server.py
```diff
@@ -678,6 +682,47 @@ def _build_instructions(self) -> Optional[str]:
...
+        1.x turned a raised exception into `isError=True`; 2.0 turns it into a
+        JSON-RPC protocol error instead, which an agent reads as a transport
+        failure rather than a tool failure. Return the tool error explicitly to
+        keep the response identical across both versions.
+        """
+        try:
+            content = await self._call_tool(params.name, params.arguments or {})
+        except Exception as e:
+            return CallToolResult(content=[TextContent(type="text", text=str(e))], isError=True)
+        return CallToolResult(content=content)
```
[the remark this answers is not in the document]
kentwelcome: Checked this against both SDKs, and `isError` is the correct spelling for both — no per-version branching needed.

- mcp 2.0: the field is `is_error` with **alias** `isError`. Pydantic validates by alias, so `CallToolResult(..., isError=True)` sets it, and `model_dump(by_alias=True)` emits `{"isError": true}`.
- mcp 1.29: the field is literally `isError`.

Verified just now on a clean mcp 2.0.0 install:

```
field name: is_error | alias: isError
constructed with isError= -> True
wire dump: {'isError': True}
```

`isError` is also the field name fixed by the MCP wire protocol, so it cannot drift without a spec change — whereas the Python attribute name did drift between majors. Choosing the name by inspecting `model_fields` would add a branch that resolves to the same value on both, and would break if a future SDK renames the attribute again while keeping the alias.

The end-to-end coverage is in `tests/mcp_compat.py`, which normalises the read side (`is_error` / `isError`) so the assertions run unchanged on both majors; a real `ClientSession` handshake on 2.0 returns `isError: True` for a failing tool.

recce/mcp_server.py
```diff
@@ -43,6 +43,11 @@
 
 logger = logging.getLogger(__name__)
 
+# mcp 2.0 dropped the `@server.list_tools()` / `@server.call_tool()` decorators in
+# favour of constructor handlers with a `(ctx, params) -> Result` signature.
+# ponytail: one flag, two registration paths; drop the 1.x branch when the floor is mcp>=2.
+MCP_V2 = not hasattr(Server, "list_tools")
```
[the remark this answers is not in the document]
kentwelcome: Good catch — that was a stray marker from my tooling with no meaning to a reader. Rewritten in 30d49a4a: "This flag selects the registration path; the 1.x branch can go once the floor is mcp>=2."

.github/workflows/integration-tests.yaml
```diff
@@ -49,3 +49,33 @@
...
+          uv sync --no-dev --python 3.12
+          uv pip install dbt-core dbt-duckdb
+
+      - name: Run smoke test - mcp server
+        env:
+          SMOKE_SERVER: mcp-server
+          SMOKE_MCP_VERSION: ${{ matrix.mcp-version }}
+        run: |
+          source .venv/bin/activate
+          ./integration_tests/dbt/smoke_test.sh
```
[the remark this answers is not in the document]
kentwelcome: Fixed in 30d49a4a — added a top-level `permissions: contents: read` block to the workflow. Both jobs only check out the repo and run tests, so read is sufficient.

integration_tests/dbt/smoke_test.sh
kentwelcome: Reworked in bad52311 — the stdin approach is gone. The server is now started as a background HTTP server and every request goes over the MCP port from inside `check_mcp_server_status`:

```bash
echo "Starting the MCP server..."
recce mcp-server --sse --port "$MCP_PORT" &
check_mcp_server_status
```

Inside the function: poll `/health` until it answers (same `timeout ... until curl` shape as `check_server_status`), open the `GET /sse` response stream, POST `initialize` / `notifications/initialized` / `tools/list` to the session endpoint that stream hands out, then assert `serverInfo.name == "recce"` and a non-empty tool list. It stops the server on the way out, like `check_server_status` does.

Verified against a real `recce mcp-server --sse` on mcp 1.29 and on 2.0 — both report 20 tools — and against a stub that reproduces the original startup failure, where `/health` never answers and the check exits 1.

integration_tests/dbt/smoke_test.sh
kentwelcome: Updated in bad52311 — `recce mcp-server --sse --port "$MCP_PORT" &` is launched before the check, and the requests/response assertions all live inside `check_mcp_server_status` now. See the reply on the other thread for the details.

recce/mcp_server.py
```diff
@@ -678,6 +682,47 @@ def _build_instructions(self) -> Optional[str]:
...
+    async def _handle_call_tool(self, ctx, params) -> CallToolResult:
+        """`tools/call` in the mcp 2.0 handler signature (also used by tests on 1.x).
+
+        1.x turned a raised exception into `isError=True`; 2.0 turns it into a
+        JSON-RPC protocol error instead, which an agent reads as a transport
+        failure rather than a tool failure. Return the tool error explicitly to
+        keep the response identical across both versions.
+        """
+        try:
+            content = await self._call_tool(params.name, params.arguments or {})
```
iamcxa: **HIGH — mcp 2.0 does no input-schema validation; 1.x does, by default.**

`arguments` goes straight to the handler here. On mcp 1.x the SDK validated first — `mcp/server/lowlevel/server.py` in 1.29.0:

```python
def call_tool(self, *, validate_input: bool = True):
    ...
    if validate_input and tool:
        try:
            jsonschema.validate(instance=arguments, schema=tool.inputSchema)
        except jsonschema.ValidationError as e:
            return self._make_error_result(f"Input validation error: {e.message}")
```

The 1.x branch of `_build_server` registers with that default, so it keeps validating. In mcp 2.0 the low-level server imports `jsonschema` nowhere at all (it survives only in `mcp/client/session.py`, for client-side *output* schemas). Nothing replaces it here, so `keep the response identical across both versions` does not hold for invalid arguments.

It fails silently rather than loudly. Using this repo's own schema — `impact_analysis` declares `"skip_value_diff": {"type": "boolean"}` (`recce/mcp_server.py:1533`) and the handler reads it at `:2009` / `:2161`:

```python
skip_value_diff = arguments.get("skip_value_diff", False)
...
if not skip_value_diff:
```

An agent sending `{"skip_value_diff": "false"}` gets, on **1.x**, a clean `isError=True, "Input validation error: 'false' is not of type 'boolean'"`. On **2.0** the string is truthy, the value comparison is **silently skipped**, and the agent receives a result that looks complete. Same class applies to missing `required` fields and to array-typed params receiving a bare string.

Validating in the adapter restores parity and adds no dependency — `jsonschema` is already required by mcp on both majors (and by dbt-core):

```python
async def _handle_call_tool(self, ctx, params) -> CallToolResult:
    arguments = params.arguments or {}
    tool = next((t for t in await self._list_tools() if t.name == params.name), None)
    if tool is not None:
        # `Tool.inputSchema` on 1.x, `Tool.input_schema` on 2.0 — this adapter is
        # also driven by the tests on 1.x, so read it version-agnostically.
        schema = getattr(tool, "input_schema", None)
        if schema is None:
            schema = tool.inputSchema
        try:
            jsonschema.validate(instance=arguments, schema=schema)
        except jsonschema.ValidationError as e:
            return CallToolResult(
                content=[TextContent(type="text", text=f"Input validation error: {e.message}")],
                isError=True,
            )
    try:
        content = await self._call_tool(params.name, arguments)
    ...
```

1.x caches the tool definitions (`_tool_cache`) rather than rebuilding them per call; worth mirroring if `_list_tools()` turns out to be hot, but correctness first.

tests/mcp_compat.py
```diff
@@ -0,0 +1,38 @@
...
+    isError: bool
+
+
+async def invoke_call_tool(server, name: str, arguments: Optional[Dict[str, Any]] = None) -> ToolResult:
+    """Call a tool and normalise the result, including the error case."""
+    result = await server._handle_call_tool(None, CallToolRequestParams(name=name, arguments=arguments or {}))
+    is_error = getattr(result, "is_error", None)
+    if is_error is None:
+        is_error = getattr(result, "isError", None)
+    return ToolResult(content=list(result.content), isError=bool(is_error))
```
iamcxa: **HIGH — `tests/test_mcp_e2e.py` was not migrated; 5 tests fail on mcp 2.0.**

This module normalises the error flag for `test_mcp_server.py` and `test_mcp_cloud_backend.py`, but `tests/test_mcp_e2e.py` still reads `result.isError` off a real SDK `CallToolResult` at lines 1265, 1273, 1283, 1293, 1303, 1317, 1326. Reproduced against this head in a clean venv with `mcp==2.0.0`:

```
AttributeError: 'CallToolResult' object has no attribute 'isError'. Did you mean: 'is_error'?
5 failed, 55 passed
```

All five failures share that one cause. Substituting `.isError` -> `.is_error` in a scratch copy gives **60 passed** — which incidentally confirms the server-side 2.0 error contract works end to end through a real `ClientSession`.

The environment problem noted in the PR body (duckdb rejecting `python_scan_all_frames` during fixture setup) was real, and it masked this: the fixtures died before any assertion ran, so the file looked merely un-runnable rather than incompatible. Once fixtures can run, it fails for a *related* reason.

A bare `.is_error` would break 1.x, so the fix belongs here — the same normalisation `invoke_call_tool` already performs:

```python
def is_error(result) -> bool:
    """Read a ``CallToolResult`` error flag regardless of SDK field naming."""
    value = getattr(result, "is_error", None)
    if value is None:
        value = getattr(result, "isError", None)
    return bool(value)
```

pyproject.toml
```diff
@@ -46,7 +46,7 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-mcp = ["mcp~=1.23"]
+mcp = ["mcp>=1.23,<3"]
```
iamcxa, line 49: **MEDIUM — the widened pin is not propagated to the lockfile or to the test matrix.**

**a. `uv.lock` is stale.** It still records `specifier = "~=1.23"` for this extra (`uv.lock:2561`) and pins `mcp 1.28.1` (`uv.lock:1251`). `uv lock --check` at this head reports:

```
The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
```

CI stays green only because no workflow uses `--locked` / `--frozen`, so every `uv sync` silently re-resolves and rewrites the lock in the working tree. `uv lock` + commit.

**b. The MCP test suite never runs on 2.x.** `tests-python.yaml` -> `Test Python Versions` uses the tox `[testenv:{3.10,3.11,3.12,3.13}]` env, which carries no `mcp` dep, so the three MCP modules are skipped at collection (`collected 1334 items / 3 skipped`). Only `[testenv:dbtlatest]` has `mcp`, pinned `mcp>=1.0.0` — below this project's own floor, with no upper bound — and on this run it resolved to **`mcp==1.27.1`** (visible in the `Test DBT Versions` job log). So the whole MCP suite runs exactly once in CI, on 1.27.1. That is why the `test_mcp_e2e.py` breakage is invisible.

tox's native extras key inherits the declared window instead of drifting from it:

```ini
[testenv:dbt{1.6,1.7,1.8,1.9,latest}]
extras =
    dbtlatest: mcp
```

An explicit 2.x leg that runs `tests/test_mcp*.py` is what actually closes the gap.

integration_tests/dbt/smoke_test.sh
```diff
@@ -104,10 +128,79 @@ function check_server_status() {
...
+        echo "The MCP server did not hand out a session endpoint."
+        exit 1
+    fi
+    # SSE lines are CRLF-terminated; a trailing CR makes the POST url malformed.
+    endpoint=$(awk '/^data: \//{sub(/^data: /,""); sub(/\r$/,""); print; exit}' "$stream")
+
+    for request in \
+        '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}' \
+        '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
+        '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'; do
```
iamcxa: **MEDIUM — the smoke test never calls a tool.**

The handshake covers `initialize` + `tools/list`. `tools/call` is never sent, so `_handle_call_tool` — the adapter this PR adds specifically to preserve the error contract — has no automated coverage on **either** major.

This is also now the only end-to-end path left on 1.x: the migrated unit tests call `_handle_call_tool` directly, which on 1.x is *not* the registered handler, so nothing exercises the decorator path any more.

Two more POSTs in this same loop close it on both majors over the real wire: one cheap successful call (`get_server_info`), and one deliberately failing call (a nonexistent tool name) asserting `"isError":true` in the SSE response. If the failing call used a schema-invalid argument instead, it would also have caught the validation gap flagged on `recce/mcp_server.py`.

integration_tests/dbt/smoke_test.sh
```diff
@@ -104,10 +128,79 @@ function check_server_status() {
...
+    local base="http://localhost:$MCP_PORT"
+    local stream stream_pid endpoint responses server_name tool_count
+    stream=$(mktemp)
+
+    echo "Waiting for the MCP server to respond..."
+    if ! timeout 60 bash -c "until curl -sf $base/health > /dev/null; do
+    echo \"MCP server not ready yet...\"
+    sleep 2
+    done"; then
+        echo "Failed to start the MCP server within the time limit."
```
iamcxa, line 149: **LOW — failure paths leave the server and the SSE stream running.**

Every `exit 1` in this function (lines 148, 157, 168, 174, 184) fires after `recce mcp-server` is already backgrounded, and after line 153 also leaves `curl -sN` alive. Both inherit the CI step's stdout, so a failed smoke test can keep the step open past the point where it has anything left to say.

`check_server_status` has the same shape, so this is not a regression — but the new function orphans two processes rather than one, and a `trap 'kill $(jobs -p) 2>/dev/null || true' EXIT` covers both cheaply.

integration_tests/dbt/smoke_test.sh
```diff
@@ -104,10 +128,79 @@ function check_server_status() {
...
+        echo "The MCP server did not answer tools/list."
+        exit 1
+    fi
+
+    responses=$(awk '/^data: \{/{sub(/^data: /,""); print}' "$stream")
+    kill "$stream_pid" 2>/dev/null || true
+    server_name=$(jq -r 'select(.id == 1) | .result.serverInfo.name' <<< "$responses")
+    tool_count=$(jq -r 'select(.id == 2) | .result.tools | length' <<< "$responses")
+    assert_string_value "$server_name" "recce"
+    if [ "${tool_count:-0}" -lt 1 ]; then
```
iamcxa, line 196: **LOW — `tool_count -lt 1` is a weaker gate than the risk it guards.**

The failure mode this job exists to catch is *registration*, and in server mode `list_tools` advertises 20 tools. `>= 1` still passes if 19 of them silently stop registering. Asserting a known name is one more `jq` line and does not get brittle as tools are added:

```bash
jq -e 'select(.id == 2) | [.result.tools[].name] | index("lineage_diff")' <<< "$responses" > /dev/null
```

integration_tests/dbt/smoke_test.sh
```diff
@@ -5,6 +5,30 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$SCRIPT_DIR"
 pwd
 
+# Which server surface to smoke test: "server" (default) or "mcp-server".
+SMOKE_SERVER="${SMOKE_SERVER:-server}"
+# Only used when SMOKE_SERVER=mcp-server. Major.minor, e.g. "1.29" or "2.0":
+# the mcp SDK majors register tool handlers differently, so the version under
+# test has to be explicit, but `~=` still picks up the latest patch release.
```
iamcxa: **NIT — the comment describes patch releases; PEP 440 gives minor too.**

`mcp~=1.29` expands to `>= 1.29, == 1.*`, i.e. `>=1.29,<2.0` — a future `1.30` is picked up, not just patches. Your own reply on the earlier `~=` thread states this correctly ("a new patch or minor release is picked up automatically"); only the in-code comment is off.

tests/mcp_compat.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+async def invoke_list_tools(server) -> List[Tool]:
+    """Return the advertised tools."""
+    return (await server._handle_list_tools(None, None)).tools
+
+
+def input_schema(tool: Tool) -> Dict[str, Any]:
+    """Read a tool's JSON schema regardless of SDK field naming."""
+    return getattr(tool, "input_schema", None) or tool.inputSchema
```
iamcxa: **NIT — test for `None`, not falsiness.**

`getattr(tool, "input_schema", None) or tool.inputSchema` falls through to `tool.inputSchema` when the schema is an empty dict, which raises `AttributeError` on 2.0 (no such attribute there). Unreachable today — every tool declares a non-empty schema — but it is the same `is None` guard this codebase already documents elsewhere, and `invoke_call_tool` above already uses that shape.

iamcxa (changes requested):
@kentwelcome — reviewed at head `c5c3395d`. **Event: REQUEST_CHANGES** (two blocking items; everything below them is advisory).

The compat rewrite is the right shape and it genuinely boots on both majors — I verified
that independently, not just from the green CI. Two things do not survive the widened pin:
mcp 2.0's low-level server **dropped input-schema validation** that 1.x performs by default,
so `tools/call` behaviour is not actually identical across the two; and
`tests/test_mcp_e2e.py` was never migrated to the new compat helpers, so it **fails on
mcp 2.0** — invisibly, because no CI job runs the Python tests against 2.x.

## PR intent

| | |
|---|---|
| **Goal (PR body / release note)** | `recce mcp-server` starts and serves tools on both mcp 1.x and mcp 2.0 |
| **Verdict** | ✅ **Achieved** for start-up and `tools/list` — verified on both majors |
| **Gap** | The stronger claim — *"keep the response identical across both versions"* — does not hold for `tools/call` with invalid arguments (finding 1) |
| **Gap** | The sub-claim *"Tests move to `tests/mcp_compat.py` helpers"* is incomplete: `tests/test_mcp_e2e.py` still reads `result.isError` (finding 2) |

## Verification performed

Independent of CI, in clean venvs (Python 3.12; one with `mcp==2.0.0`, one with `mcp==1.29.0`,
recce installed from this head):

| Check | Result |
|---|---|
| `pytest tests/test_mcp_server.py tests/test_mcp_cloud_backend.py` on mcp 2.0.0 | **194 passed** |
| `pytest tests/test_mcp_e2e.py` on mcp 2.0.0 | **5 failed**, 55 passed → finding 2 |
| same file with only `.isError` → `.is_error` | **60 passed** — the production code is fine; the defect is test-side |
| `pip check` in the clean mcp-2.0 env (you asked for this) | **clean**, 110 packages compatible |
| `uv lock --check` at this head | **fails** — "The lockfile at `uv.lock` needs to be updated" → finding 3 |
| `CallToolResult(isError=True)` on 2.0 | correct — field `is_error`, alias `isError`, `validate_by_alias=True`. Confirms your reply to Copilot; not re-raising it |
| `Server.__init__` on 2.0 | accepts `on_list_tools` / `on_call_tool`; `hasattr(Server, "list_tools")` is `False`, so `MCP_V2` detection is right |
| `Tool(inputSchema=...)` on 2.0 | still valid via the alias — the 20 unchanged construction sites are safe |
| `stdio_server`, `SseServerTransport`, `Server.run(...)`, `create_initialization_options()` on 2.0 | all present |
| Stale refs to `_setup_handlers` / `request_handlers` | none remain anywhere |
| flake8 / isort on changed files | clean (black deferred to the pinned pre-commit 26.3.1; my local 25.9.0 only wants to reformat pre-existing `textwrap.dedent` blocks this PR does not touch) |
| `permissions: contents: read` | addresses the CodeQL finding |

`httpx2 2.10.0` (mcp 2.0's new transitive dep) declares `github.com/pydantic/httpx2` as its
home. That is package metadata, not an independent provenance audit — worth a human look
before anyone raises the floor to `mcp>=2`.

## Break-point coverage

- **Break point**: `recce/mcp_server.py:670` (`self.server = self._build_server()`) — the original
  `'Server' object has no attribute 'list_tools'` crash site.
- **Failure chain**: `recce mcp-server` → `RecceMCPServer.__init__` → handler registration →
  `AttributeError` on 2.0 → server never starts.
- **Verified at runtime**: `mcp-smoke-test (1.29)` and `(2.0)` are both green on this head — a real
  `recce mcp-server --sse`, a real handshake, a non-empty tool list. Plus a real `ClientSession`
  round-trip on 2.0 locally.
- **Residual — stdio transport**: the smoke test drives only `--sse`. The default
  `recce mcp-server` (stdio) is exercised on neither major. The crash site is shared, so the
  reported regression *is* covered; a future stdio-specific divergence would not be.
- **Residual — `tools/call`**: not exercised over the wire on either major (finding 4).

---

## Blocking

### 1. HIGH — mcp 2.0 does no input-schema validation; 1.x does, by default

`_handle_call_tool` forwards `params.arguments or {}` straight to the handler. On mcp 1.x the
SDK validated first — `mcp/server/lowlevel/server.py` in 1.29.0:

```python
def call_tool(self, *, validate_input: bool = True):
    ...
    # input validation
    if validate_input and tool:
        try:
            jsonschema.validate(instance=arguments, schema=tool.inputSchema)
        except jsonschema.ValidationError as e:
            return self._make_error_result(f"Input validation error: {e.message}")
```

The 1.x branch of `_build_server` registers with that default, so it keeps validating. In mcp
2.0 the low-level server imports `jsonschema` nowhere at all (it survives only in
`mcp/client/session.py`, for client-side *output* schemas). Nothing replaces it here.

That is a silent divergence, not a louder error. Concrete case using this repo's own schema —
`impact_analysis` declares `"skip_value_diff": {"type": "boolean"}`
(`recce/mcp_server.py:1533`) and the handler reads it at `recce/mcp_server.py:2009` /
`:2161`:

```python
skip_value_diff = arguments.get("skip_value_diff", False)
...
if not skip_value_diff:
```

An agent sending `{"skip_value_diff": "false"}` gets, on **1.x**, a clean
`isError=True, "Input validation error: 'false' is not of type 'boolean'"`. On **2.0** the
string is truthy, the value comparison is **silently skipped**, and the agent receives a
result that looks complete. Same class applies to missing `required` fields and to
array-typed params receiving a bare string.

Validating in the adapter restores parity, and adds no dependency — `jsonschema` is already
required by mcp on both majors (and by dbt-core):

```python
async def _handle_call_tool(self, ctx, params) -> CallToolResult:
    arguments = params.arguments or {}
    tool = next((t for t in await self._list_tools() if t.name == params.name), None)
    if tool is not None:
        # `Tool.inputSchema` on 1.x, `Tool.input_schema` on 2.0 — this adapter is
        # also driven by the tests on 1.x, so read it version-agnostically.
        schema = getattr(tool, "input_schema", None)
        if schema is None:
            schema = tool.inputSchema
        try:
            jsonschema.validate(instance=arguments, schema=schema)
        except jsonschema.ValidationError as e:
            return CallToolResult(
                content=[TextContent(type="text", text=f"Input validation error: {e.message}")],
                isError=True,
            )
    try:
        content = await self._call_tool(params.name, arguments)
    ...
```

1.x caches the tool definitions (`_tool_cache`) rather than rebuilding them per call; worth
mirroring if `_list_tools()` turns out to be hot, but correctness first.

*(Raised by a cross-model reviewer; I confirmed it against the source of both SDKs, quoted above.)*

### 2. HIGH — `tests/test_mcp_e2e.py` was not migrated; 5 tests fail on mcp 2.0

The PR adds `tests/mcp_compat.py` and routes `test_mcp_server.py` /
`test_mcp_cloud_backend.py` through it, but `tests/test_mcp_e2e.py` still reads
`result.isError` off a real SDK `CallToolResult` at lines 1265, 1273, 1283, 1293, 1303, 1317,
1326. Reproduced against this head with `mcp==2.0.0`:

```
AttributeError: 'CallToolResult' object has no attribute 'isError'. Did you mean: 'is_error'?
5 failed, 55 passed
```

All five failures share that one cause. Substituting `.isError` → `.is_error` in a scratch copy
gives **60 passed** — which incidentally confirms the server-side 2.0 error contract works end
to end through a real `ClientSession`.

The environment problem you hit (duckdb rejecting `python_scan_all_frames` during fixture
setup) was real, and it masked this: the fixtures died before any assertion ran, so the file
looked merely un-runnable rather than incompatible. Once fixtures can run, it fails for a
*related* reason.

A bare `.is_error` would break 1.x, so the fix belongs in the compat module — the same
normalisation `invoke_call_tool` already performs:

```python
def is_error(result) -> bool:
    """Read a ``CallToolResult`` error flag regardless of SDK field naming."""
    value = getattr(result, "is_error", None)
    if value is None:
        value = getattr(result, "isError", None)
    return bool(value)
```

---

## Non-blocking

### 3. MEDIUM — the widened pin is not propagated to the lockfile or to the test matrix

Two consequences of `mcp = ["mcp>=1.23,<3"]` that did not follow it:

**a. `uv.lock` is stale.** It still records `specifier = "~=1.23"` for the extra (`uv.lock:2561`)
and pins `mcp 1.28.1` (`uv.lock:1251`). `uv lock --check` at this head reports *"The lockfile at
`uv.lock` needs to be updated"*. CI stays green only because no workflow uses `--locked` /
`--frozen`, so every `uv sync` silently re-resolves and rewrites the lock in the working tree.
`uv lock` + commit.

**b. The MCP test suite never runs on 2.x.** `tests-python.yaml` → `Test Python Versions` uses the
tox `[testenv:{3.10,3.11,3.12,3.13}]` env, which carries no `mcp` dep, so the three MCP modules
are skipped at collection (`collected 1334 items / 3 skipped`). Only `[testenv:dbtlatest]` has
`mcp`, pinned `mcp>=1.0.0` — below this project's own floor, with no upper bound — and on this
run it resolved to **`mcp==1.27.1`** (visible in the `Test DBT Versions` log). So the whole MCP
suite runs exactly once in CI, on 1.27.1. That is why finding 2 is invisible.

Using tox's native extras key inherits the declared window instead of drifting from it:

```ini
[testenv:dbt{1.6,1.7,1.8,1.9,latest}]
extras =
    dbtlatest: mcp
```

An explicit 2.x leg that runs `tests/test_mcp*.py` is what actually closes the gap.

### 4. MEDIUM — the smoke test never calls a tool

The handshake covers `initialize` + `tools/list`. `tools/call` is never sent, so
`_handle_call_tool` — the adapter this PR adds specifically to preserve the error contract —
has no automated coverage on **either** major. Note this is also now the only end-to-end path
on 1.x: the migrated unit tests call `_handle_call_tool` directly, which on 1.x is *not* the
registered handler, so nothing exercises the decorator path any more.

Two more POSTs in the same loop close it on both majors over the real wire: one cheap
successful call (`get_server_info`), and one deliberately failing call (a nonexistent tool name)
asserting `"isError":true` in the SSE response. That would also have caught finding 1, if the
failing call used a schema-invalid argument instead.

### 5. LOW — failure paths leave the server and the SSE stream running

Every `exit 1` inside `check_mcp_server_status` (lines 148, 157, 168, 174, 184) fires after
`recce mcp-server` is already backgrounded, and after line 153 also leaves `curl -sN` alive.
Both inherit the CI step's stdout, so a failed smoke test can keep the step open past the point
where it has anything left to say. `check_server_status` has the same shape, so this is not a
regression — but the new function orphans two processes rather than one, and a
`trap 'kill $(jobs -p) 2>/dev/null || true' EXIT` covers both cheaply.

### 6. LOW — `tool_count -lt 1` is a weaker gate than the risk it guards

The failure mode this job exists to catch is *registration*, and in server mode `list_tools`
advertises 20 tools. `>= 1` still passes if 19 of them silently stop registering. Asserting a
known name is one more `jq` line and does not get brittle as tools are added:

```bash
jq -e 'select(.id == 2) | [.result.tools[].name] | index("lineage_diff")' <<< "$responses" > /dev/null
```

### 7. NIT — the `~=` comment describes patch releases; PEP 440 gives minor too

`mcp~=1.29` expands to `>= 1.29, == 1.*`, i.e. `>=1.29,<2.0` — a future `1.30` is picked up, not
just patches. Your own reply on the earlier thread states this correctly ("a new patch or minor
release is picked up automatically"); only the in-code comment is off.

### 8. NIT — `input_schema()` should test for `None`, not falsiness

`getattr(tool, "input_schema", None) or tool.inputSchema` falls through to `tool.inputSchema`
when the schema is an empty dict, which raises `AttributeError` on 2.0 (no such attribute
there). Unreachable today — every tool declares a non-empty schema — but it is the same
`is None` guard this codebase already documents elsewhere.

---

## Cross-model reconciliation

A second reviewer on a different model ran against the same diff.

| Bucket | Count | Items |
|---|---|---|
| Agreement (both) | 1 | no real `tools/call` coverage on either registration path (finding 4) |
| Claude-only | 5 | findings 2, 3b, 6, 7, 8 |
| Cross-model-only | 3 | findings 1, 3a, 5 |
| Contradictions | 0 | — |

No arbitration was needed: each cross-model-only finding was verified directly against primary
sources (SDK source for finding 1, `uv lock --check` for 3a, the script's own control flow for
5) rather than adjudicated between models. Note that the highest-severity item in this review
came from the lane the primary agents did not cover — worth weighing against any instinct that
model agreement is what makes a finding real. The human with domain context is still the decider.

## Advisory (not posted as inline comments)

- **PR body is stale against the final implementation.** It still describes the stdio design
  (*"speaks stdio, so there is no port to poll — the check feeds it `initialize` / ..."*) that
  `bad52311` replaced with HTTP/SSE, and cites *"mcp 1.28.1 and 2.0.0"* where the matrix is
  `["1.29", "2.0"]`. The body is what the release note and future archaeology inherit.
- **`.claude/skills/recce-mcp-dev/SKILL.md:18`** — *"MCP SDK quirk — Handler must **raise** for
  SDK to set `isError=True`"* is now version-dependent: true on 1.x, false on 2.0 where
  `_handle_call_tool` converts explicitly. Worth updating in the same PR, along with a note about
  input validation if finding 1 is addressed.
- **`mcp-smoke-test` runs the full `recce run` / `recce summary` assertion block** before it
  reaches the MCP server. The dbt artifacts are genuinely needed; those assertions are not. As
  written, a summary-rendering regression turns the MCP job red for an unrelated reason.

recce/mcp_server.py
iamcxa: Fixed in c7f7987b.

`_handle_call_tool` now validates `arguments` against the tool's schema before dispatch and returns 1.x's exact `Input validation error: ...` wording. Schemas come from `_get_tool_input_schema`, cached and refreshed on a miss — mirroring the SDK's own `_tool_cache`, because `set_backend` can change the advertised surface at runtime and rebuilding the list per call would forge a `Returning N tools` log line on every `tools/call`. There is a test for exactly that log side effect.

New `TestCallToolInputValidation` covers wrong type, missing required field, a well-typed call not being blocked, and an unknown tool not being misreported as a validation failure. 5 tests, green on mcp 1.29.0 and 2.0.0.

The wire is covered too: the smoke test now sends `{"select":123}` to `lineage_diff` and asserts `isError` + the message (see the `smoke_test.sh` thread).

tests/mcp_compat.py
iamcxa: Fixed in c0c0302d.

The normalisation `invoke_call_tool` already did is now a free `is_error()` function, since it also has to take results built by a real `ClientSession`. All 7 sites in `tests/test_mcp_e2e.py` go through it.

`tests/test_mcp_server.py` + `tests/test_mcp_cloud_backend.py` + `tests/test_mcp_e2e.py`: **259 passed on mcp 2.0.0 and on 1.29.0**. Full suite on both: 1583 passed, 5 skipped, with the only failures being the 5 pre-existing `test_server.py` SPA routes that need a built frontend (they fail identically on `c5c3395d`).

CI can see this now — see the `pyproject.toml` thread.

pyproject.toml
iamcxa: Fixed in 74e33136 (lock) and 23258ffe (coverage).

**Lock**: `uv lock` regenerated. The diff is 4 lines — the recorded specifier becomes `>=1.23,<3`, plus two marker refinements from a newer resolver. The resolved mcp version stays 1.28.1, since it still satisfies the new window; moving that is a separate decision (`uv lock --upgrade-package mcp`). `uv lock --check` is clean.

**Coverage**: `mcp-smoke-test` already installs both majors, so it is the cheap place to run pytest rather than standing up a new matrix. A `Run MCP tests` step now runs the three MCP modules there, after the smoke step that installs the SDK. tox takes mcp through `extras = dbtlatest: mcp` instead of restating `mcp>=1.0.0`, so that env inherits the declared window; verified with `tox c` that `extras` resolves to `mcp` for `dbtlatest` and stays empty for `dbt1.8` / `3.12`.

integration_tests/dbt/smoke_test.sh
iamcxa: Fixed in 122ffde4.

Two calls added to the same loop: id 3 calls `get_server_info` and asserts it succeeds; id 4 sends `{"select":123}` to `lineage_diff`, whose schema says string, and asserts `isError` plus `Input validation error`. That second one regression-tests the validation fix over the real wire on both majors, which was your point about `tools/call` being the untested half.

The wait now keys on id 4 rather than id 2 — responses come back in order on a single session, so the last one arriving means all of them have.

Verified against a real `recce mcp-server --sse` on mcp 1.29.0 and 2.0.0 (20 tools, both assertions pass). Also verified it is not vacuous: with the argument made valid, the id 4 assertion fails with exit 1.

integration_tests/dbt/smoke_test.sh
iamcxa: Fixed in 122ffde4 — `trap 'kill $(jobs -p) 2>/dev/null || true' EXIT` set in the `mcp-server` branch, right before the server is backgrounded, so it covers both it and the SSE reader on every `exit 1` path.

Left `check_server_status` alone: it has the same shape but only one orphan, and changing it is outside what this PR touches.

integration_tests/dbt/smoke_test.sh
iamcxa: Fixed in 122ffde4 — the check now also asserts `lineage_diff` is named in the advertised list:

```bash
jq -e 'select(.id == 2) | [.result.tools[].name] | index("lineage_diff")'
```

Kept the count check as well; it catches the zero case with a clearer message. Confirmed 20 tools on both majors against a real server.

integration_tests/dbt/smoke_test.sh
iamcxa: Fixed in 122ffde4 — the comment now reads:

> `~=` keeps it a floor, not a pin — `~=1.29` is `>=1.29,<2.0`, so later patch *and* minor releases are picked up.

Behaviour unchanged; only the description was wrong.

tests/mcp_compat.py
iamcxa: Fixed in c0c0302d — `input_schema()` now tests for `None` rather than falsiness.

The production side needed the same read (validation has to fetch the schema across both majors), so there is now a `_tool_input_schema()` in `recce/mcp_server.py` written the same way. Deliberately not shared: a test helper that imports the production helper it exists to check would validate nothing.

iamcxa:
Pushed fixes for all 8 review threads directly to this branch (`c5c3395d..6bec82b4`, 6 commits) — details are in each thread reply. Flagging that here because it is your branch and it moved under you; happy to drop the commits if you would rather take them yourself.

| | |
|---|---|
| `c7f7987b` | `fix(mcp)` — validate tool arguments on mcp 2.0 |
| `c0c0302d` | `test(mcp)` — route the e2e protocol tests through the compat helpers |
| `122ffde4` | `test(smoke)` — call a tool, and clean up when the check fails |
| `74e33136` | `chore(deps)` — sync `uv.lock` with the widened mcp window |
| `23258ffe` | `ci(mcp)` — run the MCP tests against both SDK majors |
| `6bec82b4` | `docs(mcp)` — record what differs between the two SDK majors |

Verification, on both majors rather than only the one CI resolves:

| Check | mcp 1.29.0 | mcp 2.0.0 |
|---|---|---|
| `test_mcp_server` + `test_mcp_cloud_backend` + `test_mcp_e2e` | 259 passed | 259 passed |
| Full suite (`pytest tests/`) | 1583 passed, 5 skipped | 1583 passed, 5 skipped |
| Smoke check against a real `recce mcp-server --sse` | 20 tools, both tool-call assertions pass | identical |

The 5 skips are the usual `mcp`-gated ones. The only failures anywhere were the 5 `test_server.py` SPA routes that need a built frontend — they fail identically on `c5c3395d`, so they are environmental, not from these commits.

The new smoke assertion was checked for vacuity as well: with the deliberately-invalid argument made valid, it fails with exit 1 rather than passing quietly.

One thing left for you, since it is your text: the PR body still describes the stdio design that `bad52311` replaced with HTTP/SSE, and cites mcp 1.28.1 / 2.0.0 where the matrix is `["1.29", "2.0"]`. That body is what the release note inherits.

I am not clearing my own `REQUEST_CHANGES` — the fixes are mine, so self-approving them would not mean anything. Over to you.

@kentwelcome

iamcxa:
@kentwelcome — ready for your look at the 6 commits on `6bec82b4`. Details are in the individual thread replies; nothing new to add here.

CI is green on the new head, including the coverage that was missing before:

| Job | mcp version | Result |
|---|---|---|
| `mcp-smoke-test (1.29)` | `mcp==1.29.0` | 20 tools, tool-call assertions pass, **259 passed** |
| `mcp-smoke-test (2.0)` | `mcp==2.0.0` | 20 tools, tool-call assertions pass, **259 passed** |

That second row is the point — before this, no CI job ran pytest against mcp 2.x at all.

Copilot has been re-requested on the new commits. My `REQUEST_CHANGES` is deliberately still standing: the fixes are mine, so clearing it myself would not mean anything. Your call whether to take the commits, amend them, or ask me to drop them and hand the findings back.

One item is still yours, since it is your text: the PR body describes the stdio design that `bad52311` replaced with HTTP/SSE, and cites mcp 1.28.1 / 2.0.0 where the matrix is `["1.29", "2.0"]`. The release note inherits that body.

kentwelcome:
# Code Review: PR #1504

**SHA** `6bec82b4` · **Verdict** NO-GO · **Incremental** (`c5c3395d..6bec82b4`; the earlier commits carry review from Copilot, @kentwelcome and @iamcxa)

## Issues

1. `integration_tests/dbt/smoke_test.sh:202` — the `id 3` assertion passes when the response has no `result` at all, so it is a false green for the one regression this PR exists to catch.
   Evidence: `echo '{"jsonrpc":"2.0","id":3,"error":{"code":-32603,"message":"boom"}}' | jq -e 'select(.id == 3) | .result.isError != true'` prints `true` — `null.isError` is `null`, and `null != true`. An exception escaping as a JSON-RPC protocol error on 2.0 (exactly what `_handle_call_tool` is written to prevent) satisfies it. The `id 4` assertion is immune because it requires `.result.isError == true`.
   Pass E.

2. `recce/mcp_server.py:737` — an unknown tool name rebuilds the whole tool list on **every** call, which re-announces `tools/list` to the console and to the persistent MCPLogger record. The docstring and `tests/test_mcp_server.py:3542` both assert this cannot happen.
   Evidence: instrumented `_list_tools` at this head — known tool ×2 → 1 rebuild, then 0; unknown tool ×3 → **3 rebuilds**, each emitting `[MCP] Returning 20 tools: …` and one `log_list_tools` entry. `test_schema_lookup_does_not_re_announce_the_tool_list` only calls a name that is already cached, so it passes while the invariant it documents is broken on the path `test_unknown_tool_is_not_reported_as_a_validation_error` proves is live. A hallucinated tool name is the normal way an agent hits it.
   Pass A.

## Notes

1. `recce/mcp_server.py:756` — the validation block sits outside the `try`, so `jsonschema.SchemaError` leaves `_handle_call_tool` as an exception, which on 2.0 becomes the transport error the docstring says must not happen. Reachable by shipping one malformed schema (`"required": "select"` instead of `["select"]`) — it then fires on every call to that tool, not just malformed ones.
   Evidence: patching `_get_tool_input_schema` to return `{"type": "object", "required": "select"}` gives `RAISED out of _handle_call_tool: jsonschema.exceptions.SchemaError`.
   Pass D.

2. `recce/mcp_server.py:19` — `import jsonschema` is a direct import of a package the project does not declare; it resolves only because every `mcp` release in the `>=1.23,<3` window happens to pull `jsonschema>=4.20.0` (checked 1.23.0, 1.25.0, 1.28.1, 2.0.0 on PyPI). Adding it to the `mcp` extra removes the coupling.
   Pass C.

## Suggested fixes

**Issue 1** — require the member to exist:

```bash
if ! jq -e 'select(.id == 3) | .result != null and .result.isError != true' > /dev/null <<< "$responses"; then
```

**Issue 2** — build once, invalidate where the surface actually changes. `_tool_set_backend` is the only place `self.backend` / `self.context` / `self.single_env` move after `__init__` (lines 2589, 2618–2620, 2628, 2633), so a miss can stop meaning "re-read the list":

```python
if not self._tool_schema_cache:
    self._tool_schema_cache = {tool.name: _tool_input_schema(tool) for tool in await self._list_tools()}
return self._tool_schema_cache.get(name)
```

plus `self._tool_schema_cache = {}` in `_tool_set_backend`. That also closes the stale-schema case the refresh-on-miss was reaching for: a name present both before and after a swap never refreshes today.

Extending the existing test to a second unknown-name call would have caught it.

## Verified

- `tests/test_mcp_server.py` + `test_mcp_cloud_backend.py`: 199 passed on **mcp 1.29.0** and on **mcp 2.0.0**.
- `tests/test_mcp_e2e.py`: 60 passed (mcp 1.28.1). No `.isError` reads left on real SDK objects — the remaining hits are `mcp_compat.ToolResult`'s own field.
- `black --check`, `isort --check-only`, `flake8`: clean. `shellcheck`: 2 × SC2046, both the pre-existing `kill $(jobs -p)` idiom this file already uses.
- `uv lock --check`: clean at this head.
- `tox config -e dbtlatest` → `extras = mcp`; `-e dbt1.9` → empty. The factor-conditional `extras` resolves as intended.
- Coverage on `recce/mcp_server.py` reaches every new validation line.
- Base freshness: branch contains `origin/main` (`9f367a84`), `MERGEABLE`.

kentwelcome (commented):
NO-GO at `6bec82b4` — 2 issues, 2 notes. Full review: https://github.com/DataRecce/recce/pull/1504#issuecomment-5252411907 (`--request-changes` is not available on one's own PR.)

integration_tests/dbt/smoke_test.sh
```diff
@@ -104,10 +129,108 @@ function check_server_status() {
...
+        echo "The MCP server did not advertise lineage_diff."
+        exit 1
+    fi
+    if [ "${tool_count:-0}" -lt 1 ]; then
+        echo "The MCP server started but advertised no tools."
+        exit 1
+    fi
+    echo "MCP server is up and advertised $tool_count tools."
+
+    if ! jq -e 'select(.id == 3) | .result.isError != true' > /dev/null <<< "$responses"; then
```
kentwelcome: **ISSUE** — this passes when the response carries no `result` at all: `null.isError` is `null`, and `null != true`. An exception escaping as a JSON-RPC protocol error on 2.0 — the regression `_handle_call_tool` exists to prevent — satisfies it.

```suggestion
    if ! jq -e 'select(.id == 3) | .result != null and .result.isError != true' > /dev/null <<< "$responses"; then
```

recce/mcp_server.py
```diff
@@ -678,6 +698,77 @@ def _build_instructions(self) -> Optional[str]:
...
+    async def _get_tool_input_schema(self, name: str) -> Optional[Dict[str, Any]]:
+        """Look up a tool's JSON schema, refreshing the cache on a miss.
+
+        Mirrors the `_tool_cache` the mcp 1.x SDK keeps: `set_backend` can change the
+        advertised surface at runtime, so a miss means "re-read the list", not "unknown
+        tool". The cache is what keeps validation quiet — `_list_tools` logs its result
+        and writes an MCPLogger entry, so rebuilding it per call would forge a
+        `Returning N tools` line on every `tools/call`.
+        """
+        if name not in self._tool_schema_cache:
```
kentwelcome: **ISSUE** — a miss means "re-read the list", so every call with an unknown tool name rebuilds all 20 tool definitions and re-announces `tools/list` to the console and to the MCPLogger record. Measured at this head: unknown name ×3 → 3 rebuilds, 3 `Returning 20 tools` lines.

`_tool_set_backend` is the only place the advertised surface moves after `__init__`, so build once and invalidate there instead (plus `self._tool_schema_cache = {}` in that method):

```suggestion
        if not self._tool_schema_cache:
```

integration_tests/dbt/smoke_test.sh
kentwelcome: Fixed in `0350e2a7`. Verified all three shapes: a JSON-RPC error response for id 3 is now rejected, a real success still passes, and `isError: true` still fails.

recce/mcp_server.py
kentwelcome: Fixed in `0350e2a7`.

`_get_tool_input_schema` now builds once (`if not self._tool_schema_cache`), and `_tool_set_backend` clears the cache inside its lock — it is the only place `backend` / `context` / `single_env` move after `__init__`, so that also closes the stale-schema case refresh-on-miss was reaching for.

Measured at the new head: unknown name ×3 → **0** rebuilds (was 3). The existing test now calls an unknown name twice and fails against the old code; a second test asserts `set_backend` empties the cache.

kentwelcome:
## Review outcome — `0350e2a7`

Both issues from [the review](https://github.com/DataRecce/recce/pull/1504#issuecomment-5252411907) are fixed and pushed. Re-verdict: **GO**.

| Finding | Status | Evidence |
|---|---|---|
| `smoke_test.sh:202` — id-3 assertion false green | Fixed | Rejects a JSON-RPC error response, still passes a real success, still fails on `isError: true` |
| `mcp_server.py:737` — unknown name re-announces the tool list | Fixed | Unknown name ×3 → 0 rebuilds (was 3); the extended test fails against the pre-fix code |

Both notes are left open deliberately, as non-blocking author calls:

- `mcp_server.py:756` — `jsonschema.SchemaError` still escapes as a transport error if a tool ships a malformed schema.
- `mcp_server.py:19` — `jsonschema` stays an undeclared direct import, satisfied by `mcp`'s own dependency across the whole `>=1.23,<3` window.

Verified at this head: 200 tests on **mcp 1.29.0** and on **mcp 2.0.0**, 60 e2e on 1.28.1, `black` / `isort` / `flake8` clean, and the real `recce mcp-server --sse` smoke check green on both majors (20 tools, success and invalid-input assertions).

GitHub does not allow `--approve` on one's own PR, so the review state cannot be flipped from here; the verdict above is the record.

iamcxa (approved):
LGTM!

PR #1505, opened 2026-08-12
fix(ci): bump twine to >=7 so PyPI uploads accept Metadata-Version 2.5

kentwelcome:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

`fix` / CI

The nightly PyPI release has been failing at the upload step for both packages:

```
ERROR    InvalidDistribution: Invalid distribution metadata: '2.5' is not a
         valid metadata version
```

Failing run: https://github.com/DataRecce/recce/actions/runs/31554418260/job/93983773260

Nothing in the repo changed. Two unpinned moving parts lined up:

1. `hatchling` is a build requirement, so `uv build` fetches the latest into an isolated env on every run. Current hatchling emits `Metadata-Version: 2.5`.
2. The locked `twine==6.2.0` monkeypatches `packaging.metadata._VALID_METADATA_VERSIONS` to its own hardcoded list that stops at `2.4` (the "Monkeypatch Metadata 2.0 support" block in `twine/package.py`). So it rejected the dist even though the locked `packaging==26.2` already accepts 2.5.

`twine 7.0.0` removed that monkeypatch and defers to `packaging`. This PR bumps `twine` to `>=7` in the `dev` extra and relocks. twine 7 requires Python >=3.10; the release jobs already `uv sync --python 3.10`.

This unblocks all four upload steps, which share the root workspace lock:

- `.github/workflows/nightly.yaml:119` (recce-nightly)
- `.github/workflows/nightly.yaml:209` (recce-cloud-nightly)
- `.github/workflows/release.yaml:77` (recce)
- `.github/workflows/release.yaml:136` (recce-cloud)

Closes DRC-4132

Verified against the relocked env, reproducing the exact command shape that failed in CI:

```
$ uv run twine --version
twine version 7.0.0 (... packaging: 26.2)

$ uv build --package recce-cloud
$ unzip -p dist/*.whl '*/METADATA' | head -1
Metadata-Version: 2.5

$ uv run twine check dist/*
Checking recce_cloud-1.32.0.dev0-py3-none-any.whl: PASSED
Checking recce_cloud-1.32.0.dev0.tar.gz: PASSED
```

Before the bump, the same `twine check` failed with the CI error. The only lockfile movement is `twine 6.2.0 -> 7.0.0`.

Two follow-ups deliberately left out of this PR, tracked on DRC-4132:

- This class of break will recur. `hatchling` is unpinned in `[build-system] requires`, `astral-sh/setup-uv` runs `version: "latest"`, and twine floats. Pinning hatchling would make release builds reproducible.
- `uv publish` would drop twine from the release path entirely, but it does not read `.pypirc` — it would need a `UV_PUBLISH_TOKEN` secret instead of the current `PYPI` pypirc secret.

```release-note
NONE
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,9 @@ dev = [
     "pre-commit>=4.2.0",
     "pytest-mypy",
     "pytest-cov",
-    "twine",
+    # twine<7 monkeypatches packaging to cap Metadata-Version at 2.4, which
+    # rejects the 2.5 metadata current hatchling emits (DRC-4132).
+    "twine>=7",
     "tox",
     "tox-uv",
     "pandas",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ dev = [
     "pytest-mypy",
     "pytest-cov",
     # twine<7 monkeypatches packaging to cap Metadata-Version at 2.4, which
-    # rejects the 2.5 metadata current hatchling emits (DRC-4132).
+    # rejects the 2.5 metadata current hatchling emits.
     "twine>=7",
     "tox",
     "tox-uv",
--- a/uv.lock
+++ b/uv.lock
@@ -807,7 +807,7 @@ name = "exceptiongroup"
 version = "1.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
+    { name = "typing-extensions" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/0b/9f/a65090624ecf468cdca03533906e7c69ed7588582240cfe7cc9e770b50eb/exceptiongroup-1.3.0.tar.gz", hash = "sha256:b241f5885f560bc56a59ee63ca4c6a8bfa46ae4ad651af316d4e81817bb9fd88", size = 29749, upload-time = "2025-05-10T17:42:51.123Z" }
 wheels = [
@@ -2583,7 +2583,7 @@ requires-dist = [
     { name = "sqlglot", specifier = ">=27.29.0,<31.0" },
     { name = "tox", marker = "extra == 'dev'" },
     { name = "tox-uv", marker = "extra == 'dev'" },
-    { name = "twine", marker = "extra == 'dev'" },
+    { name = "twine", marker = "extra == 'dev'", specifier = ">=7" },
     { name = "uvicorn" },
     { name = "watchdog" },
     { name = "websockets" },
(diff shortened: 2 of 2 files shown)


kentwelcome:
# Code Review: PR #1505

**SHA** `406d8221` · **Verdict** GO

Passes A, B, F (config/infra). No blockers, no issues.

The central risk in this change is that `twine check` passing locally does not prove PyPI will accept the artifact — twine validates against its own copy of `packaging`, Warehouse validates independently. Verified directly: `hatchling 1.32.0` and `hatch 1.18.0` are live on PyPI right now carrying `Metadata-Version: 2.5`, both uploaded 2026-08-11. Warehouse accepts 2.5, so the bump is a sufficient fix rather than a local-check pass.

Root cause corroborated on the timeline: hatchling 1.32.0 released 2026-08-11T05:03 with "bump default core metadata version to 2.5" (PEP 794); the nightly failed 2026-08-12T01:41, the first run after it. `hatchling` is unpinned in `[build-system] requires`, so `uv build` picks it up fresh every run.

Verification performed:

- `uv sync --all-extras --python 3.10` — the exact nightly interpreter, which **no CI job on this branch exercises** (tests-python uses 3.12). Resolves to Python 3.10.20 + twine 7.0.0. twine 7 requires `>=3.10`, so this was the one way the fix could have silently failed to reach the release job. It does not.
- Both published packages, not just the one cited in the PR body: `recce-1.60.0.dev0` and `recce_cloud-1.32.0.dev0` build to `Metadata-Version: 2.5` and pass `twine check` (wheel + sdist).
- `pytest tests/` — 1589 passed, 5 skipped.
- twine 7.0.0's only breaking change is dropping metadata **2.0**, which hatchling never emits here. `.pypirc` reading is intact (7.0.0 only changed it to specify UTF-8), so the workflows' `$HOME/.pypirc` auth path is unaffected.
- No new transitive dependencies: twine's dependency set in the lock is byte-identical across the bump.
- All 20 CI checks green.

## Notes

1. `PR body` — The claim "The only lockfile movement is `twine 6.2.0 -> 7.0.0`" is not accurate. Three dependency markers were also relaxed.
   Evidence: `uv.lock` — `exceptiongroup`'s `typing-extensions` marker `python_full_version < '3.11'` → unconditional; `secretstorage`'s `cryptography` and `jeepney` markers `python_full_version < '3.12' or sys_platform != 'win32'` → unconditional. Both are provably inert: `exceptiongroup` is itself only installed under `python_full_version < '3.11'` (`uv.lock:60,2129`) and `secretstorage` only under `sys_platform == 'linux'` (`uv.lock:1111`), so the old markers were already always-true wherever those packages land. Harmless, but the statement should not be relied on by a future reader diffing the lock.
   Pass F.

2. `pyproject.toml:62` — This fixes metadata 2.5 specifically, not the class. The locked `packaging==26.2` tops out at 2.5; `packaging 26.3` (already released) is the first to accept 2.6. Since twine 7 defers to `packaging`, the next hatchling core-metadata bump reproduces this exact error until the lock is refreshed. Mitigating: `packaging` is unconstrained, so any routine relock picks up support. Already captured in DRC-4132's follow-ups; raising it here so the deferral is a recorded decision rather than an oversight.
   Pass F.

3. `.github/workflows/` — Nothing in CI exercises `uv build` + `twine check`, which is why a packaging break could only surface in the 2 AM nightly. A build-and-check step gated on changes to `pyproject.toml` / `uv.lock` / `recce_cloud/pyproject.toml` would have caught this pre-merge, and would catch the recurrence in Note 2. Out of scope for a release-unblocking fix; worth its own issue.
   Pass F.

## Limits

Pass B was scoped to supply-chain facts rather than a data-flow trace: the diff contains no source code, no entry points, and no runtime dependency change. Checked instead: artifact provenance (`files.pythonhosted.org`, sha256-pinned in the lock), no new transitive deps, and twine remaining `extra == 'dev'` only, so nothing reaches the published wheel.

wcchang1115 (approved):
LGTM, thanks!

PR #1508, opened 2026-08-17
feat(cloud): add list-orgs and list-projects commands

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Feature.

`recce-cloud init --org X --project Y` already binds a directory without prompts. But nothing in the CLI lists the organizations and projects you can pass to it. Those names appear only inside the interactive picker in `init`, which needs a person to answer. An agent or a script therefore cannot finish the setup on its own.

This PR adds `recce-cloud list-orgs` and `recce-cloud list-projects`. Both print a table, or JSON with `--json`. There is no API change: the client already had `list_organizations` and `list_projects`, and until now only the interactive branch of `init` called them.

Read the second commit only. The first commit is black formatting of the same two files, with no logic change. They were written at 88 columns while the repository setting is 120, so any edit reformats them in full.

Yes.

Add `recce-cloud list-orgs` and `recce-cloud list-projects`. You can now find the organization and project IDs that `recce-cloud init --org ... --project ...` needs without going through the interactive picker. Both commands support `--json` for scripts and agents.

### DIFF
--- a/recce_cloud/tests/test_cli.py
+++ b/recce_cloud/tests/test_cli.py
@@ -1977,12 +1977,15 @@ class TestListProjects(unittest.TestCase):
     def test_no_org_available(self):
         """Exit with code 2 when no organization can be resolved."""
         with patch.dict(os.environ, {"RECCE_API_TOKEN": "test_token"}, clear=True):
+            # resolver.py binds get_project_binding at import time, so the patch
+            # must target the resolver module, not project_config.
             with patch(
-                "recce_cloud.config.project_config.get_project_binding",
+                "recce_cloud.config.resolver.get_project_binding",
                 return_value=None,
-            ):
+            ) as mock_binding:
                 result = self.runner.invoke(cloud_cli, ["list-projects"])
 
+        self.assertTrue(mock_binding.called, "get_project_binding was not patched in the code under test")
         self.assertEqual(result.exit_code, 2, f"Output: {result.output}")
         self.assertIn("recce-cloud list-orgs", result.output)
 
--- a/recce_cloud/tests/test_cli.py
+++ b/recce_cloud/tests/test_cli.py
@@ -323,9 +323,7 @@ class TestUploadDryRun(unittest.TestCase):
         self.assertEqual(result.exit_code, 0, f"Command failed: {result.output}")
         self.assertIn("Dry run mode enabled", result.output)
         self.assertIn("Auto-create session and upload", result.output)
-        self.assertIn(
-            "Warning: Platform not supported for auto-session creation", result.output
-        )
+        self.assertIn("Warning: Platform not supported for auto-session creation", result.output)
 
     def test_dry_run_missing_artifacts(self):
         """Test dry-run with missing dbt artifacts."""
@@ -432,27 +430,19 @@ class TestUploadBlocksDevSession(unittest.TestCase):
         }
 
         with patch.dict(os.environ, env, clear=True):
-            with patch(
-                "recce_cloud.config.project_config.get_project_binding"
-            ) as mock_binding:
+            with patch("recce_cloud.config.project_config.get_project_binding") as mock_binding:
                 mock_binding.return_value = {
                     "org": "test-org",
                     "project": "test-project",
                 }
-                with patch(
-                    "recce_cloud.api.client.RecceCloudClient.get_organization"
-                ) as mock_org:
+                with patch("recce_cloud.api.client.RecceCloudClient.get_organization") as mock_org:
                     mock_org.return_value = {"id": "org-123", "slug": "test-org"}
-                    with patch(
-                        "recce_cloud.api.client.RecceCloudClient.get_project"
-                    ) as mock_proj:
+                    with patch("recce_cloud.api.client.RecceCloudClient.get_project") as mock_proj:
                         mock_proj.return_value = {
                             "id": "proj-456",
                             "slug": "test-project",
                         }
-                        with patch(
-                            "recce_cloud.api.client.RecceCloudClient.list_sessions"
-                        ) as mock_sessions:
+                        with patch("recce_cloud.api.client.RecceCloudClient.list_sessions") as mock_sessions:
                             mock_sessions.return_value = []  # No production session found
                             result = self.runner.invoke(
                                 cloud_cli,
@@ -485,9 +475,7 @@ class TestUploadBlocksDevSession(unittest.TestCase):
                 ["upload", "--target-path", self.temp_dir, "--dry-run"],
             )
 
-        self.assertEqual(
-            result.exit_code, 0, f"Dry run should succeed: {result.output}"
-        )
+        self.assertEqual(result.exit_code, 0, f"Dry run should succeed: {result.output}")
         self.assertIn("Blocked", result.output)
         self.assertIn("feature branch with no PR/MR detected", result.output)
 
@@ -710,9 +698,7 @@ class TestDownloadDryRun(unittest.TestCase):
 
         # Assertions
         self.assertEqual(result.exit_code, 0, f"Command failed: {result.output}")
-        self.assertIn(
-            "Warning: Target path exists (use --force to overwrite)", result.output
-        )
+        self.assertIn("Warning: Target path exists (use --force to overwrite)", result.output)
 
     def test_dry_run_session_id_with_prod_warning(self):
         """Test dry-run shows warning when both --session-id and --prod are provided."""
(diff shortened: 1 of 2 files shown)


kentwelcome:
# Code Review: PR #1508

**SHA** `947f04cf` · **Verdict** NO-GO

## Issues

1. `recce_cloud/tests/test_cli.py:1981` — The patch target never reaches the code under test, so `test_no_org_available` asserts nothing and makes a live HTTP call.
   Evidence: `recce_cloud/config/resolver.py:16` binds `get_project_binding` at import time, so patching `recce_cloud.config.project_config.get_project_binding` leaves `resolver.get_project_binding` untouched — `mock.called` is `False`. The test passes only because the cwd happens to have no `.recce/config`. Run from a directory that does:
   ```
   AssertionError: 'recce-cloud list-orgs' not found in
   'Error: Failed to list projects: Failed to verify token.\n'
   ```
   That message is an unmocked outbound request with the literal token `test_token`. Patching `recce_cloud.config.resolver.get_project_binding` instead gives `mock called: True`, `exit: 2`, and the intended output.
   The 20+ existing sites using this string work because `cli.py:396` imports the symbol *inside* a function, where the attribute is re-resolved per call; `resolve_org_id` does not.
   Pass C, Pass E.

## Notes

1. `recce_cloud/recce_cloud/cli.py:1160` — `--org` and the server's `display_name` are interpolated into rich markup, so a value with a stray closing tag exits 1 on an unhandled exception with no output: `recce-cloud list-projects --org '[/red]'` → `MarkupError("closing tag '[/red]' at position 56 doesn't match any open tag")`. Same via `display_name` at `:1184`/`:1187`/`:1191`. Not introduced here — `init` at `cli.py:270` crashes identically on the same input — but these two commands are the agent/script entry points, where a traceback is the worst failure shape. `escape()` or `markup=False` on the interpolated value.
2. `recce_cloud/recce_cloud/cli.py:1094` and `:1203` — `str(org.get("id", "-"))` prints the literal `None` when `id` is present but null; the sibling fields on the adjacent lines use `or "-"` and handle it.
3. `recce_cloud/recce_cloud/api/client.py:57` — `RecceCloudClient._request` passes no `timeout`, so both new commands can block indefinitely. Pre-existing and shared with `list-sessions` (`ReportClient._request` at `:876` does set `timeout=60`), but it matters more for the unattended use this PR is for.
4. `recce_cloud/recce_cloud/cli.py:1024` — `_require_api_token` duplicates the four-line block still inlined at `:223`, `:910`, and `:1648` rather than replacing them. The helper's `sys.exit(2)` matches `:913`/`:1651`; `init` at `:226` still exits 1 for the same condition.
5. `recce_cloud/README.md:292-315` — the "Additional Commands" list enumerates `list`, `delete`, `review`, `report`, `doctor`, `version`; the two new commands are not added. (`RECCE_CLOUD.md:75-80` is separately stale — it still says only the version command is implemented.)

## Verified

- `d1b53ae2` is AST-identical to its parent for both files — the reformat carries no behavior change.
- `black --check`, `isort --check`, `flake8` clean on both changed files; 67 tests pass.
- Merges cleanly into `main`; base has no commits since the branch diverged.
- Pass B: the token reaches only the `Authorization` header. `RecceCloudException.reason` is the server response body, and `logger.debug(..., exc_info=True)` carries no locals — no leak in output, logs, or error text.
- `--org` never reaches URL construction: `get_organization` lists all orgs and filters in Python, so no path injection from the flag.
- The `sys.exit(2)` calls inside the `try` at `:1160`/`:1164` do fire — `SystemExit` derives from `BaseException`, so `except Exception` does not swallow them.

recce_cloud/tests/test_cli.py
```diff
@@ -1963,5 +1896,197 @@ def test_dry_run_session_base_with_gitlab_ci_job_token(self):
...
+                "status": "archived",
+                "repository": {"full_name": "acme/old-dbt"},
+            },
+        ]
+
+    def test_no_org_available(self):
+        """Exit with code 2 when no organization can be resolved."""
+        with patch.dict(os.environ, {"RECCE_API_TOKEN": "test_token"}, clear=True):
+            with patch(
+                "recce_cloud.config.project_config.get_project_binding",
```
kentwelcome: This patch target never reaches the code under test: `resolver.py:16` binds `get_project_binding` at import time, so `resolver.get_project_binding` keeps pointing at the real function and `mock.called` stays `False`. The test passes only because the cwd has no `.recce/config`, and it makes a live HTTP call.

```suggestion
                "recce_cloud.config.resolver.get_project_binding",
```

kentwelcome (changes requested):

kentwelcome:
✅ **Code Review Loop — fixes pushed** (`a8e04a9f`)

- `test_cli.py:1981` patch target — fixed in `a8e04a9f`. The patch now targets `recce_cloud.config.resolver.get_project_binding`, and the commit adds `assertTrue(mock_binding.called)` so a wrong target fails loudly instead of passing silently.

Re-verified against the pushed HEAD, independently of the fix agent:

- The exact scenario that failed before — running the test from a directory containing a bound `.recce/config` — now passes in 0.19s, confirming both the env-dependence and the live API call are gone.
- 67/67 tests pass; `black --check`, `isort --check`, `flake8` clean on the changed file.

The five **Notes** in the review were out of scope for this fix round and remain open for the author. None of them block merge.

recce_cloud/tests/test_cli.py
kentwelcome: @kentwelcome Confirmed and fixed in `a8e04a9f`. Applied your suggestion — the patch target is now `recce_cloud.config.resolver.get_project_binding` (`recce_cloud/tests/test_cli.py:1983`), and I added `self.assertTrue(mock_binding.called, ...)` at `:1988` so a wrong target fails the test instead of passing silently.

Reproduced your evidence first. From a temp dir containing a bound `.recce/config` (`cloud.org_id: "42"`, `cloud.project_id: "7"`), before the fix:

```
E       AssertionError: 'recce-cloud list-orgs' not found in 'Error: Failed to list projects: Failed to verify token.\n'
recce_cloud/tests/test_cli.py:1987: AssertionError
1 failed, 66 deselected in 1.21s
```

Root cause is exactly as you described: `recce_cloud/recce_cloud/config/resolver.py:16` is a module-level `from recce_cloud.config.project_config import get_project_binding`, and `resolve_org_id` (`resolver.py:149`) calls that module-local name. `cli.py:1142` imports `resolve_org_id` from `resolver`, so nothing re-resolves per call — unlike `cli.py:396`, which imports `get_project_binding` inside the function body, which is why the other patch sites in this file are unaffected.

Side-by-side probe, run from the same bound temp dir, both targets against `list-projects`:

```
target=recce_cloud.config.project_config.get_project_binding
  mock called: False
  exit: 2
  'recce-cloud list-orgs' in output: False
  output: 'Error: Failed to list projects: Failed to verify token.\n'

target=recce_cloud.config.resolver.get_project_binding
  mock called: True
  exit: 2
  'recce-cloud list-orgs' in output: True
  output: "Error: No organization specified\nPass --org, set the RECCE_ORG environment variable, or run 'recce-cloud init' to\nbind this directory\nRun 'recce-cloud list-orgs' to see the available organizations\n"
```

After the fix, the whole file passes from that bound temp dir (67 passed) as well as from the repo root, so the test no longer depends on the working directory and no longer reaches the network.

kentwelcome:
@kentwelcome Addressed the blocking finding from the review of `947f04cf`. Pushed `a8e04a9f`.

**Issues 1 — `recce_cloud/tests/test_cli.py:1981` (patch target never reaches the code under test): valid, fixed.**

The patch target is now `recce_cloud.config.resolver.get_project_binding` (`test_cli.py:1983`), and `test_no_org_available` asserts `mock_binding.called` at `:1988` so this class of mistake can't pass silently again. Details, including the reproduction and a before/after probe showing `mock called: False` → `True`, are in the [thread on the inline comment](https://github.com/DataRecce/recce/pull/1508#discussion_r3795207503).

Verification on `a8e04a9f`, scoped to the changed file:

- `black --check`, `isort --check`, `flake8` — clean.
- `pytest recce_cloud/tests/test_cli.py -q` — 67 passed, from the repo root.
- Same file from a temp dir holding a bound `.recce/config` — 67 passed. Before the fix that directory produced `AssertionError: 'recce-cloud list-orgs' not found in 'Error: Failed to list projects: Failed to verify token.'`, i.e. a live outbound request with the literal `test_token`. That is gone.

**Notes 1–5: not addressed in this commit**, left for the author to decide on. For the record, my read of each:

1. Rich markup injection via `--org` / `display_name` (`cli.py:1160`, `:1184`/`:1187`/`:1191`) — real, and worth fixing here since these are the agent/script entry points. `escape()` on the interpolated value.
2. `str(org.get("id", "-"))` prints `None` for a present-but-null `id` (`cli.py:1094`, `:1203`) — real, one-line fix, inconsistent with the sibling fields.
3. No `timeout` in `RecceCloudClient._request` (`api/client.py:57`) — pre-existing and shared with `list-sessions`; touches every cloud command, so probably its own PR.
4. `_require_api_token` not applied at `:223`, `:910`, `:1648` (`cli.py:1024`) — cleanup; note `init` at `:226` exits 1 where the helper exits 2, so consolidating changes an exit code.
5. `recce_cloud/README.md:292-315` missing the two new commands (and `RECCE_CLOUD.md:75-80` stale) — docs, cheap to add.

kentwelcome (approved):
LGTM

PR #1514, opened 2026-08-24
feat(histogram): show semantic Base/Current overlap

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Feature / accessibility

- Renders the true geometric overlap between Base and Current histogram bars with the semantic purple crosshatch token.
- Adds an exact Base / Current / Overlap legend while preserving normal Base and Current visibility toggles.
- Keeps tooltips backed by the two source datasets and expands the chart's accessible description to explain overlap.
- Covers category, numeric, and datetime histograms; hidden datasets; malformed geometry; light/dark theming; contrast; and visual grayscale behavior.

Closes DRC-2852

The overlap painter uses the actual rectangle intersection of aligned, unstacked source bars. Its translucent semantic fill is pre-composited against the themed chart background before painting so it replaces the mixed Base/Current pixels without clearing unrelated canvas content.

Validation:

- `pnpm lint:fix`
- `pnpm type:check`
- `pnpm test` — 4,134 passed, 5 skipped
- `pnpm run build`
- Targeted Storybook visual suite — 5 passed in light, dark, and grayscale

```release-note
Histograms now identify the shared Base and Current distribution with a distinct purple crosshatch and an Overlap legend entry.
```


kentwelcome:
# Code Review: PR #1514

**SHA** `4bd515c1` · **Verdict** NO-GO

## Blockers

1. `js/packages/ui/src/components/data/HistogramChart.tsx:349` — The overlap plugin is captured at mount only, so a light/dark switch recolors the bars and the legend but leaves the crosshatch painted in the previous theme's palette.
   Evidence: `react-chartjs-2@5.3.1/dist/index.js` reads `plugins` only inside `renderChart()`; the post-mount effects are keyed `[redraw, options]`, `[redraw, data.labels]`, `[redraw, data.datasets]`, `[…, updateMode]`, and `[type]` — none lists `plugins`. `overlapPlugin` (line 168) closes over `overlapPalette` (line 153), which is rebuilt when `isDark` flips, but the live chart keeps the mount-time instance. Everything carried in `options` *does* refresh (`setOptions` does `Object.assign`), so `generateLabels` and the dataset fills switch while the painter does not: after toggling to dark you get `compositeHex("#805AD5A5", "#FAFAFA")` — light lavender — filling the overlap on a dark chart. Second-order effect: the module-level single-slot `patternCache` (histogramOverlap.ts:41) then alternates between the stale plugin key and the fresh legend key, allocating a new tile `<canvas>` and `CanvasPattern` on every draw.
   Fix: register the plugin once and pass the palette through chart.js plugin options (`options.plugins.histogramOverlap`), read from the hook's third argument — `options` is reassigned every render, the `plugins` array is not.
   Pass A.

## Issues

2. `js/packages/ui/src/components/data/histogramOverlap.ts:91` — Two of the four crosshatch strokes fall outside the 8×8 tile and paint nothing, so the tiled pattern breaks at every tile corner.
   Evidence: the tile clips to `x, y ∈ [0, 8]`. Lines 91–92 trace `y = x − 8`, whose only in-range point is `(8, 0)`; lines 95–96 trace `y = −x`, whose only in-range point is `(0, 0)`. The two that do draw are `y = x + 2` → `(0,2)–(6,8)` and `y = −x + 10` → `(2,8)–(8,2)`. Their 8px-period continuations are `y = x − 6` and `y = −x + 2`, which is what those dead strokes appear to have been aiming at.
   Pass A.

3. `js/packages/ui/src/components/data/histogramOverlap.ts:112` — `Number.isFinite` accepts chart.js's off-scale sentinel `−32768`, so floating-bar datasets yield a ~33,000px overlap rect, and `fillRect` draws it with no `chartArea` clip.
   Evidence: driving chart.js 4.5.1 with this component's datetime config (`counts.map((v, i) => [binEdges[i], v])`, HistogramChart.tsx:180 — an array entry is parsed as a *floating bar*, chart.js `parseFloatBar`) returns bars as `{x: -32768, y: -32768, width: 517.3, base: 170.8}`. `intersectBars` turns that into `{left: -33034.7, top: -32768, width: 517.3, height: 32858.5}`. It is invisible today only because `x` is off-canvas too; `afterDatasetsDraw` runs outside chart.js's dataset clip, so any on-canvas `x` would paint that rect over the axes, title and legend.
   Fix: clamp the rect to `chart.chartArea` before filling (or `ctx.clip()` to it), which also makes the painter safe against any future scale `min`/`max` that puts bar geometry outside the plot.
   Pass A.

4. `js/packages/ui/src/theme/__tests__/semanticColors.test.ts:7` — The new contrast test composites with the test file's own `compositeHex`, not the one that ships, so it cannot fail when the shipped math drifts.
   Evidence: `histogramOverlap.ts:47` adds a second, module-private `compositeHex` (alpha guarded by `foreground.length >= 9`, space-separated `rgb()` output); the copy at `semanticColors.test.ts:7` is unguarded and comma-separated. The shipped copy is what decides the crosshatch's opaque fill, and nothing asserts on it. Export one implementation and have both call it.
   Pass E.

5. `js/packages/ui/src/components/data/histogramOverlap.ts:13` — The hardcoded dataset indices are coupled to HistogramChart's dataset ordering, and no test pins that ordering.
   Evidence: `HistogramChart.tsx:206-211` builds `[current, base]`; the module hardcodes `CURRENT_DATASET_INDEX = 0` / `BASE_DATASET_INDEX = 1`. `HistogramChart.test.tsx:259` asserts only `toHaveLength(2)`, and the legend test feeds a hand-built mock chart, so swapping the two `buildDataset` calls would invert every legend swatch and every click-to-hide target with all 4134 tests still green. Assert `data.datasets.map((d) => d.label)` equals `["Current", "Base"]`.
   Pass C.

## Notes

- `histogramOverlap.ts:83` — when `tile.getContext("2d")` returns null the fallback fill is returned without caching, so a canvas-less context allocates a fresh `<canvas>` on every draw.
- `chartjs-adapter-date-fns` is imported only by `js/packages/storybook/.storybook/preview.tsx`, never by the shipped app, so the `dataType="datetime"` path reaching `HistogramResultView.tsx:117` runs a `timeseries` scale with no date adapter. Pre-existing and outside this diff — but it is the reason finding 3 is latent instead of visible.

## Verification

`pnpm lint` and `pnpm type:check` clean. `pnpm test`: 196 files, 4134 passed / 5 skipped. Branch merges cleanly into `main` at `5110a5f9` with no base commits since divergence. The `stacked: true` → `grouped: false` swap on the category axis is geometrically identical (verified against chart.js 4.5.1: same `x`, `width`, `y`, `base` for every bar in both configs), and the three `ColorSemanticsMatrix` snapshot updates are expected — that story renders `HistogramChart` at `ColorSemanticsMatrix.stories.tsx:411`.

Ran with Node 26.7.0; `.nvmrc` pins 26.5.0, which is not installed on this machine.

js/packages/ui/src/components/data/histogramOverlap.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  tileContext.fillStyle = opaqueOverlapFill;
+  tileContext.fillRect(0, 0, PATTERN_SIZE, PATTERN_SIZE);
+  tileContext.strokeStyle = overlap.foreground;
+  tileContext.lineWidth = 1;
+  tileContext.beginPath();
+  tileContext.moveTo(-2, 0);
+  tileContext.lineTo(PATTERN_SIZE, PATTERN_SIZE + 2);
+  tileContext.moveTo(PATTERN_SIZE - 2, -2);
+  tileContext.lineTo(PATTERN_SIZE + 2, 2);
```
kentwelcome: Two of the four crosshatch strokes fall outside the 8x8 tile and paint nothing, so the tiled pattern breaks at every tile corner. This pair traces `y = x - 8`, whose only in-range point is `(8, 0)`; the seamless continuation of the `y = x + 2` stroke above it is `y = x - 6`.

```suggestion
  tileContext.moveTo(PATTERN_SIZE - 4, -2);
  tileContext.lineTo(PATTERN_SIZE + 2, 4);
```

js/packages/ui/src/components/data/histogramOverlap.ts
```diff
@@ -0,0 +1,230 @@
...
+  tileContext.moveTo(PATTERN_SIZE + 2, 0);
+  tileContext.lineTo(0, PATTERN_SIZE + 2);
+  tileContext.moveTo(2, -2);
+  tileContext.lineTo(-2, 2);
```
kentwelcome: Same defect on the other diagonal family: this pair traces `y = -x`, whose only in-range point is `(0, 0)`. The seamless continuation of the `y = -x + 10` stroke above it is `y = -x + 2`.

```suggestion
  tileContext.moveTo(4, -2);
  tileContext.lineTo(-2, 4);
```

js/packages/ui/src/components/data/histogramOverlap.ts
```diff
@@ -0,0 +1,230 @@
...
+      ctx.fillStyle = pattern;
+
+      for (let index = 0; index < length; index += 1) {
+        const current = readBarGeometry(currentBars[index]);
+        const base = readBarGeometry(baseBars[index]);
+        if (!current || !base) continue;
+
+        const overlap = intersectBars(current, base);
+        if (!overlap) continue;
+        ctx.fillRect(overlap.left, overlap.top, overlap.width, overlap.height);
```
kentwelcome: `Number.isFinite` accepts chart.js's off-scale sentinel `-32768`, which floating-bar datasets (the `dataType="datetime"` path) produce. `afterDatasetsDraw` runs outside chart.js's dataset clip, so the resulting ~33,000px rect can paint over the axes, title and legend. Clamp to the plot area:

```suggestion
        const overlap = intersectBars(current, base);
        if (!overlap) continue;
        const { left, top, right, bottom } = chart.chartArea;
        const x0 = Math.max(overlap.left, left);
        const y0 = Math.max(overlap.top, top);
        const x1 = Math.min(overlap.left + overlap.width, right);
        const y1 = Math.min(overlap.top + overlap.height, bottom);
        if (x1 <= x0 || y1 <= y0) continue;
        ctx.fillRect(x0, y0, x1 - x0, y1 - y0);
```

kentwelcome (changes requested):
Code Review Loop: 1 blocker, 4 issues. See review comment.

kentwelcome:
✅ **Code Review Loop — fixes pushed** (`18774675`)

- **Blocker 1** (stale theme palette) — fixed in `18774675`. The painter is now a stable, stateless `histogramOverlapPlugin` and the palette travels in `options.plugins.histogramOverlap`. Verified against chart.js 4.5.1 directly: with a plugin registered at construction, `Object.assign(chart.options, next)` + `chart.update()` — exactly what react-chartjs-2's `setOptions` effect does — delivers the new palette to `afterDatasetsDraw`. Two new tests cover it.
- **Issue 2** (dead crosshatch strokes) — fixed in `18774675`; both stitch pairs now trace `y = x - 6` and `y = -x + 2`, the 8px continuations of the strokes above them.
- **Issue 3** (unclipped off-scale rect) — fixed in `18774675`; every rect is clamped to `chart.chartArea` before painting, with a regression test using chart.js's `-32768` sentinel.
- **Issue 4** (test composited with its own copy) — fixed in `f95aa156`. `compositeRgbChannels` in `semanticColors.ts` is now the single implementation. It is split from the `rgb()` formatting deliberately: the canvas takes the space-separated form the repo's CSS convention requires, while MUI's `decomposeColor` only parses the comma-separated form. The contrast test now exercises the shipped arithmetic.
- **Issue 5** (dataset-order coupling) — partially rebutted, then covered. The claim that no test pinned the dataset order was wrong: `HistogramChart.test.tsx` already asserted `datasets[0].label === "Current"` / `datasets[1].label === "Base"`, proven by swapping the `buildDataset` calls and watching that test fail. What genuinely was uncovered is the link from `CURRENT_DATASET_INDEX` / `BASE_DATASET_INDEX` to those labels; the new test walks each legend item's `datasetIndex` back into `data.datasets` and asserts the label matches.

Verified independently at `18774675`: `pnpm lint` clean, `pnpm type:check` clean, `pnpm test` 196 files / 4141 passed / 5 skipped (up from 4134 — 7 new tests, no snapshot churn).

js/packages/ui/src/components/data/histogramOverlap.ts
kentwelcome: Applied as suggested in `18774675`.

Confirmed the geometry: the stroke above it runs `(-2, 0)` → `(8, 10)`, i.e. `y = x + 2`, so the seamless 8px-period continuation is `y = x - 6` — exactly `(4, -2)` → `(10, 4)`. The old pair traced `y = x - 8`, which only touches the tile at `(8, 0)`.

`histogramOverlap.ts:87-97` now draws two strokes per diagonal family, one tile-period apart, with a comment saying why the second one exists so it does not get "simplified" away again.

js/packages/ui/src/components/data/histogramOverlap.ts
kentwelcome: Applied as suggested in `18774675`.

Same check on this family: the stroke above runs `(10, 0)` → `(0, 10)`, i.e. `y = -x + 10`, so the continuation is `y = -x + 2` — `(4, -2)` → `(-2, 4)`. The old pair traced `y = -x`, which only touches the tile at `(0, 0)`.

js/packages/ui/src/components/data/histogramOverlap.ts
kentwelcome: Applied in `18774675`, with one deviation: the `chart.chartArea` destructure is hoisted out of the loop since it is loop-invariant. The clamp arithmetic is exactly as suggested.

`histogramOverlap.ts:216-234`:

```ts
// afterDatasetsDraw runs outside chart.js's dataset clip, and bar geometry
// can sit far off-scale (chart.js reports -32768 for unplottable values),
// so every rect is clamped to the plot area before it is painted.
const { left, top, right, bottom } = chart.chartArea;

for (let index = 0; index < length; index += 1) {
  ...
  const x0 = Math.max(overlap.left, left);
  const y0 = Math.max(overlap.top, top);
  const x1 = Math.min(overlap.left + overlap.width, right);
  const y1 = Math.min(overlap.top + overlap.height, bottom);
  if (x1 <= x0 || y1 <= y0) continue;
  ctx.fillRect(x0, y0, x1 - x0, y1 - y0);
}
```

Two regression tests in `histogramOverlap.test.ts`:

- **"does not paint chart.js off-scale bar geometry over the plot area"** — feeds your exact sentinel geometry (`{x: -32768, y: -32768, width: 517.3, base: 170.8}`). `intersectBars` still returns the ~33,000px rect, the clamp yields `x1 = -32509.35 <= x0 = 0`, and `fillRect` is never called.
- **"clamps the overlap rect to the plot area"** — a bar that is on-canvas in x but overflows the plot area on all four sides now paints `fillRect(0, 0, 400, 200)` against a `chartArea` of `{left: 0, top: 0, right: 400, bottom: 200}`, so the clamp is exercised in both directions rather than only proving the rect is dropped.

I left `readBarGeometry`'s `Number.isFinite` check alone. The clamp is the general guard — it also covers a future scale `min`/`max` that puts real, finite bar geometry outside the plot area, which a sentinel-specific check would not.

kentwelcome:
## Review response — 4bd515c1 → 18774675

Thanks, this was a good catch list. Four of the five findings held up and are fixed. Finding 5's evidence does not hold, but the assertion you asked for is in anyway, along with the coupling that genuinely was not covered.

| # | Finding | Assessment | Action | Commit |
|---|---------|------------|--------|--------|
| 1 | Overlap plugin captured at mount, crosshatch keeps stale palette | **Valid** | Stateless plugin; palette moved into `options.plugins.histogramOverlap` | `18774675` |
| 2 | Two crosshatch strokes paint nothing | **Valid** | Both inline suggestions applied | `18774675` |
| 3 | `-32768` sentinel yields a ~33,000px unclipped rect | **Valid** | Rect clamped to `chart.chartArea` + 2 regression tests | `18774675` |
| 4 | Contrast test uses its own `compositeHex` | **Valid** | One shared `compositeRgbChannels` in the theme module | `f95aa156` |
| 5 | Dataset order not pinned by any test | **Partly invalid** — order *was* pinned; the index↔label coupling was not | Requested assertion added, plus a test tying legend `datasetIndex` back to the dataset label | `18774675` |

---

### 1 — Blocker: theme-following overlap painter

Confirmed against `react-chartjs-2@5.3.1/dist/index.js`. `plugins` is destructured at line 79 and read only inside `renderChart()` (line 85), which is called from the mount-only `useEffect(..., [])` and the `[type]` effect. No other effect lists it. So the live chart kept the mount-time plugin instance and its closed-over palette.

Fixed the way you suggested. `histogramOverlap.ts` now exports a single stateless `histogramOverlapPlugin`, and the palette rides in `options.plugins.histogramOverlap`:

```ts
export const histogramOverlapPlugin: Plugin<"bar", HistogramOverlapPluginOptions> = {
  id: "histogramOverlap",
  afterDatasetsDraw(chart, _args, options) {
    const palette = options?.palette;
    if (!palette) return;
    ...
```

`HistogramChart.tsx:227-232` passes `histogramOverlap: { palette: overlapPalette }`, and `plugins={[histogramOverlapPlugin]}` is now a stable module-level reference, so the mount-only read is harmless.

Why the new path does refresh, traced through chart.js 4.5.1:

- `setOptions` does `Object.assign(chart.options, nextOptions)`, so `options.plugins` is replaced wholesale on every options change.
- `Chart.update()` (`chart.js:5913`) calls `config.update()` → `clearCache()` (drops `_scopeCache` and `_resolverCache`) and then `this._plugins.invalidate()` (`chart.js:5921`).
- `_createDescriptors` therefore re-reads `config.options.plugins` and rebuilds the resolver via `pluginOpts` → `config.createResolver(...)`.

I also checked the one thing that worried me about routing a nested object through chart.js's option resolver — `needsSubResolver` (`helpers.dataset.js:1810`) turns plain objects into sub-resolvers, so I wanted proof the nested reads survive. Driving `_createResolver` directly with this palette shape:

```
palette truthy: true
canvasBackground: #FAFAFA
overlap.chartFill: #805AD5A5
overlap.foreground: #553C9A
base.border: #B
after theme flip: #171923 #B794F4A5
missing palette -> undefined
```

So nested string leaves resolve, a replaced palette resolves to the new values, and an absent plugin options block resolves to `undefined` — which is what the `if (!palette) return;` guard is written against.

Your second-order point about the pattern cache resolves with it: the painter and `generateLabels` now build their key from the same palette object, so the per-context single-slot cache stops alternating between a stale and a fresh key.

Two tests guard the mechanism directly (`HistogramChart.test.tsx`):

- **"reuses one painter instance across theme switches"** — renders light, re-renders dark, asserts `plugins[0]` is the same object. This is the invariant that makes the mount-only read safe; it fails if anyone reintroduces a per-render plugin.
- **"hands the light/dark palette to the painter through chart options"** — asserts `options.plugins.histogramOverlap.palette` matches `getSemanticColorTheme(isDark)` for both themes. This is what would have failed on the old code.

### 4 — Shared composite math

Split into `compositeRgbChannels` (the arithmetic) and `compositeHex` (the formatting) in `theme/semanticColors.ts`, both exported. `histogramOverlap.ts` uses `compositeHex`; the test uses `compositeRgbChannels`.

The reason they are not one function: `js/CLAUDE.md` requires space-separated `rgb()` for colors, and MUI's `decomposeColor` splits `rgb()` on commas for `type === "rgb"` (`@mui/system/colorManipulator`), so `getContrastRatio("rgb(171 146 226)", …)` parses as NaN and the contrast assertions would pass vacuously. Splitting the math from the separator keeps the convention and still makes drift in the arithmetic fail the test. The test carries a comment saying exactly that.

`histogramOverlap.ts` imports from `../../theme/semanticColors` rather than the `../../theme` barrel, because the barrel re-exports from `components/data/agGridTheme` and this file lives in `components/data`.

### 5 — Dataset ordering

The claim that "swapping the two `buildDataset` calls would invert every legend swatch and every click-to-hide target with all 4134 tests still green" does not hold. `HistogramChart.test.tsx:236-241` ("creates datasets with correct labels") already asserted `datasets[0].label === "Current"` and `datasets[1].label === "Base"`. I swapped the two calls on the reviewed tree and ran it:

```
Expected: "Current"
Received: "Base"
 ❯ packages/ui/src/components/data/__tests__/HistogramChart.test.tsx:239:38
```

`HistogramChart.test.tsx:259` is the `it.each` "co-locates source datasets" case, which does only assert `toHaveLength(2)` — but it is not the only test covering ordering.

That said, there was a real hole next to it: nothing tied the *dataset index* to the label, so `CURRENT_DATASET_INDEX` / `BASE_DATASET_INDEX` drifting relative to the data was uncovered. New test "pins the dataset order the painter and legend index into" does both — your requested `data.datasets.map((d) => d.label)` equals `["Current", "Base"]`, plus every legend entry that carries a `datasetIndex` must point at the dataset whose label matches its text.

### Notes

- `histogramOverlap.ts:83` (uncached fallback when `getContext("2d")` returns null) — acknowledged, not changed. It is a real allocation-per-draw, but only on a canvas-less context where nothing paints anyway, and it is outside what the blockers touch. Happy to fix it here if you would rather not leave it.
- `chartjs-adapter-date-fns` not imported by the shipped app — agreed, pre-existing and outside this diff. Worth its own issue; the clamp in finding 3 means the missing adapter no longer has a way to paint over the axes.

### Verification

From `js/`, Node 26.7.0:

```
$ pnpm lint:fix
$ biome check --write --diagnostic-level=error
Checked 683 files in 673ms. No fixes applied.

$ pnpm type:check
$ tsc --noEmit   (root, @datarecce/ui, @datarecce/storybook — all clean)

$ pnpm test
 Test Files  196 passed (196)
      Tests  4141 passed | 5 skipped (4146)
```

4134 → 4141 passed, i.e. 7 new tests, no snapshot churn.

kentwelcome (approved):
Code Review Loop: blocking findings fixed and verified at 18774675.

PR #1516, opened 2026-08-24
fix(schema): show column names and types in the checklist schema diff

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

fix

In the checklist, a schema diff result was unreadable. The table had one column, "Name", and each cell showed text like `order_id|false|false|false`. The data type icon was missing.

Regression range: 1.44.0 is correct, 1.44.1 and every later version show the fault. It came from `b88b71b2`, which moved three callers onto `/api/models/` but re-added `resource_type` in `NodeViewOss` only.

`SchemaSummary` gets the same fix, but no route in this repo renders it. It is exported from `@datarecce/ui`, so the cloud app may render it.

Tests: `SchemaDiffView.test.tsx` and `SchemaSummary.test.tsx` each mount the view with a mocked `SchemaView` that reports back the `resource_type` it received. `SchemaDiffView` is covered for all three arm shapes a real diff produces — modified (both arms), added (empty `base`), removed (empty `current`) — so a fix applied to only one arm still fails. Both were verified red against the unfixed component.

**Reactivated behavior — please read.** `resource_type` is what `SchemaView` gates `profileNode` on (`SchemaView.tsx:263`), so restoring it does more than repaint the Name column. `profileNode` has been `undefined` in these two views since `b88b71b2` (2026-04-20), which is *before* inline profiling landed in `c458fc6d` (2026-05-29) — so these code paths have never actually run here. Supplying it turns two of them on for the first time:

1. **Inline profile distribution.** `SchemaView` passes `model: profileNode?.name` to `useInlineProfileDistribution`, whose `queryEnabled = flagEnabled && enabled && !!model` (`useInlineProfileDistribution.ts:138`) was pinned `false` by the missing name. It now evaluates normally. It still requires **both** server flags — `inline_profile` (the hook's own gate) and `new_cll_experience` (`selectInlineProfileScope` returns `profileEnabled: false` outright when it is off) — plus at least one changed column. With both flags on, each rendered schema table submits a synchronous `POST /api/runs` `profile_distribution` run scoped to its changed columns. In `SchemaSummary` that is one run **per card**, and the OSS fetch client sets no request timeout, so a slow warehouse adapter holds those requests open with no client-side abort. `SchemaDiffView` renders one table at a time and passes no `columnChanges` for added/removed nodes, so it is the milder of the two.
2. **Per-column kebab menu in `SchemaSummary`.** `SchemaSummary` does not pass `showMenu`, which defaults to `true` (`SchemaView.tsx:228`), so the column menu now renders in the summary cards. Rendered outside a `RecceActionProvider` its items fall back to `runAction: () => void 0` (`RecceActionContext.tsx:58`) — visible, clickable, silent. `SchemaDiffView` passes `showMenu={false}` and is unaffected.

Both are the pre-`b88b71b2` intended behavior of a correctly-populated `SchemaView`, not new features, and neither is reachable in this repo today (no route renders `SchemaSummary`; the flags are off by default). Flagging them so the cloud app and the release notes are not surprised. If the summary cards should stay menu-less or profile-free, `showMenu={false}` / a `SchemaView` opt-out is a one-line follow-up — say the word and I will add it here.

Yes. Schema diff results in the checklist render the Name column correctly again (column name plus data type icon) instead of raw `order_id|false|false|false` text.

Additionally, where the `inline_profile` and `new_cll_experience` server flags are both enabled, the schema-diff check view and the schema summary cards now show the inline paired column distribution, and the summary cards now show the per-column menu. Both were unreachable in these two views because of the same missing `resource_type`; see the reviewer notes above.


gcko:
# Code Review: PR #1516

**SHA** `6b2de1bd` · **Verdict** NO-GO

The fix mechanism is correct and I traced it end to end: `toSchemaDataGrid.ts:175` sets `cellRenderer: node ? … : undefined`, so when `node` (i.e. `profileNode`) is undefined ag-grid falls back to rendering the raw `valueGetter` string — which is literally `` `${row.name}|${definitionChanged}|${isImpacted}|${changeUnknown}` ``, matching the reported `order_id|false|false|false`. `DataTypeIcon` lives inside `ColumnNameCell`, so the missing icon has the same cause. `profileNode` needs `resource_type` (`SchemaView.tsx:263`), and the OSS backend genuinely omits it — `recce/adapter/dbt_adapter/__init__.py:541` returns only `columns`, `primary_key`, `raw_code`. Both the `base` and `current` arms are covered, which matters because an added node gets `base = {}` and a removed node gets `current = {}`.

Two things block.

## Issues

1. `js/packages/ui/src/components/check/SchemaDiffView.tsx:267` — The regression this PR exists to fix is left with no test; the one test added covers the component the PR body says nothing renders.
   Evidence: the only new test is `js/src/components/summary/__tests__/SchemaSummary.test.tsx:178`, for `SchemaSummary` — which the description states "no route in this repo renders". No test references `SchemaDiffView` anywhere (`rg -l SchemaDiffView --glob '*.test.*'` → empty), while `packages/ui/src/components/check/__tests__/` already holds four test files, so the pattern is available. `b88b71b2` broke both sites by one mechanical swap (`node.data.data.base` → `/api/models/`); the identical refactor recurs against a green suite.
   Pass E.

2. `js/packages/ui/src/components/summary/SchemaSummary.tsx:67` and `check/SchemaDiffView.tsx:267` — Supplying `resource_type` also switches on inline profile-distribution runs in these two views for the first time ever. Not mentioned in the description, not covered by a test.
   Evidence: `SchemaView.tsx:263` gates `profileNode` on `resource_type ∈ ["model","seed","snapshot","source"]`; `SchemaView.tsx:327` passes `model: profileNode?.name`; `useInlineProfileDistribution.ts:138` computes `queryEnabled = flagEnabled && enabled && !!model`. Inline profiling landed in `c458fc6d` (2026-05-29) — *after* the regression `b88b71b2` (2026-04-20) — so `!!model` has been `false` in both views for the entire life of the feature. With `inline_profile` on, `SchemaSummary` now submits one synchronous `POST /api/runs` per card, and that hook's own docstring notes "the OSS fetch client sets no request timeout … a slow remote-warehouse adapter would hold this request open with no client-side abort — there is NO timeout error or Sentry capture for that case today."
   Pass F.

## Notes

1. `SchemaSummary.tsx:67` — `showMenu` is not passed, so it defaults to `true` (`SchemaView.tsx:228`), and the per-column kebab menu renders in the summary cards for the first time. Outside a `RecceActionProvider` its items resolve to `runAction: () => void 0` (`contexts/action/RecceActionContext.tsx:58`) — visible, clickable, silent. `SchemaDiffView` passes `showMenu={false}` and is unaffected.

2. Both reconstructions omit `package_name`, which the reference site passes (`NodeViewOss.tsx:357,367`). Nothing in the schema render path reads it today — only `model.name` (`ColumnNameCell.tsx:157`) and `model.resource_type` (`:353`) — so it is inert, but the same object is now assembled in two shapes.

3. `toSchemaDataGrid.ts:175` gates the *Name cell renderer* on `node`, an object that exists to describe profiling eligibility. That coupling is the reason a missing `resource_type` degrades to raw text rather than just disabling a menu. `SchemaDiffView.tsx:154-159` filters to exactly the profiling allowlist so it is unreachable there; `SchemaSummary`'s `listChangedNodes` applies no resource-type filter.

## Verification

`pnpm lint`, `pnpm type:check`, `pnpm test` (197 files, 4151 passed / 5 skipped) and `pnpm run build` all pass at `6b2de1bd`. Base is fresh (merge-base == `origin/main`), merge is clean, no file overlap with commits landed since divergence. `@datarecce/ui` resolves to source in `vitest.config.mts:139-146`, so the new test does exercise the change: without `resource_type` the mock renders no `data-base-resource-type` attribute and the assertion fails — it is not vacuous.

gcko (changes requested):
Code Review Loop: blocking findings at `6b2de1bd`. See the review comment above.

gcko:
✅ **Code Review Loop — fixes pushed** (`6ef49648`)

- **Issue 1** (no test for the regression's own file) — fixed in `6ef49648`. New `packages/ui/src/components/check/__tests__/SchemaDiffView.test.tsx` covers all three arm shapes (modified / added-empty-base / removed-empty-current). Independently verified red: deleting both `resource_type:` lines fails all three cases. The node is deliberately a `snapshot`, not a `model`, so the assertion cannot pass off a downstream default.
- **Issue 2** (undisclosed reactivation) — addressed by the PR description rewrite. It documents the inline profile-distribution and kebab-menu activation, and adds the `new_cll_experience` gate (`selectInlineProfileScope.ts:80`) that the review had missed — so the activation needs **both** flags, not just `inline_profile`. Verified accurate.

Notes 1–3 were left to the author's judgement, as intended.

Re-verified at `6ef49648`: `pnpm lint`, `pnpm type:check`, `pnpm test` (198 files, 4154 passed / 5 skipped — up 3 from the reviewed SHA). Review state updated to approve below.

gcko:
Thanks — both Issues are addressed in `6ef49648` plus a PR description rewrite. Detail per finding:

**Issue 1 — `SchemaDiffView` had no test.** Valid, and the sharper form of it is right: the test I added covered the component the description itself says nothing renders. Added `js/packages/ui/src/components/check/__tests__/SchemaDiffView.test.tsx`, following the relative-import/`vi.mock` pattern of the existing four files in that directory. It mounts `SchemaDiffView` with a mocked `SchemaView` that reports `base?.resource_type` / `current?.resource_type` back as data attributes, same assertion shape as the `SchemaSummary` test.

Three cases rather than one, because you flagged that both arms matter: modified (both arms populated), added (`base = {}` → base arm absent, current arm still typed), removed (`current = {}` → the mirror). A fix applied to only one arm fails two of the three.

Verified red, not vacuous: deleting the two `resource_type: node.data.resourceType` lines from `SchemaDiffView.tsx` fails all three (`Received: null` for the attribute), restoring them passes. Full suite is 198 files, 4154 passed / 5 skipped.

**Issue 2 — undisclosed reactivation.** Valid; I traced it and reached the same conclusion, with one correction. The description now has a "Reactivated behavior — please read" section and a filled-in "Does this PR introduce a user-facing change?" covering both the inline distribution and the `SchemaSummary` column menu.

The correction: inline profiling needs **both** flags, not just `inline_profile`. `selectInlineProfileScope` returns `profileEnabled: false` unconditionally when `new_cll_experience` is off (`selectInlineProfileScope.ts:80-88`), and that value is what `SchemaView` passes as the hook's `enabled` (`SchemaView.tsx:330`), which `queryEnabled = flagEnabled && enabled && !!model` then ANDs with the `inline_profile` gate. It also needs a non-empty changed-column scope. So the trigger is `inline_profile` ∧ `new_cll_experience` ∧ ≥1 changed column — narrower than "with `inline_profile` on", though the per-card fan-out you describe is exactly right once both are on, and `SchemaSummary`'s `listChangedNodes` guarantees every card it renders clears the changed-column condition.

I did not change the activation itself. It is the pre-`b88b71b2` intended behavior of a correctly-populated `SchemaView` rather than something this PR invents, both flags are off by default, and no route in this repo renders `SchemaSummary` — so gating it here would be a product decision made blind to the cloud app's intent. Disclosure felt like the honest fix. If you would rather the summary cards stay profile-free and menu-less, `showMenu={false}` plus a `SchemaView` opt-out is a one-line follow-up and I will add it to this PR on request.

**Notes 1–3:** Note 1 (`showMenu` defaulting to `true` in `SchemaSummary`) is now disclosed as part of Issue 2 but deliberately not changed, for the reason above. Notes 2 (`package_name` omitted from both reconstructions — inert; nothing in the schema render path reads it) and 3 (`toSchemaDataGrid` coupling the Name cell renderer to the profiling-eligibility object) are both accurate and both left alone: the first is a no-op today and the second is a refactor of a shared helper that would widen this three-line fix well past its blast radius. Note 3 in particular reads like the real root cause worth a separate ticket — happy to file one.

gcko (approved):
Code Review Loop: blocking findings fixed and verified at `6ef49648`. SchemaDiffView regression test independently verified red without the fix; the reactivation is now disclosed in the PR description.

gcko:
CI fix pushed: `e742ca12`.

`wire-enum-lint` failed on the test fixture I added in `6ef49648` — `change.category: "non_breaking"` trips the DRC-3553 guard for new lines under `js/packages/ui/src/`. The check's suggested remedy does not apply: `LineageGraphNode` types `change.category` as the legacy union only, so `model_wide`/`column`/`additive` are not assignable there. Used the `// wire-enum-ok` marker the script documents for this case, with a comment recording why. The category is inert for the test regardless — `SchemaDiffView` reads `change.columns` and never the category.

All 8 checks pass at `e742ca12`; the approval above (cited at `6ef49648`) still stands.

PR #1517, opened 2026-08-25
feat(profile): polish headers and add explicit delta modes

gcko:
## Stack

1. This PR: Profile readability and delta modes
2. #1518: Histogram readability and one-step launch

Merge this PR first. GitHub Stack #1519 tracks the dependency chain.

## Tickets

- [DRC-2833](https://linear.app/recce/issue/DRC-2833)
- [DRC-2866](https://linear.app/recce/issue/DRC-2866)

## Summary

- Humanizes Profile statistic headers while preserving raw field names and primary-key identity.
- Adds explicit percentage-point and percent-change modes.
- Keeps every legacy delta mode and persisted value backward compatible.
- Handles zero, negative, structural, modified, overflow, and tooltip cases explicitly.

## Verification

- Frontend lint passed.
- Root, UI package, and Storybook type checks passed.
- 198 frontend test files passed: 4,212 tests passed and 5 skipped.
- Production frontend build passed.
- Adversarial implementation and whole-layer reviews are clean.

## Compatibility

The new delta modes are additive. Existing saved checks and legacy mode values retain their prior semantics.

### DIFF
--- a/js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx
+++ b/js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx
@@ -34,6 +34,7 @@ interface RecceColumnContext {
   columnType?: ColumnType;
   columnRenderMode?: ColumnRenderMode;
   showStructuralIndicator?: boolean;
+  enableProfileDiffPercentModes?: boolean;
 }
 
 /**
@@ -61,6 +62,48 @@ export interface InlineRenderCellConfig {
    * (e.g., toast notifications on copy).
    */
   DiffTextComponent?: ComponentType<InlineDiffTextProps>;
+}
+
+/** Parses only finite numbers and complete numeric strings for explicit modes. */
+export function parseFiniteNumeric(value: RowDataTypes): number | undefined {
+  if (typeof value === "number")
+    return Number.isFinite(value) ? value : undefined;
+  if (typeof value !== "string" || value.trim() === "") return undefined;
+  const parsed = Number(value);
+  return Number.isFinite(parsed) ? parsed : undefined;
+}
+
+function formatPercent(value: number): string {
+  return new Intl.NumberFormat("en-US", {
+    style: "percent",
+    maximumFractionDigits: 2,
+  }).format(value);
+}
+
+function formatSignedChange(value: number, suffix: string): string {
+  return `${value >= 0 ? "+" : ""}${formatSmartDecimal(value)}${suffix}`;
+}
+
+function renderUndefinedRelativeChange(base: number, current: number) {
+  const tooltipText = `Base: ${base}\nCurrent: ${current}\nChange: N/A`;
+
+  return (
+    <Tooltip
+      title={tooltipText}
+      slotProps={{
+        tooltip: { sx: { whiteSpace: "pre-line" } },
+      }}
+      enterDelay={300}
+      placement="top"
+    >
+      <Typography
+        data-direction="equal"
+        sx={{ color: "text.secondary", fontSize: "0.75rem" }}
+      >
+        N/A
+      </Typography>
+    </Tooltip>
+  );
 }
 
 /**
@@ -91,6 +134,8 @@ export function createInlineRenderCell(config: InlineRenderCellConfig = {}) {
     const colDef = params.colDef as ColDefWithMetadata;
     const columnType = colDef?.context?.columnType;
     const columnRenderMode = colDef?.context?.columnRenderMode;
+    const enableProfileDiffPercentModes =
+      colDef?.context?.enableProfileDiffPercentModes === true;
     const columnKey = colDef?.field ?? "";
 
     if (!params.data) {
@@ -123,6 +168,13 @@ export function createInlineRenderCell(config: InlineRenderCellConfig = {}) {
       columnRenderMode,
     );
 
+    const isExplicitPercentMode =
+      enableProfileDiffPercentModes &&
+      (columnRenderMode === "percent_delta" ||
+        columnRenderMode === "percent_change");
+    const baseNumericValue = parseFiniteNumeric(row[baseKey]);
+    const currentNumericValue = parseFiniteNumeric(row[currentKey]);
+
     // No change - render single value.
     // Must use the same type-dispatched comparison the row status and the
     // side-by-side cell classes use: a raw `===` here would render a
@@ -132,12 +184,27 @@ export function createInlineRenderCell(config: InlineRenderCellConfig = {}) {
     // query-diff views, so that mismatch was the whole reported symptom of
     // DRC-3025.
     if (!isCellChanged(row[baseKey], row[currentKey], columnType)) {
+      if (
+        isExplicitPercentMode &&
+        columnRenderMode === "percent_change" &&
+        baseNumericValue !== undefined &&
+        currentNumericValue !== undefined &&
+        baseNumericValue < 0
+      ) {
+        return renderUndefinedRelativeChange(
+          baseNumericValue,
+          currentNumericValue,
+        );
+      }
+
       return (
         <Typography
           component="span"
           style={{ color: currentGrayOut ? "gray" : "inherit" }}
         >
-          {currentValue}
+          {isExplicitPercentMode && currentNumericValue !== undefined
+            ? formatPercent(currentNumericValue)
+            : currentValue}
         </Typography>
       );
     }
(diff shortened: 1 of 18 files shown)


kentwelcome:
# Code Review: PR #1517

**SHA** `a8b8ad16` · **Verdict** NO-GO

## Blockers

1. `js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx:116` — The zero-change tooltip is a JSX string-literal attribute, so its `\n` renders as the two literal characters `\n`, not line breaks.
   Evidence: transpiling `title="Base: 0\nCurrent: 0\nChange: 0"` emits `title: "Base: 0\\nCurrent: 0\\nChange: 0"` (escaped backslash), while the template-literal tooltips at `:92`, `:318` and `:403` emit real newlines. `whiteSpace: "pre-line"` at `:118` therefore has nothing to break on, and the user sees one line of `Base: 0\nCurrent: 0\nChange: 0`.
   Pass A.

2. `js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx:91-108`, `:255`, `:267`, `:380` — `renderUndefinedRelativeChange` and `renderZeroRelativeChange` return the change indicator *as the entire cell*, so the column's actual value is not rendered at all.
   Evidence: with `percent_change` and `baseNum <= 0`, `:379-381` returns only `<Typography>N/A</Typography>`; `inlineRenderCell.test.tsx:361-372` asserts `0 → 0.2` renders `"N/A"` and never asserts `0.2` is present. `avg` and `median` are FLOAT profile stats (`recce/tasks/profile.py:28-33`) that go negative, and `row_count` legitimately goes `0 → N` — the most interesting rows in a Profile Diff are exactly the ones whose value disappears. The sibling undefined-change fallback 6 lines below at `:387-396` returns `renderInlineValues(...)`, which *does* show both values, so the same function hides the value in one undefined case and shows it in the other.
   Pass A.

## Issues

3. `js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx:362-371` — In `percent_delta` mode, added/removed rows render the proportion as a raw decimal while every other row in the same column renders it as a percentage.
   Evidence: an added row has no `base__` key, so `hasBase` is false and the guard at `:367` skips the percent block; control falls to `renderInlineValues` at `:448`, whose `currentValue` comes from `toRenderedValue(..., "percent_delta")` → `columnRenderedValue` (`gridUtils.ts:449-483`) has no `percent_delta` branch → `formatSmartDecimal(0.94, 2)` = `"0.94"`. The unchanged branch at `:279-281` formats the identical value as `"94%"`. A newly added model column then reads `0.94` in a column of `94%` values — a 100× misread.
   Pass A.

4. `js/packages/ui/src/components/ui/dataGrid/__tests__/inlineRenderCell.test.tsx:328-331` (and `createDataGrid.profile.test.tsx:511-514`) — The tooltip assertions cannot fail on Blocker 1.
   Evidence: `/Base: 0/`, `/Current: 0/` and `/Change: 0/` are substring regexes; all three match the single unbroken string `Base: 0\nCurrent: 0\nChange: 0` just as they match the three-line version. There is also no `percent_delta` + `__status: "added"/"removed"` case anywhere — `:394-406` covers only `percent_change` — which is why Issue 3 shipped.
   Pass E.

## Notes

5. `inlineRenderCell.tsx:403-406` — The new tooltips print unrounded IEEE-754 values: a `row_count` of 100 → 129 shows `Change: +28.999999999999996%` beside a cell reading `(+29%)`. Deliberate and asserted at `inlineRenderCell.test.tsx:417`, but the legacy delta tooltip rounds its percent with `.toFixed(2)` (`:320-322`), so the two tooltip styles in one file now disagree.

6. `dataGridFactory.tsx:425-433` — `profileDiffPercentMode` hardcodes `["distinct_proportion", "not_null_proportion"]`. That matches `recce/tasks/profile.py:150-165` exactly today, but nothing ties the two lists together; a future `*_proportion` stat silently gets relative-percent semantics.

## Verification

- `pnpm lint` — clean (686 files).
- `pnpm type:check` — clean (root, `@datarecce/ui`, storybook).
- `pnpm test` — 198 files, 4,212 passed / 5 skipped.
- Base freshness: `origin/main` is the merge base; `git merge-tree` clean; no file overlap with commits landed since divergence.

## Limits

Pass B ran as a targeted backstop over the diff (no `dangerouslySetInnerHTML`, `innerHTML`, `eval`, `new Function`, network calls, or storage writes introduced; `displayName` is derived from schema column keys and rendered as escaped React text) rather than a full `security-review` invocation. The diff is presentation-layer number formatting with no trust boundary, so the scoped check is the proportionate one — flagging it here rather than claiming a pass I did not run.

js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx
```diff
@@ -63,6 +64,109 @@ export interface InlineRenderCellConfig {
...
+        N/A
+      </Typography>
+    </Tooltip>
+  );
+}
+
+function renderZeroRelativeChange() {
+  return (
+    <Tooltip
+      title="Base: 0\nCurrent: 0\nChange: 0"
```
kentwelcome: The zero-change tooltip is a JSX string-literal attribute, so its `\n` renders as the two literal characters `\n`, not line breaks. Wrapping it in braces makes it a JS string literal, where `\n` is a real escape.

```suggestion
      title={"Base: 0\nCurrent: 0\nChange: 0"}
```

kentwelcome (changes requested):

kentwelcome:
✅ **Code Review Loop — fixes pushed** (`279516dd`)

All four blocking findings re-verified against the pushed HEAD by reading the fix diff, not by taking the fix agent's word for it.

- **Blocker 1** (tooltip `\n` rendered literally) — fixed. Both indicator tooltips now build their text as a JS string argument to `renderChangeIndicator`, where `\n` is a real escape.
- **Blocker 2** (`N/A` / `0%` replaced the whole cell) — fixed. `renderChangeIndicator` takes the cell's values and renders the indicator beside them: current alone on unchanged rows, base and current on changed rows.
- **Issue 3** (`percent_delta` added/removed rows rendered as raw decimals) — fixed. `displayBaseValue` / `displayCurrentValue` apply the column's percentage unit for every row status, including the added/removed fall-through.
- **Issue 4** (tooltip assertions could not fail) — fixed. `openTooltipText` compares exact `textContent`, which discriminates a real three-line tooltip from the literal `\n` string. Added `percent_delta` coverage for `added` and `removed` at both the renderer and the grid factory.

Notes 5 and 6 were non-blocking and are correctly untouched.

Verified at `279516dd`: `pnpm lint` clean (686 files), all three `type:check` targets clean, `pnpm test` 198 files / 4,221 passed / 5 skipped (up 9 tests from the reviewed SHA).

js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx
kentwelcome: @kentwelcome Fixed in `279516dd`. Took the fix rather than the literal suggestion, because the same function also had to stop replacing the cell's value (Blocker 2): both indicator tooltips now build their text as a JS string passed into `renderChangeIndicator` (`js/packages/ui/src/components/ui/dataGrid/inlineRenderCell.tsx:146` for the zero case, `:139` for `N/A`), so the `\n` is a real escape.

The assertion that would have caught this is now exact rather than a substring regex — `inlineRenderCell.test.tsx` compares `tooltip.textContent` to `"Base: 0\nCurrent: 0\nChange: 0"`. Verified by stashing the source fix and re-running: the tightened tests fail on the pre-fix code, pass after.

kentwelcome:
## Review response — all four findings addressed in `279516dd`

Both blockers and both issues are fixed. The two Notes are left as-is.

### Blocker 1 — literal `\n` in the zero-change tooltip

Fixed. `renderZeroRelativeChange` no longer passes the tooltip as a JSX string-literal attribute. Both indicator tooltips now build their text as a JS string handed to a shared `renderChangeIndicator` helper — `inlineRenderCell.tsx:146` (`"Base: 0\nCurrent: 0\nChange: 0"`) and `:139` (the `N/A` template literal). Real newlines, so `whiteSpace: "pre-line"` at `:105` has something to break on.

### Blocker 2 — the indicator replaced the cell's value

Fixed. `renderUndefinedRelativeChange` and `renderZeroRelativeChange` now take the values as a parameter and render them beside the indicator, in the same `Tooltip > Box` shape as the normal changed-value render:

- Unchanged rows (`:294`, `:306`) show the current value, matching what the unchanged branch below them shows.
- The changed `baseNum <= 0` row (`:409`) shows base **and** current via `renderInlineValues`, matching the sibling non-finite fallback you cited at `:431-437`.

A `0 → 0.2` `row_count` now reads `0 0.2 N/A` instead of a bare `N/A`. The indicator strings (`N/A`, `0%`) and the tooltips are unchanged.

### Issue 3 — `percent_delta` unit inconsistent across row statuses

Fixed, but not in `columnRenderedValue`. Adding a `percent_delta` branch there would also change the deliberate mismatch fallback — a `percent_delta` mode persisted onto `row_count` would start rendering `1,200%`, which `createDataGrid.profile.test.tsx` asserts should stay `12`. So the unit is applied in the renderer instead, where `isExplicitPercentMode` has already confirmed the mode matches the column: `inlineRenderCell.tsx:257-266` derives `displayBaseValue` / `displayCurrentValue`, and all four render paths use them (`:416-417`, `:433-434`, `:468`, `:489-490`). Added and removed rows now render `94%` like every other row in the column. If you'd rather the root cause move into `gridUtils.ts`, say so and I'll do it together with the fallback semantics.

### Issue 4 — tests could not fail on Blocker 1

Fixed, and verified negatively. Tooltip assertions now compare exact `textContent` instead of substring regexes (`inlineRenderCell.test.tsx`, `createDataGrid.profile.test.tsx`). Added coverage for `percent_delta` with `__status: "added"` / `"removed"` at both levels — the renderer directly (bare current value, and a null-base placeholder row) and through `createDataGrid` with a real structural row.

Evidence that the new tests bite: stashing only `inlineRenderCell.tsx` and re-running the two files fails **11** tests, including `renders zero-to-zero percent_change through the actual factory cell`. Restored, all pass.

### Notes 5 and 6

Left unchanged, as agreed — non-blocking.

### Verification

- `pnpm lint:fix` — clean, 686 files.
- `pnpm type:check` — clean (root, `@datarecce/ui`, storybook).
- `pnpm test` — 198 files, 4,221 passed / 5 skipped (was 4,212; +9 new tests).

kentwelcome (approved):
Code Review Loop: blocking findings fixed and verified at 279516dd.

PR #1518, opened 2026-08-25
feat(histogram): polish bin geometry and add one-step launch

gcko:
## Stack

1. #1517: Profile readability and delta modes
2. This PR: Histogram readability and one-step launch

This PR is based on #1517. Merge the stack bottom-up. GitHub Stack #1519 tracks the dependency chain.

## Tickets

- [DRC-2838](https://linear.app/recce/issue/DRC-2838)
- [DRC-2848](https://linear.app/recce/issue/DRC-2848)

## Summary

- Adds Decimal-safe nice numeric bins, exact zero anchoring, full-domain coverage, and exact internal and terminal boundary ownership.
- Renders literal edge ticks, variable-width bars, domain-aware labels, accurate tooltip hit testing, and stable animated geometry.
- Launches a model-level Histogram run exactly once after an explicit eligible column selection.
- Preserves the generic Execute path and the existing schema-column and lineage-column direct launch paths.
- Requires columns to exist in both environments before one-step submission and records distinct launcher telemetry sources.

## Verification

- Frontend lint-fix passed across 692 files with no changes.
- Root, UI package, and Storybook type checks passed.
- 204 frontend test files passed: 4,243 tests passed and 5 skipped.
- Production frontend build passed with no tracked generated changes.
- Python suite passed: 1,630 tests passed and 5 skipped.
- Focused backend Histogram suite passed: 45 tests.
- Final adversarial whole-layer review: Ready to merge.

## Compatibility and follow-up

Existing Histogram wire fields and direct launch behavior remain compatible. Large num_bins SQL growth is tracked in [DRC-4232](https://linear.app/recce/issue/DRC-4232); this stack intentionally avoids an incompatible hard cap until adapter-specific exact bucketing is proven.

### DIFF
--- a/js/packages/ui/src/components/data/HistogramChart.tsx
+++ b/js/packages/ui/src/components/data/HistogramChart.tsx
@@ -92,6 +92,8 @@ function formatBinRange(binEdges: number[], index: number): string {
   return `${formatAsAbbreviatedNumber(start)} - ${formatAsAbbreviatedNumber(end)}`;
 }
 
+const MAX_EDGE_TICK_LABELS = 8;
+
 /**
  * HistogramChart Component
  *
@@ -168,10 +170,6 @@ function HistogramChartComponent({
 
   // Build chart data
   const chartData = useMemo<ChartData<"bar">>(() => {
-    const labels = binEdges
-      .slice(0, -1)
-      .map((_, i) => formatBinRange(binEdges, i));
-
     const buildDataset = (
       data: HistogramDataset,
       label: string,
@@ -197,7 +195,8 @@ function HistogramChartComponent({
     };
 
     return {
-      labels,
+      // Labels represent bin edges; data remains one value per bin.
+      labels: binEdges,
       datasets: [
         buildDataset(
           currentData,
@@ -212,10 +211,10 @@ function HistogramChartComponent({
   // Build chart options
   const chartOptions = useMemo<ChartOptions<"bar">>(() => {
     const maxCount = Math.max(...currentData.counts, ...baseData.counts);
-
-    const labels = binEdges
-      .slice(0, -1)
-      .map((_, i) => formatBinRange(binEdges, i));
+    const edgeTickInterval = Math.max(
+      1,
+      Math.ceil((binEdges.length - 1) / (MAX_EDGE_TICK_LABELS - 1)),
+    );
     const dataTypeLabel = isDatetime
       ? "Date Range"
       : dataType === "string"
@@ -297,8 +296,15 @@ function HistogramChartComponent({
               type: "category",
               grid: { display: false },
               ticks: {
-                callback(_val, index) {
-                  return labels[index];
+                autoSkip: false,
+                callback(_value, index) {
+                  if (
+                    index % edgeTickInterval !== 0 &&
+                    index !== binEdges.length - 1
+                  ) {
+                    return undefined;
+                  }
+                  return formatAsAbbreviatedNumber(binEdges[index]);
                 },
                 color: themeColors.textColor,
               },
--- a/js/packages/ui/src/components/data/HistogramChart.tsx
+++ b/js/packages/ui/src/components/data/HistogramChart.tsx
@@ -244,6 +244,17 @@ function HistogramChartComponent({
       responsive: true,
       maintainAspectRatio: false,
       animation: animate ? undefined : false,
+      // Numeric bar widths come from their literal bin edges. Let Chart.js keep
+      // animating vertical properties, but do not create x/width animations
+      // that would later overwrite the edge geometry plugin's authoritative
+      // pixel values.
+      animations: isNumeric
+        ? {
+            numbers: {
+              properties: ["y", "base", "height"],
+            },
+          }
+        : undefined,
       plugins: {
         histogramBinGeometry: isNumeric ? { binEdges } : undefined,
         // The overlap painter reads its palette from here, not from a closure,
--- a/js/packages/ui/src/components/data/HistogramChart.tsx
+++ b/js/packages/ui/src/components/data/HistogramChart.tsx
@@ -19,6 +19,7 @@ import {
   formatAsAbbreviatedNumber,
   formatIntervalMinMax,
 } from "../../utils/formatters";
+import { histogramBinGeometryPlugin } from "./histogramBinGeometry";
 import {
   createHistogramLegendLabels,
   handleHistogramLegendClick,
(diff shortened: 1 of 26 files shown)


kentwelcome:
# Code Review: PR #1518

**SHA** `a71e8616` · **Verdict** NO-GO

## Blockers

1. `recce/tasks/histogram.py:512` — `_decimal_result_value` on the raw `min`/`max` echo fails the whole Histogram check for any numeric column whose extrema exceed float64 precision. The histogram itself is fine — bin edges are nice round numbers and round-trip cleanly — only the min/max passthrough raises, and it takes the run with it.
   Evidence: run against the shipped code —
   ```
   query_numeric_histogram OK for BIGINT id > 2^53 | edges 52
      min echo FAIL BIGINT id > 2^53 -> ValueError
   query_numeric_histogram OK for DECIMAL(38,18) | edges 45
      min echo FAIL DECIMAL(38,18) -> ValueError
   query_numeric_histogram OK for epoch nanos INT64 | edges 51
      min echo FAIL epoch nanos INT64 -> ValueError
   ```
   These are ordinary shapes, not the `1E400` extrema the new tests cover: a Snowflake/BigQuery sequence id past `9007199254740993`, an epoch-nanosecond `INT64`, a `DECIMAL(38,18)` amount. `_decimal_result_value(Decimal("0.3333333333333333333"))` raises too.
   Before this PR the same columns rendered: `result["min"]` carried the `Decimal` through and pydantic serialized it as a JSON string, which the numeric axis never reads —
   ```
   "result":{"min":"9007199254740993","max":"9007199254741993","bin_edges":[1,2]}
   ```
   So this is a straight regression from working chart to failed check, and it is reachable from both the UI launchers and the MCP `histogram_diff` tool.
   The strict check is right for `bin_edges` (line 306) — those define bucket boundaries and must be exact. It is wrong for the display-only extrema. Separate the two:
   ```python
   def _decimal_display_value(value):
       """Extrema are display-only: nearest finite float beats failing the run."""
       decimal_value = _decimal_value(value)
       result = float(decimal_value)
       if not math.isfinite(result):
           raise ValueError(f"Histogram Decimal {decimal_value} is not a finite JSON number")
       return int(decimal_value) if decimal_value == decimal_value.to_integral_value() else result
   ```
   and call it at lines 512-513, leaving line 306 on `_decimal_result_value`.
   Pass A.

## Issues

1. `recce/tasks/histogram.py:189` — `histogram_bucket_index` is dead production code. Its only caller is `tests/tasks/test_histogram.py:94`.
   Evidence: `grep -rn histogram_bucket_index recce/ tests/` returns the definition and that one test line; nothing in `query_numeric_histogram` or `_generate_histogram_sql` calls it. It re-implements in Python the ownership rule that the SQL `CASE` ladder implements (`_generate_histogram_sql:216-227`), so `test_histogram_bucket_index_counts_internal_and_terminal_edges_once` asserts the copy, not the thing that runs. Two sources of truth for the terminal-edge rule, and only one of them ships. The real SQL rule is already covered by `test_fractional_histogram_uses_decimal_edges_for_labels_and_real_sql_boundaries`. Delete the helper and the test that guards it, or call it from the SQL path.
   Pass F.

2. `js/packages/ui/src/components/histogram/HistogramDiffForm.tsx:191` — when a model has numeric columns but none exist in both environments, the select is disabled while the placeholder still reads "Select column" and every option is greyed out. Dead control, no explanation.
   Evidence: line 165 gates on `eligibleColumns.length === 0`, line 191 gates the message on `columns.length !== 0`. Reachable on a modified model where every numeric column was renamed (`amount` → `amount_usd`): `unionColumns` returns both, `columnAvailability` marks neither present in both, so `eligibleColumns` is empty while `columns` is not. The added/removed-model route is closed by `NodeView.tsx:523`, this one is not. The predicate that says the right thing is already computed on line 146.
   Pass A.

## Notes

- `recce/tasks/histogram.py:303` — `labels` is built one entry per edge, so the last entry describes `[last_edge, last_edge + width]`, a bin that does not exist. Same off-by-one as the code it replaces, and nothing in `js/` reads `result.labels`, so no behavior depends on it. Worth trimming while the function is open.
- `js/packages/ui/src/components/data/__tests__/HistogramChart.geometry.test.tsx` — every case does `plugins?.filter(({ id }) => id === "histogramBinGeometry")`, so the overlap painter is never in the chart under test. `histogramOverlap.intersectBars` reads `x ± width/2`, which this PR just made non-uniform; nothing asserts the crosshatch still lands on the shared interval.
- `js/packages/ui/src/components/lineage/NodeViewOss.tsx:305` — the new `trackProps` carries only `source`. The sibling launcher at `ColumnNameCell.tsx:174-178` sends `{action, source, node_count}`, so run-submit events from the model node will be missing `action` when the two are compared.
- `recce/tasks/histogram.py:198-204` — `_generate_histogram_sql` now interpolates the raw `{column}` into roughly one `WHEN` clause per bin instead of two sites total. `recce/tasks/profile.py:71` does the same job through `adapter.quote(column_name)`. Pre-existing pattern, not introduced here, but the PR multiplies the sites by ~25.
- `js/packages/ui/src/components/data/HistogramChart.tsx:185` — the `labels` array is still built for numeric charts and discarded on line 221.

kentwelcome (changes requested):

kentwelcome:
✅ **Code Review Loop — fixes pushed** (`aaa74189`)

- **Blocker** `histogram.py:512` min/max `ValueError` — fixed in `da8603be`. Split the strict round-trip check: bin edges keep `_decimal_result_value` (line 311) because they define bucket boundaries, the display-only extrema go through the new `_decimal_display_value`. Overflow and underflow are still rejected; a `BIGINT` past 2^53 and an epoch-nanosecond `INT64` now come back exact as Python ints, a `DECIMAL(38, 18)` comes back as the nearest double. New parametrised test fails against the pre-fix code and passes after.
- **Issue 1** dead `histogram_bucket_index` — removed in `da8603be`, along with the test that asserted the Python copy of the SQL ownership rule. `grep` over `recce/` and `tests/` returns nothing. The real rule stays covered by `test_fractional_histogram_uses_decimal_edges_for_labels_and_real_sql_boundaries`.
- **Issue 2** placeholder on a disabled picker — fixed in `aaa74189`. Line 191 now gates on the same `eligibleColumns` list as the `disabled` prop on line 165, and names the unmet condition. New test fails without the source change.

The five NOTEs are untouched — they are the author's call.

Gates on `aaa74189`: 1632 Python tests passed, 4244 frontend tests passed across 204 files, flake8, biome, and all three type-check projects clean. Unrelated `black` drift that `make format` produced in `mcp_server.py` and three other test files was reverted, not committed.

kentwelcome (approved):
Code Review Loop: blocking findings fixed and verified at aaa74189.

PR #1500, opened 2026-08-06
feat(ui): add semantic color contract and visual regression suite

gcko:
**PR checklist**

- [x] Ensure you have added or ran the appropriate tests for your PR.
- [x] DCO signed

**What type of PR is this?**

Feature / accessibility

**What this PR does / why we need it**:

- Adds a role-first semantic color contract for comparison, structural change, and direction meanings while preserving the legacy `DiffText.colorPalette` API.
- Migrates targeted shared grids, charts, lineage/CLL canvas components, row-count renderers, legends, and edge renderers away from overloaded red/green presentation.
- Adds redundant symbols, labels, patterns, accessible series descriptions, and low-zoom cues so meaning survives grayscale, Protanopia, Deuteranopia, and assistive-technology use.
- Adds a production-renderer Storybook matrix with deterministic Light, Dark, and Grayscale Darwin baselines; the production chart contracts also have OS-independent unit coverage in required CI.
- Documents the Phase B semantic contract, contrast requirements, CVD review procedure, and canonical visual-capture environment.

**Which issue(s) this PR fixes**:

Part of [DRC-2843](https://linear.app/recce/issue/DRC-2843).

This is the shared `@datarecce/ui` Phase B prerequisite. It intentionally does not close DRC-2843: Cloud adoption remains DRC-4115, and the parent still requires its per-surface child-ticket coverage and final team/Karen sign-off.

**Special notes for your reviewer**:

- Adversarial review fixes on `a0f3dd2a`: dark mode now reaches both React Flow node types; added-state rails pass compound/interaction contrast checks; inline Base/Current values expose assistive labels; low-zoom nodes retain structural symbols; and caller edge style overrides remain supported.
- Fresh reviewer verification on the pushed head: lint checked 677 files; root/UI/Storybook TypeScript passed; Vitest passed 194 files / 4,099 tests with 5 skipped; `@datarecce/ui` production and CSS builds passed; the isolated semantic Playwright suite passed 5/5 in 25.5 seconds.
- Light, Dark, and Grayscale baselines are reviewed at 1440×1466. Chromium vision-deficiency emulation was reviewed for Protanopia and Deuteranopia in both themes.
- The repository's full visual command still includes unrelated pre-existing missing histogram baselines and stale Timeline selectors; the semantic matrix is verified with its focused command.
[...]

PR #1521, opened 2026-08-26
fix(lineage): restore the status color on changed nodes

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Fix

Pull request #1500 added the `+`, `-` and delta icons to lineage nodes so colorblind readers can tell the states apart without color. It also shrank the colored area. A changed node kept only a 3px line on its left edge, and the minimap lost its color completely: every bar became the same gray.

This fills the node status block with the status color, carries the same color on the node border, and returns it to the minimap. The icons stay, so the non-color cue #1500 added is unaffected.

Closes DRC-4233

`COLOR_SEMANTICS.md` is the part to review. It said the secondary accent is never a fill. It now allows the fill on the lineage node status block and the minimap node, and leaves every other surface under the old rule.

The node border is colored, not only the status block, because an impacted node already gets a full colored border. Without this, the node that only inherits a change is more visible than the node that changed.

The colors are #1500's accessible palette, not the pre-#1500 ones.

The regenerated baselines also pick up recent histogram and profile-grid changes that merged without updating them. Those parts of the image diff are not from this PR.

Yes. Changed models in the lineage graph are marked with their status color again, on the node status block, the node border, and the minimap. The `+`, `-` and delta icons are unchanged.


gcko:
# Code Review: PR #1521

**SHA** `0610ec65` · **Verdict** NO-GO

## Blockers

None.

## Issues

1. `js/packages/ui/src/components/lineage/styles.tsx:58` — The `secondaryAccent` field doc still states the rule this PR exists to change.
   Evidence: `/** Transitional accent for a narrow rail or status symbol only */` — but `LineageNode.tsx:366` now assigns it to the card border, `:386` to the status-block fill, and `COLOR_SEMANTICS.md:6` was rewritten to permit both. The PR's own reviewer note calls the doc "the part to review"; the copy at the type site was missed, so the next author reading `ChangeStatusStyle` gets the superseded contract.
   Fix:
   ```ts
   /** Transitional accent: the lineage node border and status block, the MiniMap node, or a status symbol */
   secondaryAccent: string;
   ```
   Pass C.

2. `js/packages/ui/src/components/lineage/nodes/LineageNode.tsx:373` — `invertedText` has two branches; only the light one is tested.
   Evidence: the three new/changed assertions (`LineageNode.test.tsx:183`, `:202`, `:212`) all render at the default `isDark = false`, and the `dark mode` describe (`:694-708`) asserts only `toBeInTheDocument()`. The dark accents are *light* colors — `#4ADE80` / `#FCD34D` / `#F87171` (`semanticColors.ts:118-121`) — so inverting this ternary would put white on `#FCD34D` at ~1.2:1 and no test would fail. I confirmed the current behavior is correct by probe (`block-bg=#FCD34D block-color=#000000`), so this is a coverage gap, not a defect — but it is the one line in the diff whose regression is a WCAG failure.
   Fix: add the dark counterpart of the status-block test.
   ```tsx
   it("inverts the status block foreground in dark mode", () => {
     render(
       <LineageNode
         {...createMockNodeProps({ isDark: true }, { changeStatus: "modified" })}
       />,
     );

     const block = getComputedStyle(
       screen.getByTestId("lineage-node-status-block"),
     );
     expect(block.backgroundColor).toBe("#FCD34D");
     expect(block.color).toBe("#000000");
   });
   ```
   Pass E.

## Notes

- `js/packages/ui/src/components/lineage/columns/LineageColumnNode.tsx:328` — a changed *column* keeps the neutral border plus a 3px accent rail while the model node above it now carries a full accent border and a filled block. Permitted by the rewritten rule ("on every other surface it stays on a border or a symbol"), so not a defect — but the two treatments now sit adjacent in one canvas.
- The three regenerated `ColorSemanticsMatrix` baselines are correct and the unrelated drift is real: the changed regions outside the lineage row are the histogram edge-tick axis (DRC-2838, `80e5c1ee`/`d03b825a`/`441142b4`) and the humanized profile stat headers (DRC-2833, `9baaf804`), none of which regenerated the snapshots when they merged. Nothing in `.github/workflows/` runs the Playwright visual suite, so no CI gate depends on these files either way.
- Light-theme contrast of the new fills against the inverted foreground: added 5.02:1, removed 4.83:1, modified 5.02:1 — all above 4.5:1. Under grayscale the three fills collapse to near-identical luminance (L≈0.16), so the `+`/`−`/`Δ` icons remain the sole status discriminator, as #1500 intended.

## Verification

`pnpm lint`, `pnpm type:check`, `pnpm test` (205 files, 4260 passed) all clean at `0610ec65`. Branch merges cleanly into `main`; no base commits landed since it diverged.

gcko (changes requested):
Code Review Loop: 2 non-blocking-severity issues found (stale contract comment, untested dark-mode branch). See review comment.

gcko:
✅ **Code Review Loop — fixes pushed** (`14649e63`)

- Issue 1 — `styles.tsx:58` stale `secondaryAccent` contract — fixed in `14649e63`. The field doc now names the border, the status block, and the MiniMap node, matching `COLOR_SEMANTICS.md`.
- Issue 2 — untested dark branch of `invertedText` — fixed in `14649e63`. Added `inverts the status block foreground in dark mode`, pinning `#FCD34D` fill / `#000000` foreground / `#FCD34D` card border. Mutation-checked: flipping the ternary to `isDark ? colors.white : colors.black` fails both status-block tests; reverting turns them green again.

No production behavior changed — one comment and one test.

`pnpm lint`, `pnpm type:check`, `pnpm test` (205 files, 4261 passed, 5 skipped) clean at `14649e63`.

gcko (approved):
Code Review Loop: both issues fixed and verified at 14649e63.

PR #1522, opened 2026-08-26
fix(profile): include TIME and TIMETZ in temporal min/max contract

gcko:
## Summary
Addresses [DRC-4214](https://linear.app/recce/issue/DRC-4214/profile-include-timetimetz-in-the-temporal-minmax-contract)

In `recce/tasks/profile.py`, the `is_date_or_time` Jinja predicate previously checked `column_type.startswith('date')` and `column_type.startswith('timestamp')`, but missed warehouse aliases starting with `time` (such as `time`, `timetz`, `time with time zone`, `time without time zone`, `time(3)`, `timetz(6)`, etc.).

This PR adds `column_type.startswith('time')` to `is_date_or_time` so that:
- `min` and `max` compute extrema and cast them to string representation (`cast(min(...) as string)`), consistent with `DATE` and `TIMESTAMP` behavior.
- `avg` and `median` remain `cast(null as numeric)` (avoiding cross-warehouse temporal arithmetic).
- Non-temporal, numeric, boolean, struct, and array behaviors remain untouched.

## Validation
- Render & SQL dialect parsing tests added across DuckDB, Postgres, Redshift, Snowflake, BigQuery, Trino, and T-SQL/SQL Server.
- In-memory DuckDB test execution verifying `ProfileTask` and `ProfileDiffTask` with `TIME` columns and null/empty table handling.
- Full backend pytest suite passing (`1636 passed, 5 skipped`).
- Full frontend checks and static build passing (`4258 passed | 5 skipped`).


wcchang1115:
# Code Review: PR #1522

**SHA** `816d8ccc` · **Verdict** NO-GO

The one-line predicate change is correct. Reverting it fails all three new tests;
restoring it passes them. Full backend suite green (`1637 passed, 5 skipped`).
`is_date_or_time` has a single use (`recce/tasks/profile.py:101`), so the change
cannot reach `avg`, `median` or the `distinct_*` aggregates. Every finding below is
in the tests.

## Blockers

None.

## Issues

1. `tests/tasks/test_profile.py:662` — `test_profile_time_columns_empty_and_null` never
   profiles an empty table, so the `empty` half of its name is unverified.
   Evidence: the only empty table is `base_schema.empty_temporal_events` (line 674);
   line 678 inserts 2 rows into the current table; line 682 runs `ProfileTask`, whose
   `execute` reads `get_columns(model, base=False)` only (`recce/tasks/profile.py:319`).
   Deleting line 673-676 does not change the result. Pass E.

   The path itself is sound — profiling a 0-row TIME column returns
   `row_count=0, min=None, max=None, not_null_proportion=None` with no error — so this
   is missing coverage, not a hidden defect. Reach it with `ProfileDiffTask` and assert
   on `result.base`, or drop the base table and rename the test.

## Notes

1. `tests/tasks/test_profile.py:594-599` — `assert "null" in select_exprs["avg"].lower()`
   is a substring test, so a column whose name contains `null` satisfies it even when
   `avg` is a real aggregation.
   Evidence: rendering an `integer` column named `null_count` yields
   `AVG("null_count") AS avg`, and the assertion returns `True`. It holds today only
   because the column is `event_time`. `startswith("cast(null")` pins it. Pass E.

2. `tests/tasks/test_profile.py:523,543` — `DummyAdapter` and `DummyDbt` are now defined
   twice in one file, and the two copies already disagree.
   Evidence: `type_string` maps `bigquery`->`STRING`, `postgres`->`TEXT` at line 259; the
   new copy returns `VARCHAR` for every dialect at line 553. The stub defaults also differ
   (`"Postgres"` line 226, `"postgres"` line 525). Hoist one copy to module scope. Pass E.

3. `tests/tasks/test_profile.py:690` — the row loop has no `else` and no match counter, so
   renaming a column in the `CREATE TABLE` above leaves the test green with zero
   assertions executed.
   Evidence: `test_profile_task_with_time_columns` guards the same loop shape at line 639
   with `assert time_row is not None`. Pass E.

4. `recce/tasks/profile.py:48` — `column_type.startswith('timestamp')` can no longer decide
   the branch, because `startswith('time')` on line 49 matches every `timestamp*`. Pass F.

5. `tests/tasks/test_profile.py:611,617` — `event_date` and `event_ts` are created in both
   schemas but no assertion reads them. Pass E.

wcchang1115 (changes requested):
Changes requested — 1 issue, 5 notes. Full review: https://github.com/DataRecce/recce/pull/1522#issuecomment-5420118184

wcchang1115:
Resolved in 8c9c0933 (`test(profile): make the TIME tests verify what they claim`), pushed to this branch.

| finding | outcome |
|---|---|
| Issue 1 — empty base table created but never profiled | fixed: the test now calls `ProfileDiffTask` and asserts `row_count == 0`, null extrema, null `not_null_proportion` and null `is_unique` on the base result |
| Note 1 — `"null" in ...` substring check | fixed: matches the `cast(null` prefix instead. A column named `null_count` satisfied the old form while `avg` was a real aggregation |
| Note 3 — row loop could run zero assertions | fixed: a presence assertion now fails when a column is renamed |
| Note 5 — `event_date` / `event_ts` unasserted | fixed: their extrema are asserted |
| Note 2 — duplicate `DummyAdapter` / `DummyDbt` | **left open**: hoisting them restructures a test that already passed before this PR |
| Note 4 — dead `startswith('timestamp')` | **left open**: it edits the production file for a cosmetic gain, and the redundant branch names the two type families |

Every added assertion was mutation-tested: each one was broken in turn and confirmed to fail. Reverting the production predicate still fails three tests. Full backend suite `1637 passed, 5 skipped`.

The empty-table path was already correct before this commit, so that part is added coverage, not a defect fix.

Notes 2 and 4 are yours to take or drop.

wcchang1115 (dismissed):
Blocking issue resolved in 8c9c0933, which I pushed rather than the author, so this approval covers my own fix for that finding. Notes 2 and 4 are left open deliberately.

wcchang1115 (approved):
I reviewed 8c9c0933 a second time. It is my own fix commit. The review found no error. All 21 CI checks pass.

PR #1523, opened 2026-08-26
refactor(lineage): centralize CLL frontend state

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Refactor, tests, and documentation.

- Adds one internal `useCllState` controller for the current CLL result, fetch/cache-patch lifecycle, optional-input history, interaction generations, and published impact snapshots.
- Keeps lineage view options, focus, selection, presentation mode, and layout ownership in `LineageViewOss`, preserving the existing layout-generation commit guard.
- Makes async history restoration transactional: a previous entry is popped only after successful application and only if no newer entry replaced it while the layout was in flight.
- Resets TanStack mutation observer state with controller data, preventing stale pending/success/error status after a reset or late request completion.
- Records the per-node transport decision with reproducible payload measurements. The count-matched large map measured 15.64 MB raw JSON and about 19.4 MB retained heap; the backend already builds and caches the complete map.
- Confirms the single-environment release flag gates affordances rather than the CLL data path, and leaves state-library/context slicing to DRC-3315.

Closes DRC-3021

- This is the base PR for Stack A. DRC-3218 will branch from and target this branch so its offscreen-chain UI change stays reviewable independently.
- Closed draft PR #1244 was used only as an immutable captured-fixture source and performance reference; none of its stale client-side full-map implementation or bulk fixtures were revived.
- Independent review found and drove tests for mutation reset and overlapping async history restoration.

Verification from final HEAD:

- `pnpm lint`: 696 files checked, no errors
- `pnpm type:check`: root, `@datarecce/ui`, and Storybook pass
- `pnpm test`: 206 files; 4,269 passed, 5 skipped
- `pnpm run build`: Next.js static build succeeds

No visible behavior change.

```release-note
NONE
```

### DIFF
--- a/docs/cll-state-and-fetching.md
+++ b/docs/cll-state-and-fetching.md
@@ -39,4 +39,80 @@ Measurements ran locally with Node.js. Raw bytes use compact `JSON.stringify`, g
 
 These results are directional, not a production browser benchmark: they use a warm local Node runtime, omit network latency, and reproduce the proprietary project's counts and captured object shapes rather than its source payload. They are sufficient for the transport choice because the browser cost is paid up front even when a user inspects only one column.
 
+### Reproduction
+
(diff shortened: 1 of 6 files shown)


kentwelcome (approved):
LGTM

PR #1524, opened 2026-08-26
fix(lineage): signal filtered CLL chain boundaries

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Bug fix, tests, and documentation.

- Detects ancestry columns omitted when the new CLL column chain crosses models outside the current graph filter.
- Traverses only hidden parent/child components, stops at visible re-entry, and deduplicates counts per visible boundary model and direction.
- Renders compact `← N` / `N →` badges on the first/last visible annotation rows, with semantic image roles and descriptive accessible names.
- Emits no count or badge when the complete chain is visible; classic CLL and the global ancestry walk are unchanged.
- Records the expansion decision: no automatic/button/context-menu expansion for now, consistent with DRC-3273's overwhelm concern. Existing lineage filters remain the deliberate way to widen the graph.

Closes DRC-3218

- This PR is stacked on #1523 and intentionally targets its branch. Review this PR's four commits/diff only; after #1523 merges, GitHub can retarget this PR to `main` without code changes.
- Commit `5708b7e` is the RED checkpoint: visible A → hidden two-column B → visible C renders without either required boundary count.
- Root cause: `addColumnAncestryNodes` applied `filterSet` before retaining offscreen annotations or their relationships, making natural chain ends indistinguishable from filtered truncation.
- No deviations from the root-cause hypothesis. The only test correction changed “property absent” to the behaviorally relevant “optional count is undefined.”
- Independent review verified branch/cycle termination, deduplication, direction semantics, visible re-entry, unrelated-entry exclusion, classic-path isolation, and DRC-3273 alignment; it also drove the semantic accessibility assertion.

Verification from final HEAD:

- `pnpm lint`: 696 files checked, no errors
- `pnpm type:check`: root, `@datarecce/ui`, and Storybook pass
- `pnpm test`: 206 files; 4,271 passed, 5 skipped
- `pnpm run build`: Next.js static build succeeds
- Pre-push related tests: 106 files; 2,064 passed, 5 skipped

Yes. When a selected column's lineage continues through models hidden by the current graph filter, the visible boundary rows now show exact upstream/downstream hidden-column counts.

```release-note
Lineage now signals when a selected column's upstream or downstream chain continues beyond the models visible in the current graph filter.
```

### DIFF
--- a/docs/cll-chain-truncation.md
+++ b/docs/cll-chain-truncation.md
@@ -1,0 +1,35 @@
+# CLL chain truncation and expansion
+
+Status: accepted for [DRC-3218](https://linear.app/recce/issue/DRC-3218/cll-column-click-signal-and-resolve-truncation-when-the-impact-chain), 2026-08-26.
+
+## Decision
+
+Show a compact, non-interactive count at each visible column-lineage boundary, and do not add an expansion affordance for now.
+
+The upstream badge reads `← N`; the downstream badge reads `N →`. Its accessible label states the direction, exact number of ancestry columns hidden by the current graph view, and why they are absent. A full visible chain has no badge.
+
+## Options considered
+
+### Button on the annotation
+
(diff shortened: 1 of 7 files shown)


kentwelcome (approved):
LGTM

PR #1525, opened 2026-08-26
feat(lineage): move node analysis into dedicated tab

gcko:
## Summary

- replace the permanent six-button diff toolbar with a compact header Row Count action and a dedicated Analysis tab
- keep Profile, Value, Top-K, and Histogram launchers always visible; preserve the existing column-picker flows for Top-K and Histogram
- move Query into Code without changing the prefilled SQL or primary-key behavior
- add a flat, model-scoped Recent list with relative timestamps and an accessible View action that reopens the existing bottom result pane
- refresh only the exact run-list cache when a viewed run completes, without refetching the active run detail
- add a production Storybook scenario and regression coverage for tab inventory, focus changes, filtering, result reopen, and cache behavior

## Stack

This is the base PR for Stack B in the M4 Lineage & CLL burndown. DRC-2854 will target this branch so its NodeView cleanup builds on the settled tab structure.

## Cloud consumer

Cloud launch renders the shared `LineagePageOss -> LineageViewOss -> NodeViewOss` chain, so no Cloud-only component fork is required. The tab reaches Cloud when the next `@datarecce/ui` release containing this PR is adopted; that package-version adoption remains the deployment boundary.

## Test plan

- `pnpm lint` — 694 files clean
- `pnpm type:check` — root, `@datarecce/ui`, and Storybook clean
- `pnpm test` — 205 files; 4,272 passed, 5 skipped
- `pnpm run build` — production Next.js static build succeeds
- `pnpm --filter @datarecce/storybook run build` — production Storybook build succeeds
- pre-push — 88 related files; 1,804 passed, 5 skipped

## Deviations

- “Past runs” excludes only runs still in `Running`. Finished, Failed, and Cancelled analysis attempts remain in Recent because the existing bottom pane is the diagnostic surface for both results and errors.

Linear: https://linear.app/recce/issue/DRC-3463/nodeview-sidebar-move-the-diff-analysis-buttons-into-an-analysis-tab

### DIFF
--- a/js/packages/storybook/stories/lineage/NodeView.stories.tsx
+++ b/js/packages/storybook/stories/lineage/NodeView.stories.tsx
@@ -13,6 +13,7 @@ import type {
   NodeViewActionCallbacks,
   NodeViewNodeData,
   NodeViewProps,
+  RecentAnalysisRun,
   RunTypeIconMap,
 } from "@datarecce/ui/advanced";
 import { NodeView, RowCountSummary } from "@datarecce/ui/advanced";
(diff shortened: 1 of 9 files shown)


kentwelcome (approved):
LGTM

PR #1527, opened 2026-08-26
fix(cli): skip Sentry init for development builds

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

fix

`init()` in `recce/event/__init__.py` started the Sentry SDK for every environment, `development` included. Every git checkout carries a `.dev` version, so `_get_sentry_env()` returns `development`.

The result is that our own test runs and local development reported errors into the same Sentry project as released builds. This PR returns from `init()` before starting the SDK when the environment is `development`.

Fixes DRC-4259

The guard sits after the Amplitude setup, so a development build still configures the event collector.

Worth checking: Recce Cloud staging is not affected. Staging installs the `recce-nightly` package, whose version looks like `1.62.0.20260825`, and that maps to the `nightly` environment, not `development`. Released builds map to `production`.

NONE


kentwelcome (approved):
LGTM

PR #1532, opened 2026-08-28
fix(artifact): make comparability side-aware

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Bug fix

- Classifies base and current dbt artifacts independently and exposes bounded, side-aware health through Recce's info contract.
- Carries catalog evidence through lineage and schema coverage so checks, summaries, MCP approval, and persisted results fail closed when either side is incomplete.
- Adds shared UI types and helpers that render targeted base/current warnings without treating an incomplete comparison as a schema match.

Closes DRC-3303

- Companion Recce Cloud PR: https://github.com/DataRecce/recce-cloud-infra/pull/1703
- Merged with `main` to resolve the conflict with #1533 (`create_check`'s new `approve` argument). Both gates now apply: a check is approved only when the caller asked for approval **and** the evidence is complete-coverage with a verified-empty schema diff.

### Bugs found and fixed after an adversarial review of the tests

Three were regressions this branch itself introduced:

1. **One-sided models auto-approved as "no changes."** `classify_node_schema_comparison` returns `not_applicable` for a node present on exactly one manifest side, so it landed in neither the checked nor the unchecked bucket, and every consumer iterated `checked_node_ids` alone. `schema_diff_should_be_approved` returned `True` for a dropped model on this branch and `False` on `main`. The removal also vanished from the PR summary and the MCP payload. One-sidedness now has its own bucket — it is verified structural evidence, not a coverage gap.
2. **`unknown` coverage erased every schema row.** The grid suppressed rows for both `unchecked` and `unknown`. Only `unchecked` earns that; `unknown` means no coverage evidence exists at all, and suppressing there turns absence of evidence into affirmative absence of changes. This was live for Recce Cloud, whose lineage graph carries no `schema_comparison_status` — every Cloud schema check would have shown an incomplete banner over a table with only unchanged columns. (Cloud also now carries the field, in the companion PR.)
3. **Every SQLMesh project broken.** SQLMesh has no `catalog.json`; `columns_to_types` is authoritative. The adapter never set `catalog_status`, so the classifier read the missing key as legacy evidence: permanently partial coverage, all schema diffs suppressed, and remediation copy telling SQLMesh users to run `dbt docs generate`.

Also: `build_merged_lineage` now degrades an unparseable side health to `unknown` instead of raising, where a status literal from a producer on another version took down the entire lineage response.

### Test strengthening

Roughly 30 assertions added or rewritten, each verified to fail against a specific mutation the previous suite passed — including AC1 (empty by set intersection, not raw catalog totals), AC2 (all three degradation sides), AC3 (bounded samples at *and* over the cap), AC4 (`not_applicable` with a catalog present), AC5 (verified removals still visible in grid, summary and MCP payload), and AC10 (version skew). Two notable rewrites:

- The `approve=False` tests in **both** MCP backends were vacuous — their fixtures were unapprovable anyway, so deleting the gate was green in both. Rewritten so `approve` alone decides.
- No test asserted the payload an agent actually receives. Stripping `schema_coverage` from the `run_check` response, or counting compared nodes instead of changes, were both green.

### Open product questions for the reviewer

1. **MCP no longer auto-approves any non-schema check.** This branch removes `main`'s "blanket-approve any successful run — intentional per PM decision" and gates everything on `schema_result_is_approvable()`, which requires a schema-diff-shaped result. A `row_count_diff` check with matching counts created via MCP no longer auto-approves, though the CLI path (`run_should_be_approved`) still would. Intended?
2. **A schema check whose selector matches zero nodes reports `complete` and auto-approves** — an affirmative "verified clean" from a comparison that examined nothing. Currently pinned as desired by `test_tool_run_check_with_schema_diff`.

**Verification at this head**

- Backend: 1,803 passed, 5 skipped.
- Frontend: 4,324 passed, 5 skipped; lint and type-check pass.
- CI: 24 checks passing, 0 failures.

Yes. Recce now identifies which artifact side is incomplete, shows side-specific warnings, and prevents unsafe schema approval or match reporting. Genuine removals — including models present on only one side — stay visible, and a comparison with no coverage evidence no longer hides the diff it computed.

### DIFF
--- a/integration_tests/dbt/smoke_test.sh
+++ b/integration_tests/dbt/smoke_test.sh
@@ -94,7 +94,7 @@ grep -q '^## Manifest Information$' ./recce_summary.md
 lineage_section=$(awk '/^## Lineage Graph$/,/^## Checks Summary$/' ./recce_summary.md)
 grep -q 'model.jaffle_shop.customers' <<< "$lineage_section"
 
-checks_summary_row=$(grep -A 2 '^|Checks Run|Data Mismatch Detected|$' ./recce_summary.md | tail -1)
+checks_summary_row=$(grep -A 2 '^|Checks Run|Data Mismatch Detected|Incomplete Schema Comparisons|$' ./recce_summary.md | tail -1)
 checks_run=$(awk -F'|' '{gsub(/ /,"",$2); print $2}' <<< "$checks_summary_row")
 mismatch_count=$(awk -F'|' '{gsub(/ /,"",$3); print $3}' <<< "$checks_summary_row")
 # recce.yml defines two preset checks; only the row count one mismatches.
--- a/js/packages/ui/src/api/index.ts
+++ b/js/packages/ui/src/api/index.ts
@@ -62,6 +62,8 @@ export type { RecceServerFlags } from "./flag";
 export { getServerFlag, markRelaunchHintCompleted } from "./flag";
 // Server info API (lineage, environment, dbt metadata)
 export type {
+  ArtifactHealth,
+  ArtifactHealthPair,
   CatalogCoverage,
   CatalogMetadata,
   GitInfo,
@@ -78,6 +80,7 @@ export type {
   NodeColumnData,
   NodeData,
   PullRequestInfo,
+  SchemaCoverage,
   ServerInfoResult,
   SessionStaleness,
   SQLMeshInfo,
--- a/js/packages/ui/src/api/info.test.ts
+++ b/js/packages/ui/src/api/info.test.ts
@@ -91,5 +91,11 @@ describe("ServerInfoResult schema evidence contract", () => {
     expectTypeOf<MergedNodeData["schema_comparison_status"]>().toEqualTypeOf<
       "complete" | "unchecked" | "not_applicable" | undefined
     >();
+    expectTypeOf<MergedNodeData["base_catalog_status"]>().toEqualTypeOf<
+      "covered" | "unchecked" | "not_applicable" | undefined
+    >();
+    expectTypeOf<MergedNodeData["current_catalog_status"]>().toEqualTypeOf<
+      "covered" | "unchecked" | "not_applicable" | undefined
+    >();
   });
 });
--- a/js/packages/ui/src/api/info.test.ts
+++ b/js/packages/ui/src/api/info.test.ts
@@ -1,6 +1,13 @@
 import { describe, expectTypeOf, it } from "vitest";
 
-import type { CatalogCoverage, ServerInfoResult } from "./index";
+import type {
+  ArtifactHealth,
+  CatalogCoverage,
+  MergedLineageResponse,
+  MergedNodeData,
+  SchemaCoverage,
+  ServerInfoResult,
+} from "./index";
 
 type CatalogStatus =
   | "complete"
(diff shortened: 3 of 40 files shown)


kentwelcome:
# Code Review: PR #1532

**SHA** `7bff2105` · **Verdict** NO-GO

## Blockers
1. `recce/artifact_health.py:232` — A schema comparison that selected zero nodes reports `status="complete"`, so an empty diff is approvable and an MCP schema check whose selector matches nothing is auto-approved as verified-clean.
   Evidence: `status = "partial" if unchecked else "complete"` — with an empty `node_ids` all three buckets are empty, so the vacuous case takes the "complete" branch. `tests/test_mcp_server.py:1156` sets `select_nodes.return_value = set()` and line 1178 asserts `update_check_by_id(..., is_checked=True)`. This is the one place the PR's own invariant — absence of evidence is not affirmative absence of changes — is inverted. Raised by the author as open question 2.
   Pass A.

## Issues
1. `recce/mcp_server.py:2747` — Every non-schema check created or run through MCP stops auto-approving, `lineage_diff` included.
   Evidence: `run.py:110` requires `result_schema_coverage_status(result) == "complete"`, and `_canonical_schema_coverage` (`run.py:44`) returns `None` whenever `result["schema_coverage"]` is absent. A `row_count_diff` or `lineage_diff` result has no such key, so the gate can never pass. This deletes the DRC-3307 "Passed = Approved" behaviour whose comment the diff removes at `mcp_server.py` (-2825..-2839), and diverges from `run_should_be_approved()`, which still approves a matching `row_count_diff` on the CLI path. Raised by the author as open question 1; needs the product answer recorded before merge.
   Pass F.

2. `recce/mcp_server.py:275` — Cloud MCP auto-approve now requires the deployed Recce Cloud to emit `schema_coverage` inside run results, and the two PRs do not ship atomically.
   Evidence: `schema_result_is_approvable(run.get("result"))` runs against the raw `POST sessions/{id}/checks/{id}/run` payload (same at `:311`). Nothing in this repo produces that key on the cloud side — the only evidence it exists is the hand-built fixture `tests/test_mcp_cloud_backend.py:_schema_result`. recce-cloud-infra#1703 states "the released dependency bump should follow publication", so a user upgrading Recce before the cloud deploy lands gets silent, permanent non-approval in cloud mode. Confirm the deployed run-result payload carries `schema_coverage`, or make the client gate tolerate its absence until the cloud side is live.
   Pass F.

3. `js/packages/ui/src/components/schema/SchemaView.tsx:557` — The per-node "comparison incomplete" banner prints project-wide coverage numbers and a global 50-node sample that normally does not contain the node being viewed.
   Evidence: `schemaCoverage` (`SchemaView.tsx:289`) is `lineageGraph.schemaCoverage`, built by `classify_schema_coverage(base_nodes, current_nodes, all_ids)` at `recce/models/lineage.py:172` over *every* lineage node, then sorted and capped at 50 by `schema_coverage_payload`. On a 200-node project with 60 unchecked nodes, opening `model.shop.zz_widgets` renders "60 nodes were not checked. Unchecked nodes: <50 alphabetically-first, unrelated ids>, and 10 more" — the node the user is actually looking at is absent from its own banner. Every banner test uses a one-node lineage where `unchecked_nodes === [nodeId]` (`SchemaView.test.tsx:169`, `:230`, `:293`), so the multi-node text is unpinned. The sibling test "judges this node on its own status, not the project-wide coverage" covers banner *presence* only, not its body.
   Pass A / E.

## Notes
1. `js/packages/ui/src/api/info.ts:292` — `ServerInfoResult.artifact_health` and `.schema_coverage` are declared but never emitted (the backend only nests them under `lineage`, `recce/models/lineage.py:178`) and never read anywhere in `js/`. Pass C.
2. `recce/run.py:110` — The canonical evidence predicates (`schema_result_is_approvable`, `verified_schema_diff_is_empty`, `result_schema_coverage_status`) live in the CLI `run` module, and `mcp_server.py:34` imports from it. They belong beside `artifact_health.py`, which already owns the classifier half of the contract. Pass I.
3. `recce/run.py:186` — `schema_diff_should_be_approved` builds a synthetic one-row frame `("selection", "schema", "modified")`, serialises it, and re-parses it through `schema_result_is_approvable` only to re-derive `coverage.status == "complete" and not diff`. The round-trip adds a fake row shape that nothing else produces. Pass F.
4. `js/packages/ui/src/components/schema/SchemaView.tsx:559` — Remediation text hard-codes `dbt docs generate`. A SQLMesh model whose `columns_to_types` is empty gets `catalog_status: "unchecked"` (`recce/adapter/sqlmesh_adapter.py:64`) and is told to run a dbt command. Pass A.

## Limits
- Cannot verify from this repo whether the deployed Recce Cloud run-result payload carries `schema_coverage` (Issue 2). recce-cloud-infra#1703 is open and its dependency bump is deferred.

kentwelcome (changes requested):
Code Review Loop: blocking findings at `7bff2105`. See review comment.

kentwelcome:
✅ **Code Review Loop — fixes pushed** (`1aff927e`)

Every blocking finding from the review above is resolved. Re-verified by reading the pushed diff, not by taking the fix agent's word for it.

| # | Finding | Outcome |
|---|---|---|
| Blocker | Zero-node selection reported `complete` | Fixed — `classify_schema_coverage` returns `unknown` when nothing in the selection was comparable |
| Issue 1 | Non-schema checks stopped auto-approving | Fixed — gate is now type-aware |
| Issue 2 | Cloud auto-approve depended on an unshipped cloud field | Fixed for non-schema; schema checks fail closed by design |
| Issue 3 | Per-node banner quoted project-wide numbers | Fixed — node banner makes only the node-scoped claim |
| Notes 1–4 | Dead types, module placement, synthetic frame, dbt-only copy | All applied |

**Blocker.** `"complete"` now means "we compared at least one node and left none unchecked". It can no longer mean "we compared nothing". This also covers a selection made up entirely of `not_applicable` nodes, not just an empty one. `test_tool_run_check_with_schema_diff` was flipped rather than deleted, and a test was added asserting a zero-node selection is not approved.

**Issues 1 and 2 collapsed into one fix.** New `recce/schema_evidence.py` adds `result_is_schema_comparison`, applied at all four call sites (`mcp_server.py:275`, `:311`, `:2747`, `:2838`). A schema comparison must prove complete coverage; every other result type keeps the standing *Passed = Approved* rule.

Worth recording: **issue 1's blast radius was wider than the review said.** Beyond the three call sites named, this branch had also flipped preset-check runs, the task-branch run path, and the `create_check` idempotent-update path to non-approving — four further test assertions had been rewritten to `assert_not_called()`. Those are restored, with regression comments naming what breaks if the gate is widened again.

The predicate treats the *presence* of a `schema_coverage` block as sufficient to bind a result to the coverage contract, not just the three-column frame — so a malformed or truncated schema result still fails closed rather than escaping the gate as "some other type".

For issue 2 specifically: cloud **schema** checks still will not auto-approve until the cloud emits `schema_coverage`. That is the intended fail-closed behaviour, not an outage — and cloud non-schema checks are no longer caught in the rollout gap.

**Issue 3.** The project-wide sentence is gone from the node-scoped banner; `SchemaSummary` keeps it, where the claim genuinely is project-wide. Regression test uses a 50-node sample with `unchecked_node_count: 60` and a node absent from that sample — the shape every previous test missed by using a one-node lineage.

**Verification at `1aff927e`**

- Backend: 1,834 passed, 5 skipped (was 1,803 — +31)
- Frontend: 4,326 passed, 5 skipped across 207 files (was 4,324)
- `make flake8`, `pnpm type:check`, `pnpm lint`: clean
- Pre-commit chain ran in full (`core.hooksPath` = `js/.husky/_`)

The two open product questions in the PR description are now answered in code: a zero-node comparison no longer auto-approves, and non-schema checks keep their prior behaviour. If either answer is wrong as product direction, say so and it can be changed deliberately rather than as a side effect.

kentwelcome (approved):
LGTM

PR #1534, opened 2026-09-02
fix(mcp): stop posting a run for schema and lineage checks on cloud

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

fix

In cloud mode the MCP server posted a run for `schema_diff` and `lineage_diff` checks. A Recce Cloud session rejects those runs with `[HTTP 400] Run type '<type>' not supported`. Neither type has a Task in `dbt_supported_registry`, because both read the dbt manifests instead of querying the warehouse.

Two things followed. An agent had no way to create a schema diff check on a Cloud session, although the REST API and the UI both allow it. And the check was already created when the run failed, so the raised error discarded its `check_id` and left a check the caller could not find.

`create_check` and `run_check` now skip the run for these two types. A run request that does fail now returns `check_id` together with `run_error` instead of raising.

Fixes DRC-4310

`run_check` now issues `GET /checks/{id}` before it decides, because only the session knows the check type. That is one extra request per call.

`run_check` needs the same fix even though the report names `create_check`. A `schema_diff` check created in the UI or by `recce run` presets hits the same 400 when an agent runs it.

The Recce MCP server can now create and run schema diff checks against a Recce Cloud session.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/acf915ea-1739-4756-9dc8-5adb81d77374)

### DIFF
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -1475,9 +1475,8 @@ class RecceMCPServer:
                                 "Idempotent: if a check with the same (type, params) already exists "
                                 "in this session, its name and description are updated instead of "
                                 "creating a duplicate.\n\n"
-                                "On a Recce Cloud session a schema_diff or lineage_diff check "
-                                "executes no run: it reports run_executed=false, and `approve` "
-                                "alone decides approval.\n\n"
+                                "On a Recce Cloud session a schema_diff check executes no run: it "
+                                "reports run_executed=false, and `approve` alone decides approval.\n\n"
                                 "A failed run still returns check_id with run_error, so a caller can "
                                 "find the check it created."
                             ),
--- a/recce/mcp_server.py
+++ b/recce/mcp_server.py
@@ -119,6 +119,10 @@ class CloudBackend:
         "histogram_diff": "histogram_diff",
     }
 
+    # Both read the dbt manifests, so neither has a Task in dbt_supported_registry:
+    # a run request for them answers 400 "Run type '<type>' not supported".
+    CHECK_TYPES_WITHOUT_RUN = {"lineage_diff", "schema_diff"}
+
     def __init__(self, session_id: str, api_token: str, cloud_host: str = RECCE_CLOUD_API_HOST):
         self.session_id = session_id
         self.api_token = api_token
@@ -267,6 +271,16 @@ class CloudBackend:
         check_id = arguments.get("check_id")
         if not check_id:
             raise ValueError("check_id is required")
+        # Only the session knows the check type, and the type decides if a run is possible.
+        check = await self._request("GET", f"checks/{quote(str(check_id), safe='')}")
+        check_type = check.get("type")
+        if check_type in self.CHECK_TYPES_WITHOUT_RUN:
+            # Nothing runs, so the successful-run gate below never applies.
+            return {
+                "check_id": str(check_id),
+                "type": check_type,
+                "run_executed": False,
+            }
         run = await self._request(
             "POST",
             f"checks/{quote(str(check_id), safe='')}/run",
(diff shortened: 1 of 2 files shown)


iamcxa (approved):
@wcchang1115 — reviewed at `731f783e` (head moved from `a872d900`: I applied a one-line fix directly on the branch, see below).

## Fix applied

`create_check`'s tool description mentioned only `schema_diff` in the "no run" note, while `run_check` correctly named both `schema_diff` and `lineage_diff`. Since both types share `CHECK_TYPES_WITHOUT_RUN`, the wording should be consistent across the shared tool registry. Commit `731f783e` aligns it.

---

## Findings

### MEDIUM — `schema_diff`/`lineage_diff` checks auto-approved with no run-evidence gate

`_tool_create_check` lines 336-340:

```python
elif check_id and check_type in self.CHECK_TYPES_WITHOUT_RUN:
    if arguments.get("approve", True):
        await self._auto_approve(check_id)
```

For every other run-backed type two gates must both pass before approval: `_run_succeeded(run)` **and** `run_result_is_approvable(run.get("result"))`. The second gate enforced `schema_coverage.status == "complete"` — i.e. the comparison had to produce complete, verifiable coverage data. The new path bypasses both gates.

`CloudBackend._tool_schema_diff` (line 474) explicitly sets coverage to `classify_schema_coverage(None, None, ())` ("unassessed") because the Cloud session has no coverage data to surface. The old gate would have blocked approval on that result; the new path approves with the default `approve=True`.

The PR body acknowledges this explicitly — "approve alone decides" — so this is an intentional design tradeoff, not a bug. It's worth confirming whether the approval default should stay `True` for evidence-free types or switch to `False` (opt-in). Not a merge blocker, but worth a follow-up issue if the product gate cares about the distinction.

### Advisory (not blocking)

- **`InstanceSpawningError` swallowed in `_tool_create_check:320`** — the `try/except (RecceCloudException, InstanceSpawningError)` wraps the `POST /run` call. Unlike a failed run, a spawning error means the whole session is unavailable; swallowing it here lets `create_check` return success while nothing can subsequently run. Consider excluding `InstanceSpawningError` from this catch or noting the rationale in a comment.
- **Test completeness NIT** — `test_create_check_skips_the_run_for_schema_or_lineage_diff` verifies PATCH is called at the right URL but doesn't assert the body `{"is_checked": True}`. Peer tests in the same file (e.g. line 388) do verify this; worth making it consistent.

---

The core fix is correct and necessary. Tests are well-structured and cover the new paths.

wcchang1115:
Hi Kent,

`731f783e` reverted in `4d80fce2`.

`create_check` declares eight accepted check types and `lineage_diff` is not among them, so `jsonschema.validate` in `_handle_call_tool` rejects that type before the handler runs.

```
create_check(type=lineage_diff)
  -> Input validation error: 'lineage_diff' is not one of
     ['row_count_diff', 'schema_diff', 'query_diff', 'profile_diff', ...]
```

The inconsistency you spotted is real. But the real defect is the type list, not the description: the rest of the product accepts a lineage diff check. DRC-4324 tracks accepting the type.

Thank you!

iamcxa:
> Hi Kent,
> 
> `731f783e` reverted in `4d80fce2`.
> 
> `create_check` declares eight accepted check types and `lineage_diff` is not among them, so `jsonschema.validate` in `_handle_call_tool` rejects that type before the handler runs. Checked at `731f783e`:
> 
> ```
> create_check(type=lineage_diff)
>   -> Input validation error: 'lineage_diff' is not one of
>      ['row_count_diff', 'schema_diff', 'query_diff', 'profile_diff', ...]
> ```
> 
> The inconsistency you spotted is real, but it sits in the type list rather than the description. Everything else in the product accepts a lineage diff check: `POST /api/checks`, `createLineageDiffCheck` in the UI, and `recce run` presets. `_tool_create_check` also branches on `RunType.LINEAGE_DIFF`, so that branch is unreachable today, and the commit that introduced the tool says "schema_diff/lineage_diff skip auto-run (manifest-based)". The omission looks accidental. DRC-4324 tracks accepting the type and restoring your wording together.
> 
> `run_check` keeps naming both types. It takes any existing `check_id`, so it does reach a lineage diff check created elsewhere.

Got it! LGTM

PR #1539, opened 2026-09-04
feat(lineage): add compact validation summaries

gcko:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

Feature

- Adds compact lineage-node summaries for value, profile, top-k, and histogram validation results.
- Selects the latest valid evidence deterministically and handles both persisted mappings and live Pydantic results.
- Exposes counts and per-column run IDs without embedding raw frames, rows, values, labels, or bins.
- Refreshes lineage aggregates after all supported terminal validation runs.
- Preserves existing row-count payloads and state-loader interfaces.

Closes DRC-4260

- Stack position: 1 of 2 in the lineage stack.
- Base: `main`
- Child: https://github.com/DataRecce/recce/pull/1540
- GitHub stack: https://github.com/DataRecce/recce/stacks/1541
- Final focused backend/model suites: 65 passed.
- Full Python suite: 1,824 passed, 5 skipped.
- Full frontend suite: 4,344 passed, 5 skipped.
- Frontend lint/type checks, production build, repository flake8, targeted formatting, diff checks, and hooks passed.

Yes.

```release-note
Lineage run aggregates now include compact summaries for value, profile, top-k, and histogram validations without returning raw result payloads.
```

### DIFF
--- a/js/packages/ui/src/api/index.ts
+++ b/js/packages/ui/src/api/index.ts
@@ -139,9 +139,15 @@ export type {
 export { submitRowCountDiff } from "./rowcount";
 // Runs API (core run operations and aggregation)
 export type {
+  AggregatedRunResult,
+  ColumnValidationSummary,
+  NodeRunsAggregated,
+  ProfileDiffValidationSummary,
   RunsAggregated,
   SubmitOptions,
   SubmitRunTrackProps,
+  ValidationSummary,
+  ValueDiffValidationSummary,
 } from "./runs";
 export {
   aggregateRuns,
--- a/js/packages/ui/src/api/runs.ts
+++ b/js/packages/ui/src/api/runs.ts
@@ -6,18 +6,53 @@ import { isQueryRun, type Run, type RunType } from "./types";
 // ============================================================================
 
 /**
- * Aggregated run results by model and run type
+ * A full result retained for legacy row-count lineage decorations.
  */
-export type RunsAggregated = Record<
-  string,
-  Record<
-    "row_count_diff" | "value_diff" | "row_count",
-    {
-      run_id: string;
-      result: unknown;
-    }
-  >
->;
+export interface AggregatedRunResult {
+  run_id: string;
+  result: unknown;
+}
+
+interface AvailableValidationResult {
+  result_available: true;
+}
+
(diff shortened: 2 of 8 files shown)


wcchang1115:
# Code Review: PR #1539

**SHA** `8151e17` · **Verdict** GO

## Blockers

None.

## Issues

None.

## Notes

1. `recce/apis/run_func.py:513,514` — `validation_summary` publishes two sibling counters that count different units.

   `result_count` counts runs. Each per-column `top_k_diff` or `histogram_diff` run adds one (`:508`). `difference_count` counts columns inside the latest `value_diff` result (`:478`). The names give no hint of that difference. One `value_diff` run with 5 mismatched columns gives `{'result_count': 1, 'difference_count': 5}`, so `difference_count` exceeds `result_count`.

   `difference_count` also has no defined value for a node with no `value_diff` run. Line 472 leaves it at `0`, which reads the same as "a `value_diff` ran and found nothing".

   No consumer depends on the field yet. PR #1540 reads `summary.types.value_diff.difference_count` instead (`GraphNodeOss.tsx:133`). So this PR is the last point where a change costs nothing.

2. `recce/apis/run_func.py:355` — The tie-break for recency is `str(run.run_id)`, which is random.

   `run_at` has one-second resolution (`recce/models/types.py:54`). Two runs on the same model and column in the same second tie on time. The winner is then the higher UUID4. So `latest_run_id` can name the older run.

   `tests/apis/test_run_func.py:154` and `:182` lock this in. Both `profile_diff` runs carry `run_at="2026-09-04T11:00:00Z"`, and line 243 asserts that the higher UUID wins.

   Millisecond resolution on `run_at` makes the tie-break real.

3. `recce/apis/run_func.py:390` — `_value_diff_difference_count` reads `matched_p` at fixed position `row[2]`.

   The payload carries the column names in `data["columns"]`. The code never uses them to find the index. A producer with a different column order gives a wrong count and no error.

wcchang1115 (approved):
Reviewed at `8151e17`. GO — no blockers, no issues. Three notes in https://github.com/DataRecce/recce/pull/1539#issuecomment-5578419043

PR #1551, opened 2026-09-14
test(ui): smoke the built @datarecce/ui package in CI

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

test

`@datarecce/ui` is published to npm, but nothing in this repo runs its built output. `js/vitest.config.mts` and `js/tsconfig.json` resolve `@datarecce/ui` to `packages/ui/src`, and the Next build resolves it the same way. A change to the build tooling can therefore ship a broken package while every check passes.

This adds a check that imports every JavaScript entry in the package `exports` map from `packages/ui/dist`. It runs in the JS test workflow, after the unit tests and before the static build.

None.

- I checked the new test can fail, not only pass. I edited the built `dist/index.js` so it uses a variable one line before that variable is declared. That is how the earlier bundler problem appeared. `pnpm test:dist` then failed with `ReferenceError: Cannot access '__p' before initialization`. After I undid the edit, it passed again, 19 of 19.
- The smoke test needs a built `packages/ui/dist`, so `js/vitest.config.mts` excludes it from the normal `pnpm test` run. A plain `pnpm test` with no `dist` on disk was run locally and stays at the 211 file baseline.
- The release workflow is deliberately not changed here. Adding the same check before `npm publish` is worth doing, but `release-ui.yaml` carries two open code scanning alerts that turn the CodeQL check red on any PR touching it. That change belongs in a follow-up once those alerts are cleared.
- Verified under pnpm 12.4.1 after rebasing onto the pnpm 12 upgrade in main.

NONE


gcko:
# Code Review: PR #1551

**SHA** `4acafc6f` · **Verdict** GO

## Blockers
None.

## Issues
None.

gcko (approved):
Code Review Loop: GO at 4acafc6f. See https://github.com/DataRecce/recce/pull/1551#issuecomment-5673262857

PR #1552, opened 2026-09-16
chore(track): include the installed dbt version in the command event

wcchang1115:
**PR checklist**

- [x] DCO signed

**What type of PR is this?**

chore

Recce OSS does not start when dbt v2 is the only dbt in the environment. dbt v2 ships the warehouse drivers inside a Rust binary and drops the Python modules that the dbt adapter imports.

We cannot tell how common this is. A run that fails on the missing dbt modules looks the same as a run on a machine with no dbt installed at all.

This PR adds one property, `dbt_version`, to the `command` event that Recce already sends. It adds no new event, and the existing `anonymous_tracking` opt-out still gates it.

The property sits in the `finally` block of `TrackCommand.invoke`, so it also arrives when the command fails. That failing run is the case this is for.

Checked against a real dbt v2 environment: the property reports `2.0.2` when the command fails on the missing dbt 1.x import.

NONE


gcko (approved):
LGTM

